PR #17, opened 2015-03-26
Update never-subclass-components.md

joshdholtz:
The title "Never Subclass Components" is a bit confusing. It made it seem as if no components should ever be subclassed, not even `CKCompositeComponent `.

When viewing the title and first sentenced, it read as "Never Subclass Components. You should always subclass..." which almost contradicted each other (see image below). My proposed changes makes the title and first sentence flow a little better.

screen shot 2015-03-25 at 11 50 42 pm

### DIFF
--- a/_docs/never-subclass-components.md
+++ b/_docs/never-subclass-components.md
@@ -1,5 +1,5 @@
 ---
-title: Never Subclass Components
+title: Never Subclass Other Components
 layout: docs
 permalink: /docs/never-subclass-components.html
 ---
--- a/_docs/never-subclass-components.md
+++ b/_docs/never-subclass-components.md
@@ -1,10 +1,10 @@
 ---
-title: Never Subclass Other Components
+title: Never Subclass Components
 layout: docs
 permalink: /docs/never-subclass-components.html
 ---
 
-You should always subclass `CKCompositeComponent` when creating a component. (If you need to perform advanced layout by overriding `computeLayoutThatFits:`, you may subclass `CKComponent` directly, but this is rare.) Don't subclass other component classes.
+You should never subclass any other component classes besides `CKCompositeComponent` when creating a component. (If you need to perform advanced layout by overriding computeLayoutThatFits:, you may subclass CKComponent directly, but this is rare.)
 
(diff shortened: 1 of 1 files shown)


_docs/never-subclass-components.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Never Subclass Components
+title: Never Subclass Other Components
```
natansh: "Other" is too vague, IMO. I feel the heading's not really the problem, but the content should be fixed to flow better with the heading.

joshdholtz: @natansh The reason I thought other was okay was the last sentence in that first paragraph actually used the word other.
> Don't subclass other component classes.

But I agree that it is probably better for the content to be changed.

adamjernst:
Great suggestion! I agree that the wording is confusing.

I want to keep the title as-is. It's not actually correct, but it has rhetorical "punch" and I want to keep it that way.

I agree 100% that the wording of the article could be cleaned up. Maybe it could start with a clearer first sentence?

joshdholtz:
@adamjernst How about something like...
>You should never subclass any other component classes besides `CKCompositeComponent` when creating a component. (If you need to perform advanced layout by overriding computeLayoutThatFits:, you may subclass CKComponent directly, but this is rare.)

I will adjust my PR if that sounds good :blush:

kastiglione:
For :facepunch:, perhaps "Subclassing Is Bad"?

joshdholtz:
@adamjernst @natansh Reverted the title change and updated the first paragraph :zap: 

@kastiglione I didn't change the title to "Subclassing Is Bad" yet but that does get the point across :grin:

adamjernst:
Thanks @joshdholtz for contributing to ComponentKit!

_docs/never-subclass-components.md
```diff
@@ -4,7 +4,7 @@ layout: docs
 permalink: /docs/never-subclass-components.html
 ---
 
-You should always subclass `CKCompositeComponent` when creating a component. (If you need to perform advanced layout by overriding `computeLayoutThatFits:`, you may subclass `CKComponent` directly, but this is rare.) Don't subclass other component classes.
+You should never subclass any other component classes besides `CKCompositeComponent` when creating a component. (If you need to perform advanced layout by overriding computeLayoutThatFits:, you may subclass CKComponent directly, but this is rare.)
```
kastiglione: Missed some backquotes on the `CKComponent`.

PR #18, opened 2015-03-26
Add defensive code to the preparation queue

benlodotcom:
- Protection against a nil callback block
- Add a not designated initializer assert

### DIFF
--- a/ComponentKit/DataSources/Common/CKComponentPreparationQueue.mm
+++ b/ComponentKit/DataSources/Common/CKComponentPreparationQueue.mm
@@ -219,7 +219,9 @@
             didFinishPreparingBatchOfSize:job->_batch.items.size()
                                   batchID:(NSUInteger)job->_batch.ID];
     NSArray *outputBatchCopy = [outputBatch copy];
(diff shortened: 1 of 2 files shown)


ComponentKit/DataSources/Common/CKComponentPreparationQueue.mm
```diff
@@ -164,6 +164,11 @@ - (instancetype)initWithQueueWidth:(NSInteger)queueWidth
   return self;
 }
 
+- (instancetype)init 
+{
+  CK_NOT_DESIGNATED_INITIALIZER();
```
kastiglione: Add `CK_NOT_DESIGNATED_INITIALIZER_ATTRIBUTE` in header too?

benlodotcom: Oops had forgotten to add one commit.

cfcommando:
Why not NS_UNAVAILABLE instead?

grp:
One reason: a more semantic error/warning from the compiler.

PR #19, opened 2015-03-26
Replaced deprecated sizeWithFont: with sizeWithAttributes:

MarcoSero:
`sizeWithFont:` has been deprecated and the example project showed a warning.
Unless I'm missing the reason why `sizeWithFont:` was being used over `sizeWithAttributes:`, this should remove the warning and keep the same behavior.

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -230,7 +230,9 @@ static CGSize intrinsicSize(NSString *title, UIFont *titleFont, UIImage *image,
                             UIImage *backgroundImage, UIEdgeInsets contentEdgeInsets)
 {
   // This computation is based on observing [UIButton -sizeThatFits:].
-  CGSize titleSize = [title sizeWithFont:titleFont ?: [UIFont systemFontOfSize:[UIFont buttonFontSize]]];
+  CGSize titleSize = [title sizeWithAttributes:@{
+    NSFontAttributeName : titleFont ?: [UIFont systemFontOfSize:[UIFont buttonFontSize]]
+  }];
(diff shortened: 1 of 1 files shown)


kostiakoval:
This project has iOS7 capability, so I think that's why.
Maybe it should check if `sizeWithAttributes ` is available that use it,
and if not use `sizeWithFont`

MarcoSero:
Sorry I completely missed that by overlooking at the Podfile for the example project. It makes sense.
I'm not sure what's the best way to get rid of the warning then. I see two possible options:
1. As you said, check if the string responds to the new selector
2. Just ignore the warning with `"-Wdeprecated-declarations"`

kostiakoval:
I think both.  
 
1. Check that `sizeWithAttributes` is available with "[title respondToSelector]"  
If yes use it, otherwise
2. Disable this warning and use other method
```objc
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
 [title sizeWithFont ...]
#pragma clang diagnostic pop
```

adamjernst:
Thanks for helping clean this up!

The Travis failure highlights the reason this is nontrivial. `sizeWithAttributes:` returns a size that is subtly different than `sizeWithFont:`.

The aim is to match whatever `UIButton` does under the hood, so we need to figure out which is closer to `UIButton`'s internal calculations. If `sizeWithFont:` is closer, we should silence the warning using the clang diagnostic @kostiakoval mentions; if `sizeWithAttributes:` is closer, we need to re-record the snapshot tests.

adamjernst:
(`sizeWithAttributes:` is available in iOS 7, so there's no barrier to using it in ComponentKit unconditionally.)

PR #20, opened 2015-03-26
Remove Example project warnings

kostiakoval:
Xcode shows warnings for missing icons files for Example project.

I've cleaned up Assets and fixed those warnings.
Those changes are done by Xcode, auto generated not by hand, so the are safe.


adamjernst:
Thanks for doing thiS!

kostiakoval:
It was pure pleasure. So sad that don't have time at the moment to do more :sob:

PR #23, opened 2015-03-26
API to update context

beefon:
Currently we can pass a context only once when we init the data source.
It means that context should be static: \the component tree may become
inconsistent if we will update context during mount process.
In order to have it consistent, we need an API that will change the
context only for the future change sets, while leaving the old context
value for current chagesets.
Particular usecase is Accessibility features support: some of them may change over time (like Voice Over presence, dynamic text size category, etc.). If we would pass them via context, then we need to update it somehow, so components will have same context across the whole mount process.


ComponentKit/Core/CKComponentLifecycleManager.mm
```diff
@@ -128,6 +128,11 @@ - (void)updateWithStateWithoutMounting:(const CKComponentLifecycleManagerState &
   _state = state;
 }
 
+- (void)updateContext:(id)newContext
+{
+  _context = newContext;
```
adamjernst: What's the thread safety strategy here? Should this have `CKAssertMainThread`?

ComponentKit/DataSources/CKCollectionViewDataSource.mm
```diff
@@ -96,6 +96,11 @@ - (void)enqueueChangeset:(const CKArrayControllerInputChangeset &)changeset cons
   [_componentDataSource enqueueChangeset:changeset constrainedSize:constrainedSize];
 }
 
+- (void)updateContextAndEnqeueReload:(id)newContext
+{
+  [_componentDataSource updateContextAndEnqeueReload:newContext];
```
adamjernst: Same question—I imagine this is main thread only.

ComponentKit/Core/CKComponentLifecycleManager.mm
beefon: @benlodotcom Other methods don't have this assert. I wonder should I add it here and in other place @adamjernst mentioned?

ComponentKit/DataSources/CKCollectionViewDataSource.mm
benlodotcom: This is main thread only let's add an assert here.

ComponentKit/Core/CKComponentLifecycleManager.h
```diff
@@ -60,6 +60,13 @@ extern const CKComponentLifecycleManagerState CKComponentLifecycleManagerStateEm
 - (void)updateWithState:(const CKComponentLifecycleManagerState &)state;
 
 /**
+ Updates the context to the new one. 
+ 
+ Component provider will receive new context value next time the component tree will be built.
+ */
+- (void)updateContext:(id)newContext;
```
benlodotcom: I would not add mutable state to the CLM.

We might as well take out the context from the lifecycle manager, and pass it as an extra argument of prepareForUpdateWithModel. The context should be treated as any props at this layer.

ComponentKit/DataSources/Common/CKComponentDataSource.mm
```diff
@@ -27,7 +27,7 @@
 #import "CKComponentPreparationQueueListener.h"
 
 static const NSInteger kPreparationQueueDefaultWidth = 10;
-typedef CKComponentLifecycleManager *(^CKComponentLifecycleManagerFactory)(void);
+typedef CKComponentLifecycleManager *(^CKComponentLifecycleManagerFactory)(id aContext);
```
benlodotcom: Won't be needed if we take the context out of the CLM

ComponentKit/DataSources/Common/CKComponentDataSource.mm
```diff
@@ -200,6 +212,7 @@ - (PreparationBatchID)enqueueChangeset:(const CKArrayControllerInputChangeset &)
     if (type == CKArrayControllerChangeTypeUpdate) {
       CKComponentDataSourceInputItem *oldInput = [_inputArrayController objectAtIndexPath:indexPath.toNSIndexPath()];
       lifecycleManager = [oldInput lifecycleManager];
+      [lifecycleManager updateContext:_context];
```
benlodotcom: ditto

ComponentKit/Core/CKComponentLifecycleManager.h
```diff
@@ -34,15 +35,14 @@ extern const CKComponentLifecycleManagerState CKComponentLifecycleManagerStateEm
 /**
  Designated initializer
  */
-- (instancetype)initWithComponentProvider:(Class<CKComponentProvider>)componentProvider
-                                  context:(id)context;
```
benlodotcom: id<NSObject>

ComponentKit/DataSources/Common/CKComponentDataSourceInputItem.mm
```diff
@@ -23,18 +23,21 @@ @implementation CKComponentDataSourceInputItem
...
     _lifecycleManager = lifecycleManager;
     _model = model;
+    _context = context;
     _constrainedSize = constrainedSize;
     _UUID = [UUID copy];
 
     NSUInteger subhashes[] = {
       [_lifecycleManager hash],
       [_model hash],
+      (NSUInteger)_context,
```
benlodotcom: This way you can do `[_context hash]` here

PR #37, opened 2015-03-27
Fixed deprecation warning for -[NSString sizeWithFont:]

soxjke:


ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -230,11 +230,16 @@ static CGSize intrinsicSize(NSString *title, UIFont *titleFont, UIImage *image,
...
+      attributes = [NSMutableDictionary dictionaryWithCapacity:1];
+  }
+  attributes[NSFontAttributeName] = titleFont ?: [UIFont systemFontOfSize:[UIFont buttonFontSize]];
+  CGSize titleSize = [title sizeWithAttributes:attributes];
   CGSize imageSize = image.size;
   CGSize contentSize = {
-    titleSize.width + imageSize.width + contentEdgeInsets.left + contentEdgeInsets.right,
-    MAX(titleSize.height, imageSize.height) + contentEdgeInsets.top + contentEdgeInsets.bottom
+    (CGFloat)ceil(titleSize.width) + imageSize.width + contentEdgeInsets.left + contentEdgeInsets.right,
+    MAX((CGFloat)ceil(titleSize.height), imageSize.height) + contentEdgeInsets.top + contentEdgeInsets.bottom
```
ocrickard: Hmm, are we sure just `ceil`ing this works in every case?  This is supposed to closely match the internal behavior of UIButton so we probably need to check.  Perhaps we should add some more exhaustive snapshot tests to confirm.

ocrickard:
This was also attempted in https://github.com/facebook/componentkit/pull/19  It would be great if we could use the new API, but we need to verify this is the same behavior as UIButton.

soxjke:
@ocrickard what is preferred way to test, create UIButton programmatically, or manually add snapshots to ReferenceImages folder?

ComponentKit/Components/CKButtonComponent.mm
adamjernst: At a minimum we should use `CKCeilPixelValue`.

ocrickard:
OK, I investigated this a little bit, and I believe UIButtonLabel is calling `-[NSString _legacy_sizeWithFont:forWidth:lineBreakMode:]:` internally even on iOS 8.  As a result I believe continuing to use the deprecated selector here remains more semantically correct.  We should ignore the deprecated warning with a pragma and revisit in iOS 9.

Re the question of how to approach testing: yeah, I'm not sure what the best way of testing this would be... but what you're saying makes sense.  Both adding more snapshots and iterating over many combinations of characters and verifying that the sizing output of UIButton and this function are equivalent.

ComponentKit/Components/CKButtonComponent.mm
ocrickard: Yeah, it appears UIButton is **rounding** to the nearest pixel value, so should we consider doing this as well?

dstnbrkr:
+1 to ignoring for now: https://github.com/facebook/componentkit/pull/39

PR #43, opened 2015-03-29
Standalone ComponentKit.xcodeproj

kastiglione:
`ComponentKit.xcodeproj` cannot currently be used without CocoaPods. Are there reasons for not providing a working .xcodeproj?

This pull request does some cleanup, but mainly thus far it adds a working static library target. Tests do not quite build yet, they need to be uncoupled from CocoaPods a little bit.

There's only one test that uses OCMock. Are there plans to make more widespread use of OCMock in ComponentKit's testing? If not, that one test could be done without OCMock, which would remove the one external dependency.


kastiglione:
Does the Travis CI integration use on CocoaPods?

dstnbrkr:
The current approach uses cocoapods to minimize the amount of maintenance needed on an xcodeproj. Currently, the ComponentKit.xcodeproj is dead simple and all configuration flags are set via the ComponentKit.podspec. Aside from not having to maintain settings in both a podspec and an xcodeproj, this setup forces our CI to constantly validate our podspec.

kastiglione:
> Currently, the ComponentKit.xcodeproj is

There are targets for the tests. Does adding a standard library target break things for CocoaPods?

> all configuration flags are set via the ComponentKit.podspec

There are currently five config flags, that seems reasonable to maintain. At what point would it become a challenge?

> this setup forces our CI to constantly validate our podspec.

If I wasn't clear, I had no intentions of changing anything about how CocoaPods is used, not trying to overhaul the existing setup. I only wanted to make it so that those who do not use CocoaPods can also use ComponentKit.

Is the cost of cross maintenance of the `Podfile` and `ComponentKit.podspec` great enough to exclude non-CocoaPods users?

dstnbrkr:
Guessing you're adding Carthage support? I'm inclined to wait until we've heard a demand for this a few times, but if you're feeling strongly about it, resubmit this once you have the tests passing.

You'll need to add the target for ComponentKitTestLib, which depends on FBSnapshotTestCase and is depended on by ComponentKitApplicationTests and ComponentTextKitApplicationTests.

Ideally, ComponentKit would build and pass CI on each commit so we could bisect if we had to. You can run: ```build.sh ci``` locally to verify each commit (or squash into one). I think for commits that touch the build to this degree, squash is preferable.

Also, I opened #44 to set the public headers explicitly in the podspec. This project will need to also correctly specify which headers are public (and we'll need to keep the two in sync).

kastiglione:
> Guessing you're adding Carthage support?

That could come from it, but isn't the motivation here. (I'm not yet sure if Carthage works for projects that have an iOS 7 deployment target, since Carthage builds frameworks which of course don't deploy to iOS 7).

After cloning the repo, and opening the .xcodeproj, I was surprised that there was no source to browse, and no targets to build. CocoaPods and Carthage are no the only options, there's the Xcode way of dragging a project into a workspace (or even a project), which ComponentKit doesn't allow. CocoaPods is very successful, but there are many who do not use it. I guess I see having a standalone capable .xcodeproj as being somewhat fundamental, as flawed as it is.

> resubmit this once you have the tests passing

Apologies for not marking this for what it is: Work In Progress™. This conversation is why I put it up the pull request early, to see what was at play.

Thanks for the details.

PR #50, opened 2015-03-30
Add no argument Changeset constructor

kastiglione:
The [datasource-basics](https://github.com/facebook/componentkit/blob/ff6c37c428ce878eed5e6306741e3a4925959620/_docs/datasource-basics.md) docs show being able to create a `CKArrayControllerInputChangeset` without arguments.

If this is legit, here's the update. If not, those docs should change.

### DIFF
--- a/ComponentKit/Utilities/CKArrayControllerChangeset.h
+++ b/ComponentKit/Utilities/CKArrayControllerChangeset.h
@@ -170,6 +170,7 @@ namespace CK {
       
       class Changeset final {
       public:
+        Changeset() : items({}), sections({}) {}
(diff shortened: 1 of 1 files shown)


kastiglione:
> ERROR:  While executing gem ... (Gem::UnsatisfiableDependencyError)
>    Unable to resolve dependency: 'cocoapods (= 0.36.1)' requires 'xcodeproj (~> 0.23.1)'
> The command "gem install cocoapods" failed and exited with 1 during .

kastiglione:
I see this change would let the compiler allow `-enqueueChangeset:` to be called with `{}`, so maybe it's the docs that should change?

benlodotcom:
The docs should be changed, will fix.

PR #53, opened 2015-03-31
Change type of view passed to hit test hooks

adamjernst:
If we want to support non-CKComponentRootViews that use these hooks, it has to be a generic UIView.


adamjernst:
I ran build.sh locally, it passes. Looks like Travis just barfed for no reason.

PR #68, opened 2015-04-07
Remove residual CKAssertMainThread

kastiglione:
As far as I can tell, this assert isn't necessary.

### DIFF
--- a/ComponentKit/DataSources/Common/CKComponentDataSource.mm
+++ b/ComponentKit/DataSources/Common/CKComponentDataSource.mm
@@ -10,6 +10,7 @@
 
 #import "CKComponentDataSource.h"
 
+#include <libkern/OSAtomic.h>
 #include <queue>
 
 #import <ComponentKit/CKSectionedArrayController.h>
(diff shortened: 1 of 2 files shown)


kastiglione:
cc @rohan-mehta

kastiglione:
@benlodotcom I have a feeling you might weigh in on this one.

adamjernst:
Definitely intentional at this point.

kastiglione:
Because of other callers? I'm not seeing how this method itself needs an assert.

If callers, then the assert should be moved to those.

adamjernst:
Defer to Ben Loulier.

kastiglione:
@benlodotcom Some related discussion is in #71.

benlodotcom:
Sounds good to me for the preparation queue, can you move the assert in `CKComponentDataSource` before merge though ?

kastiglione:
@benlodotcom In the update, `CKComponentDataSource` tracks whether an enqueue is in progress, and will assert if two enqueues are run concurrently.

adamjernst:
I'm actually tending more against this proposal because as I rewrite the DataSource, I want to use the main thread as a convenient synchronization point.

I could make it accept a "target queue" i suppose... Let's discuss in person.

kastiglione:
Ah, I see. It seems the two off-main modes are at odds: concurrent use makes dispatching-to-main convenient, but serial use makes dispatching-to-main more ritual.

I'm not pushing for this, it's not currently affecting me. Ideally it'd be nice to not have to dispatch to main for the serial case, but not if it has brings problems for concurrent use.

adamjernst:
Cool. Let's close for now. Once the "new" STM data source lands, we could make it take a target queue for all serialization, then mark that queue and require that all calls to it, and callbacks from it, happen on that queue.

Issue #64, opened 2015-04-06
CKCollectionViewDataSource: number of sections and items in section


MarcoSero:
Is there a reason why `CKCollectionViewDataSource` doesn't expose 

```objective-c
- (NSInteger)numberOfSections;
- (NSInteger)numberOfObjectsInSection:(NSInteger)section;
```

the same way `CKComponentDataSource` does?

Or maybe it could expose that it implements `UICollectionViewDataSource`?

I'd probably prefer the same interface as `CKComponentDataSource`, since implementing `UICollectionViewDataSource` is just an implementation detail.

benlodotcom:
These pieces of state are always in sync with the collectionView and they are exposed on `CKComponentDataSource` because it is needed to implement the `UICollectionViewDataSource` protocol. 

I'm not sure we should expose it at the `CKCollectionViewDataSource` layer as it would leak some implementation details, is there something preventing you to query it from the collectionView ?

MarcoSero:
That makes perfectly sense :+1: 
Will query the collection view.

PR #69, opened 2015-04-07
Add -numberOfSections and -numberOfObjectsInSection:

MarcoSero:
`-numberOfSections` and `-numberOfObjectsInSection:` were not exposed in
`CKCollectionViewDataSource`'s interface.
Internally they simply bridge to `_componentDataSource`.

See my question https://github.com/facebook/componentkit/issues/64


MarcoSero:
Not necessary, as explained by @benlodotcom here https://github.com/facebook/componentkit/issues/64#issuecomment-90631205

PR #71, opened 2015-04-07
Replace CKComponentDataSource queue with atomic counter

kastiglione:
This change would make this aspect of `CKComponentDataSource` thread safe.

From my reading, the queue here currently fills two needs:

1. Assert that `-enqueueBatch:block:` processes batches serially.
2. Track whether changes are actively being processed (`isComputingChanges`)

The first item isn't necessary, `-enqueueBatch:` uses a serial dispatch queue. Unless there are plans/desires to make the preparation queue operate in parallel, then the first item is essentially asserting that dispatch queues work.

The second item can be implemented with an atomic counter.

### DIFF
--- a/ComponentKit/DataSources/Common/CKComponentDataSource.mm
+++ b/ComponentKit/DataSources/Common/CKComponentDataSource.mm
@@ -284,7 +284,7 @@ CK_FINAL_CLASS([CKComponentDataSource class]);
   [_componentPreparationQueue enqueueBatch:preparationQueueBatch
                                      block:^(const CKArrayControllerSections &sections, PreparationBatchID ID, NSArray *outputBatch, BOOL isContiguousTailInsertion) {
                                        int32_t newCount = OSAtomicDecrement32(&_activeOperationCount);
-                                       CKInternalConsistencyCheckIf(oldCount >= 0, @"We dequeued more batches than what we enqueued something went really wrong.");
+                                       CKInternalConsistencyCheckIf(newCount >= 0, @"We dequeued more batches than what we enqueued something went really wrong.");
                                        [self _componentPreparationQueueDidPrepareBatch:outputBatch
                                                                               sections:sections];
                                      }];
--- a/ComponentKit/DataSources/Common/CKComponentDataSource.mm
+++ b/ComponentKit/DataSources/Common/CKComponentDataSource.mm
@@ -283,8 +283,8 @@ CK_FINAL_CLASS([CKComponentDataSource class]);
   OSAtomicIncrement32(&_activeOperationCount);
   [_componentPreparationQueue enqueueBatch:preparationQueueBatch
                                      block:^(const CKArrayControllerSections &sections, PreparationBatchID ID, NSArray *outputBatch, BOOL isContiguousTailInsertion) {
-                                       int32_t oldCount = OSAtomicDecrement32(&_activeOperationCount);
-                                       CKInternalConsistencyCheckIf(oldCount > 0, @"We dequeued more batches than what we enqueued something went really wrong.");
+                                       int32_t newCount = OSAtomicDecrement32(&_activeOperationCount);
+                                       CKInternalConsistencyCheckIf(oldCount >= 0, @"We dequeued more batches than what we enqueued something went really wrong.");
                                        [self _componentPreparationQueueDidPrepareBatch:outputBatch
                                                                               sections:sections];
                                      }];
--- a/ComponentKit/DataSources/Common/CKComponentDataSource.mm
+++ b/ComponentKit/DataSources/Common/CKComponentDataSource.mm
@@ -10,7 +10,7 @@
 
 #import "CKComponentDataSource.h"
 
-#include <queue>
+#import <libkern/OSAtomic.h>
 
 #import <ComponentKit/CKSectionedArrayController.h>
 
@@ -54,7 +54,7 @@ CKComponentLifecycleManagerAsynchronousUpdateHandler
   CKSectionedArrayController *_outputArrayController;
   CKSectionedArrayController *_inputArrayController;
   CKComponentPreparationQueue *_componentPreparationQueue;
-  std::queue<PreparationBatchID> _operationsInPreparationQueueTracker;
+  volatile int32_t _activeOperationCount;
   CKComponentLifecycleManagerFactory _lifecycleManagerFactory;
 }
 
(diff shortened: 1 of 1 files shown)


adamjernst:
Why?

kastiglione:
Sweet, test failures!

> Why?

1. These pieces are now atomic, previously not.
1. Counter is easier to reason about than a queue. The queue increases internal complexity and potential operations relative to a counter.
1. The points above about how the queue is used (asserts that will never fire, just the count)

kastiglione:
To add, I'm sure there may be reasons why not. This was a quick bus pull request. If rejected, hopefully I'll at least gain some implementation insight along the way.

adamjernst:
Here's my 2¢:

- You should only ever call `enqueueChangeset:` from one queue/thread. That's a given—otherwise the ordering is undefined and thus nonsensical.
- Right now we enforce that you only call it from the main thread. This is simple but limiting.
- Your change makes it possible to call it from any (single) thread/queue. However in my opinion adding a threadsafe counter (via `OSAtomic`) is the wrong choice, because in the case where you call it from multiple threads/queues, you're making a nonsensical call anyway.
- Thus I'd rather we just remove the `OSAtomic` calls and use regular arithmetic.
- A nice optional feature might be the ability to specify a dispatch context key so that you can say "hey, I'm going to call you from a queue with this dispatch context set, so please assert to make sure I don't violate that assertion". This would leave guard rails on but allow non-main-thread use. Probably overkill.

Make sense?

kastiglione:
Makes sense.

> You should only ever call `enqueueChangeset:` from one queue/thread. That's a given—otherwise the ordering is undefined and thus nonsensical.

What are the arguments against behaving like arrays and other mutable data structures: not inherently thread safe, but can be safely used across multiple threads when the callers guarantee serial changes. Is that too risky of an API?

> A nice optional feature might be the ability to specify a dispatch context key so that you can say "hey, I'm going to call you from a queue with this dispatch context set, so please assert to make sure I don't violate that assertion".

:+1: Interesting.

kastiglione:
> A nice optional feature might be the ability to specify a dispatch context key so that you can say "hey, I'm going to call you from a queue with this dispatch context set, so please assert to make sure I don't violate that assertion".

Wouldn't that require calling `dispatch_get_current_queue`, which is deprecated?

adamjernst:
> What are the arguments against behaving like arrays and other mutable data structures: not inherently thread safe, but can be safely used across multiple threads when the callers guarantees potential multithreaded changes are serial. Is that too risky of an API?

No, I think you're in the right here. I do want people to be caught when they screw up, but I guess we should let you take care of it yourself (like `NSMutableArray`).

> Wouldn't that require calling `dispatch_get_current_queue`, which is deprecated?

Oops, I meant `dispatch_get_specific` and friends.

So in sum: let's go back to just removing the assert, should we wish to pursue this. Sorry for the runaround.

kastiglione:
> Oops, I meant `dispatch_get_specific`

Ah :+1:

adamjernst:
Reopen on update?

benlodotcom:
> What are the arguments against behaving like arrays and other mutable data structures: not inherently thread safe, but can be safely used across multiple threads when the callers guarantees potential multithreaded changes are serial. Is that too risky of an API?

This is definitely defensive measures enqueueChangeset is not safe for concurrent access, and more things needs to happen for it to be, the counter would be a first step but the main one would be the mutation of the inputChangeset. Also making this API concurrent wouldn't make much sense because changesets are not independent (or at least not guaranteed to be) and hence need to be enqueued in a specific order. So I wouldn't bother changing the counter to be atomic this code path shouldn't be accessed concurrently anyway.

I agree with you that enforcing a call on the main thread is more than what we are looking for (which is just serialization of the inputs), happy to move to a model where we don't enforce main thread anymore but  it would still be nice to enforce that enqueues are done serially and I like @adamjernst suggestion there.

Small digression but moving away from a changesets API and just pass the whole state of the world + do diffing (basically automatically generate a changeset from the current state to the other) behind the scenes would alleviate those issues as well, in that case concurrency just becomes an internal versioning problem as each input is independent from the others and the only constraint for eventual consistency is to make sure we never apply a less "recent" state than the one currently displayed. IMO all of this doesn't belong in this layer though and could be proposed as an adapter in front of the datasources that also handles serialization of the "generated" changesets.

cfcommando:
FYI: My pipeline of data into the data source is (intentionally) bottlenecked by passing through a serial queue to ensure proper ordering of updates, but the point at which the data is actually enqueued is not necessarily on any particular queue - I only punted to the main thread because the data source unfortunately requires a reference to the view hierarchy (though strictly speaking, it only needs a size) to feed an appropriate constrained size.

I would prefer safety with an atomic counter over queue enforcement.

kastiglione:
> I only punted to the main thread because the data source unfortunately requires a reference to the view hierarchy (though strictly speaking, it only needs a size) to feed an appropriate constrained size.

@cfcommando I didn't quite follow you on this.

> I would prefer safety with an atomic counter over queue enforcement.

I wonder how common your scenario is: changes are enqueued serially, but not all from the same queue. In this case, do you have any ideas how the data source would recognize and assert when changes are enqueued concurrently?

cfcommando:
> @cfcommando I didn't quite follow you on this.
`[_dataSource enqueueChangeset:items constrainedSize:[_sizeRangeProvider sizeRangeForBoundingSize:self.collectionView.bounds.size]];` <--The only main thread dependency my code for queuing up changes has. But it was a throwaway explanation that's out of scope for this issue.

> I wonder how common your scenario is: changes are enqueued serially, but not all from the same queue. In this case, do you have any ideas how the data source would recognize and assert when changes are enqueued concurrently?

I am the gatekeeper to the data source, so I know that changes will only come down for a given data source serially, but said updates might be scheduled on an arbitrary available queue. I don't have a strict policy restricting those updates to a specific queue.

kastiglione:
> In my case I guarantee

Imagine I have some code that I make the same claim about, but I'm actually wrong. What should Components do to assert when it's being misused, and alert the author that they've missed something? Currently, it's a main thread assert, which is restrictive, but there seems to be agreement that some checks have to be in place.

cfcommando:
Doesn't your atomic counter address that?

kastiglione:
Ah, no. It made that specific code safer if called concurrently, but as @adamjernst and @benlodotcom have brought to attention, concurrent changes are undefined, so it enables something callers shouldn't do. It does not assert when misused, which is where the conversation has moved to.

cfcommando:
Ah, I see. Now that I understand better, pegging the enqueue itself to a serial queue is entirely reasonable, with a queue assert built around `dispatch_get_specific`.

PR #91, opened 2015-04-09
Remove 'treat warnings as errors' flag

dstnbrkr:
CocoaPods doesn't scope these flags to the pod

### DIFF
--- a/ComponentKit.podspec
+++ b/ComponentKit.podspec
@@ -18,6 +18,5 @@ Pod::Spec.new do |s|
   s.xcconfig = {
     'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11',
     'CLANG_CXX_LIBRARY' => 'libc++',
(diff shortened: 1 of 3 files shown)


cfcommando:
No `s.compiler_flags = '-Werror -Wall'`? That applies to individual files and is, I think, equivalent to the originally-desired effect of the GCC flag.

cfcommando:
...not that that's a problem for me, it just means less strictness for CK's compilation.

dstnbrkr:
@cfcommando yep, just fixing the primary problem for now so we can tag a release. Step two is to verify that the new settings are equivalent and merge those.

In the meantime, if you're interested in dealing with step two, please send a pull request. In the future, we prefer that you don't include code solutions with issues, since we can only use code from contributors who have signed the CLA.

cfcommando:
Indeed. I'm waiting on legal to clear the CLA.

dstnbrkr:
:+1: looking forward to your contributions!

Issue #46, opened 2015-03-30
Create an umbrella header?


kastiglione:
Is the absence of an umbrella header an oversight, or part of the framework's opinions?

adamjernst:
Opinions. I dislike umbrella headers generally. (Probably influenced by the fact that we dislike them at Facebook generally...)

jlpiedrahita:
Most system frameworks and 3er party libraries/frameworks do it.

Can you expand on why you think this should not be done? Personally I find it annoying to have to know which file to include for each symbol I need, it's not always obvious.

adamjernst:
Well, I wouldn't be opposed to a pull request I guess. We can always disallow its use internally.

(With the sheer amount of code we have, we find umbrella headers end up bloating compilation times and result in huge numbers of symbols available. But for most smaller projects it's not an issue at all.)

kastiglione:
:+1: I like giving the option.

kastiglione:
This relates to #44.

PR #97, opened 2015-04-10
Add umbrella header

jlpiedrahita:
Related issue: https://github.com/facebook/componentkit/issues/46

### DIFF
--- a/ComponentKit/ComponentKit.h
+++ b/ComponentKit/ComponentKit.h
@@ -39,9 +39,6 @@
 #import <ComponentKit/CKComponentPreparationQueueListener.h>
 #import <ComponentKit/CKComponentPreparationQueueTypes.h>
 #import <ComponentKit/CKComponentProvider.h>
-//Debug
-#import <ComponentKit/CKComponentDebugController.h>
-#import <ComponentKit/CKComponentHierarchyDebugHelper.h>
 //HostingView
 #import <ComponentKit/CKComponentFlexibleSizeRangeProvider.h>
 #import <ComponentKit/CKComponentHostingView.h>
--- a/ComponentKit/ComponentKit.h
+++ b/ComponentKit/ComponentKit.h
@@ -1,0 +1,70 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant 
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+//Accessibility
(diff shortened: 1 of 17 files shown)


ComponentKit/ComponentKit.h
```diff
@@ -0,0 +1,70 @@
...
+#import <ComponentKit/CKComponentConstantDecider.h>
+#import <ComponentKit/CKComponentDataSource.h>
+#import <ComponentKit/CKComponentDataSourceOutputItem.h>
+#import <ComponentKit/CKComponentDeciding.h>
+#import <ComponentKit/CKComponentPreparationQueueListener.h>
+#import <ComponentKit/CKComponentPreparationQueueTypes.h>
+#import <ComponentKit/CKComponentProvider.h>
+//Debug
+#import <ComponentKit/CKComponentDebugController.h>
+#import <ComponentKit/CKComponentHierarchyDebugHelper.h>
```
dstnbrkr: Let's exclude the debug headers from the umbrella header

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
```diff
@@ -11,12 +11,8 @@
 
 #import "InteractiveQuoteComponent.h"
 
-#import <ComponentKit/CKComponentGestureActions.h>
-#import <ComponentKit/CKComponentScope.h>
+#import <ComponentKit/ComponentKit.h>
 #import <ComponentKit/CKComponentSubclass.h>
-#import <ComponentKit/CKCompositeComponent.h>
-#import <ComponentKit/CKOverlayLayoutComponent.h>
-#import <ComponentKit/CKStackLayoutComponent.h>
```
dstnbrkr: I kind of like using the explicit imports in the example project, feels more explanatory

Examples/WildeGuess/WildeGuess/Components/QuoteComponent.h
```diff
@@ -9,7 +9,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#import "CKCompositeComponent.h"
+#import <ComponentKit/CKCompositeComponent.h>
```
dstnbrkr: but if we do convert the example project to using the umbrella header, let's do it 100% (all headers too)

dstnbrkr:
Umbrella header looks good to go once debug classes are removed. Example project either doesn't need conversion or needs total conversion.

jlpiedrahita:
Removed the debug headers (good catch). Regarding the example project it feels neat to use only the umbrella header, IMHO use of component classes alone should be explanatory enough.

dstnbrkr:
Thanks @jlpiedrahita

Issue #107, opened 2015-04-13
Integrating with POP


crylico:
Hi there!

I got a hacky prototype of ComponentKit working with POP animations and would love to bring to the table integrating the two in a more sound way.

The way I'd go about making these changes (which I'm happy to contribute myself when possible) is first making sure that POP adheres to NSCopying.

Then, I'd add a third animation method, like "pop_animationsFromPrevious...." that returns some struct called CKComponentPOPAnimation, possibly a sub-struct of CKComponentAnimation.

Aside from the lack of NSCopying, are there any other underlying conflicts that I may be overlooking?

adamjernst:
This sounds awesome to me!

While we don't want to introduce a dependency on POP in the core, this sounds perfect for an external repo that people could use to "plug in" POP to ComponentKit. I'd love to have this so I can refer people to it when they ask about POP!

Let me know about any changes that need to be made in the core repo to support it.

crylico:
Hey Adam,

I just need `hooks` to be protected instead of private in `CKComponentAnimation.h`.  I opted to subclass it and just add a new constructor.

I've got it working right, but need to implement copying on POPAnimation and subclasses

crylico:
Hey, I didn't even think about that other constructor! I've got it all working fine with no modifications to core project :)

adamjernst:
Sweet!

crylico:
I figured this would be the best place to put this.

 I've made the necessary changes to [POP](https://github.com/facebook/pop) to add `NSCopying`, and the changes have been merged into `master`.

Here's a nice and simple implementation: [CKComponentPOPAnimation](https://github.com/crylico/CKComponentPOPAnimation).

Let me know if anything's out of order and I'd be happy to make any adjustments :)

adamjernst:
Neat! Add a LICENSE to that repo so people can use it.

crylico:
Duh! I completely forgot about license! Good call :)

PR #110, opened 2015-04-13
made hooks protected instead of private to enable subclassing of CKCompo...

crylico:
...nentAnimation to support POP #107


adamjernst:
You should be able to use the already-public constructor that takes a `CKComponentAnimationHooks` instead of making the private field public. Give that a try?

PR #116, opened 2015-04-14
fix bug

andyqee:
when collectionView is loaded in viewDidLoad , the ```scrollViewDidScroll``` selector of UIScrollView delegate will also be triggered more than one time. <br>
In that case, the current contentSize is still (0, 0),  so the ```scrolledToBottomWithBuffer``` will always return 'YES' !  <br>
So the data will be enqueued in unexpected situation, before you actually scroll the view
```
- (void)scrollViewDidScroll:(UIScrollView *)scrollView
{
  if (scrolledToBottomWithBuffer(scrollView.contentOffset, scrollView.contentSize, scrollView.contentInset, scrollView.bounds)) {
    [self _enqueuePage:[_quoteModelController fetchNewQuotesPageWithCount:8]];
  }
}
```


benlodotcom:
Thank @andyqee, could we take the content height check out of scrolledToBottomWithBuffer ?

andyqee:
Thank @benlodotcom for your advice.

Issue #108, opened 2015-04-13
Possible to make CKNetworkImageComponent more general?


epatey:
Currently, `CKNetworkImageComponent` seems needlessly bound to the concept of `NSURL`'s and networking, when the component is really capable of handling any type of image that may need to be delivered asynchronously.

I'd propose renaming:
 `CKNetworkImageComponent` -> `CKAsyncImageComponent`
 `CKNetworkImageDownloading` -> `CKAsyncImageRetrieval`
`(NSURL *)URL` parameter -> `(id)imageIdentifier`

In my case, images come from a variety of sources (i.e. database, file system, etc) - all potentially asynchronously - via ReactiveCocoa signals.

I've written an implementation of `CKNetworkImageDownloading` that does the ReactiveCocoa subscription and disposal by cheating and passing a `RACSignal` where the `NSURL` is specified.

Does making `CKNetworkImageComponent` more general make sense? If people agree, I'd happily make a PR with the change.

adamjernst:
This is a really neat idea!

I like the general idea of a more abstract "async image loader." However I want to avoid scope creep in the core ComponentKit library.

Would you mind implementing this in a separate project (as a Cocoapod that depends on ComponentKit)? That way we could direct people to use that project when they need this functionality.

This would mean there would be duplication between the core `CKNetworkImageComponent` and the network image component in your async-image-loading library, which is unfortunate; but people could simply avoid using `CKNetworkImageComponent` entirely in that case.

epatey:
@adamjernst, thanks for considering the idea. Of course, I'm cool if you don't want to take the PR. That's fine. 

I'm not a fan of the idea of making a new component, though. Since the PR just renames stuff and generates the exact same runtime code, I think it risks being confusing. It doesn't seem worth it.

PR #118, opened 2015-04-15
Rename CKNetworkImage classes to CKAsyncImage in order to be more general.

epatey:
Here's a stab at the change I discussed in #108.

It's 99% renames of classes and parameters.


ComponentKit/Components/CKAsyncImageComponent.mm
```diff
@@ -44,31 +44,36 @@ + (instancetype)newWithURL:(NSURL *)url
...
-                   cropRect:(CGRect)cropRect
+- (instancetype)initWithIdentifier:(id)identifier
+                      defaultImage:(UIImage *)defaultImage
+                   imageDownloader:(id<CKAsyncImageRetrieval>)imageDownloader
+                         scenePath:(id)scenePath
+                          cropRect:(CGRect)cropRect
 {
   if (self = [super init]) {
-    _url = [url copy];
+    if ([identifier respondsToSelector:@selector(copyWithZone:)]) {
```
epatey: This is the only actual change to the functioning of the code. Previously, it was always appropriate to use `NSCopying` since the param was an `NSURL`. With my change, we only copy if it is supported by the `identifier` object.

adamjernst:
See discussion on #108.

PR #119, opened 2015-04-15
[CKComponentViewClass] removed deprecated constructor

majak:


adamjernst:
Looks like a test is failing.

PR #121, opened 2015-04-16
ComponentKitTestLib.podspec requires some updates to be usable (Cocoapods)

itsthejb:
Hi all,

Noticed that `ComponentKitTestLib` is available, but not currently in a usable state for Cocoapods. I assume this is coming (as a subspec?) eventually, but in the meantime a few minor updates make this usable until it's publicly available. 

For the benefit of other users, the library can be installed via Cocoapods with:

```
pod 'ComponentKitTestLib', :podspec => 'https://raw.githubusercontent.com/eyeem/componentkit/ComponentKitTestLib.podspec/ComponentKitTestLib/ComponentKitTestLib.podspec'
```

...which links to the main repo, but makes use of the corrected podspec in this fork.

Thanks!

### DIFF
--- a/ComponentKitTestLib/ComponentKitTestLib.podspec
+++ b/ComponentKitTestLib/ComponentKitTestLib.podspec
@@ -1,13 +1,13 @@
 Pod::Spec.new do |s|
   s.name             = "ComponentKitTestLib"
-  s.version          = "0.1.0"
+  s.version          = "0.10"
   s.summary          = "A React-inspired view framework for iOS"
   s.homepage         = "https://componentkit.com"
   s.license          = 'BSD'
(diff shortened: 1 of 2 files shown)


itsthejb:
Travis build fails on `pod install`, but works fine locally...

dstnbrkr:
Updating the Podfile.lock to pick up the change from 0.1.0 to 0.10 should fix the build.

itsthejb:
Ok let, me do that tomorrow.

itsthejb:
Still header import issues on Travis. Unless there's a quick solution I would be fine to close this and at least leave it for posterity for other users. One can at least use the test lib with this fix, even if it's not ready for the public repo...?

dstnbrkr:
The subspec route is a good option too, although last time I looked it wasn't obvious to me how to make it work.

PR #126, opened 2015-04-17
Remove unused `objectForUUID:` API

adamjernst:


adamjernst:
Let me think about these a bit first.

Issue #127, opened 2015-04-17
Question - ComponentKit Accessibility Support


aaronschubert0:
Hi, I've implemented various CKLabel/TextComponents and have then turned on the voice over accessibility inspector expecting the accessibility inspector to recognise each element out of the box. However the accessibility inspector doesn't recognise any component, how can this be fixed?

ocrickard:
I didn't enable it by default in our text components.  I'll put up a diff in one sec to fix/update the behavior.

aaronschubert0:
Awesome, thanks for the quick turnaround!

PR #130, opened 2015-04-17
Enable accessibility in CKLabelComponent by default

ocrickard:
Enables accessibility by default in CKLabelComponent, and provides an example on how to do more advanced accessibility setups for CKTextComponent.

Should fix https://github.com/facebook/componentkit/issues/127

### DIFF
--- a/ComponentTextKit/CKLabelComponent.mm
+++ b/ComponentTextKit/CKLabelComponent.mm
@@ -22,7 +22,7 @@
           [CKTextComponent
            newWithTextAttributes:textKitAttributes(attributes)
            viewAttributes:std::move(copiedMap)
-           accessibilityContext:{}]];
+           accessibilityContext:{.isAccessibilityElement = @(YES)}]];
 }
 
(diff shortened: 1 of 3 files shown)


ComponentTextKit/CKLabelComponent.mm
```diff
@@ -17,12 +17,12 @@ @implementation CKLabelComponent
 + (instancetype)newWithLabelAttributes:(const CKLabelAttributes &)attributes
                         viewAttributes:(const CKViewComponentAttributeValueMap &)viewAttributes
 {
-  CKViewComponentAttributeValueMap copiedMap = viewAttributes;
+  const CKViewComponentAttributeValueMap copiedMap = viewAttributes;
```
adamjernst: Don't add `const` here or the copied map can't be moved (I think).

adamjernst: Don't add const here or the copied map can't be moved (I think).

ocrickard: Yeah, good point this degrades to a copy instead of a move if it's const.

ocrickard:
Updated diff.

PR #132, opened 2015-04-18
Use new context for reload after updateContextAndEnqueueReload:

smeis:
Updates the context of all the items before the reload, currently the old context is used for the reload.

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewDataSource.mm
@@ -223,14 +223,12 @@ static NSString *const kReuseIdentifier = @"com.component_kit.collection_view_da
   if (!_processingChangeset && [_changesetQueue count]) {
     ck_changeset_applicator_t headChangeset = _changesetQueue[0];
     [_changesetQueue removeObjectAtIndex:0];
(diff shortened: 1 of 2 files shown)


smeis:
I've added another commit. UICollectionViews' `reloadItemsAtIndexPaths` doesn't work (well) with a large set of updates so in these cases its better to use `reloadData`. 

I moved to call to `performBatchUpdates` from `applyNextChangeSet` to `applyChangesetToCollectionView` because calling `reloadData` from within the block didn't seem to work well. I couldn't think of a reason for this to be problematic but please let me know if I've missed something so I can change it.

benlodotcom:
@smeis Sorry for the delay coming back to you, this is a very good catch !

Would you mind splitting the PR in two, one for using the new context and one for the reload ? Out of curiosity could you detail what you mean by `reloadItemsAtIndexPaths` doesn't work well with a large set of updates ? Do you mean this from a performance standpoint ?

smeis:
@benlodotcom Will do! I'll close this one and create two separate PR's.

Regarding `reloadItemsAtIndexPaths`, I ran some tests in an app I'm working on and saw a huge decrease in performance when updating a lot of cells at the same time when the update changed the size of the cells.

PR #142, opened 2015-04-22
Add a global identifier to CKComponentScopeFrame.

adamjernst:
This allows comparing scope frames across versions of the scope frame tree.


adamjernst:
Reopening one commit at a time.

Issue #87, opened 2015-04-09
Don't synchronously unmount components after -updateState


cfcommando:
`CKComponentActionSend` can crash if called after `updateState` because updating state may unmount the component synchronously. 

See: #86 

@adamjernst says:
> This is really a bug with updateState that we plan to address shortly. It shouldn't synchronously unmount; instead that should be deferred to the end of the run loop.

facebook-github-bot-9:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

adamjernst:
This is fixed with the transactional data source!

Issue #15, opened 2015-03-26
Subtle race condition with state vs model updates


adamjernst:
It's theoretically possible for a race condition to occur if a state update occurs between two model updates. This is so rare we haven't even noticed it when using ComponentKit in Facebook's News Feed, but we should fix it.

My plan is to implement a STM-like system for computing updates to the `CKComponentLifecycleManager`'s state to resolve this issue.

kastiglione:
:gem:

adamjernst:
This also addresses the problem of synchronously updating the size range on rotation.

facebook-github-bot-9:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

ocrickard:
This should have been resolved by the transactional datasource, right? Let's close this.

PR #145, opened 2015-04-23
Value objects for the transactional data source

adamjernst:
This work is aimed at creating a new data source to replace the current implementation. The goals are:

- Fix the race conditions described in #15
- Enable batching state updates to the end of the run loop as described in #87
- Enable asynchronous state updates in a less hacky way
- Allow all operations on a data source to be performed synchronously or asynchronously in a safe manner (right now, insertions are always async)

These objects are merely immutable value objects or announcers, not the actual data source implementation. That will come later.


adamjernst:
Ugh github whyyyy

PR #147, opened 2015-04-23
Transactional component data source

adamjernst:
A new data source to replace the current implementation. The goals are:

- Fix the race conditions described in #15
- Enable batching state updates to the end of the run loop as described in #87
- Enable asynchronous state updates in a less hacky way
- Allow all operations on a data source to be performed synchronously or asynchronously in a safe manner (right now, insertions are always async)


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
```diff
@@ -0,0 +1,28 @@
...
+@property (nonatomic, copy, readonly) NSIndexSet *removedSections;
+@property (nonatomic, copy, readonly) NSDictionary *movedIndexPaths;
+@property (nonatomic, copy, readonly) NSIndexSet *insertedSections;
+@property (nonatomic, copy, readonly) NSSet *insertedIndexPaths;
+
+/**
+ Array of NSDictionary containing all non-nil userInfo objects from the CKTransactionalComponentDataSourceChangeset
+ objects that caused this change. (Multiple changesets may result in only a single AppliedChanges.)
+ */
+@property (nonatomic, copy, readonly) NSArray *userInfos;
```
ocrickard: Not a huge fan of having to do this.  Curious what the situations are in practice in which multiple changesets result in only a single applied change?

adamjernst: Sure: if multiple changesets were enqueued asynchronously, and then a changeset is applied synchronously, we must flatten all the changesets into one and apply it synchronously. Otherwise indexes will be out of order.

I'm not a huge fan either, but given that it's restricted to userInfo only, I make an exception :)

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceListener.h
```diff
@@ -0,0 +1,30 @@
...
+
+@class CKTransactionalComponentDataSource;
+@class CKTransactionalComponentDataSourceAppliedChanges;
+@class CKTransactionalComponentDataSourceState;
+
+@protocol CKTransactionalComponentDataSourceListener
+
+/**
+ Announced on the main thread when the data source has just updated its state.
+ @param dataSource The sending data source; its state property now returns an updated object.
```
ocrickard: Hmm, so from a client perspective what's the contract on when this is called?  Why not vend the new state object here?

adamjernst: You could vend the new state object, but it's redundant since you can fetch it from the data source.

We could //not// announce the datasource, but I'm not a fan of that because if some object is listening to multiple data sources, it can't distinguish the announcements it's receiving. We had this exact issue when @modocache tried to remove the "store" parameter from the Connection Store's announcements.

So in sum: this is the right thing because it matches Connection Store.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
ocrickard: Ohhh, yeah, that's tough.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
```diff
@@ -0,0 +1,67 @@
...
+
+@implementation CKTransactionalComponentDataSourceAppliedChanges
+
+- (instancetype)initWithUpdatedIndexPaths:(NSSet *)updatedIndexPaths
+                        removedIndexPaths:(NSSet *)removedIndexPaths
+                          removedSections:(NSIndexSet *)removedSections
+                          movedIndexPaths:(NSDictionary *)movedIndexPaths
+                         insertedSections:(NSIndexSet *)insertedSections
+                       insertedIndexPaths:(NSSet *)insertedIndexPaths
+                                userInfos:(NSArray *)userInfos
```
benlodotcom: I'm not a big fan of userInfos, could we restrict this to IDs that are returned when a changeset is enqueued and then communicated at the output of the system. This way any metadata management can be done outside of the system itself.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceItem.h
```diff
@@ -0,0 +1,31 @@
...
+- (const CKComponentLayout &)layout;
+
+/** The model used to compute the layout */
+- (id)model;
+
+/** Announces a given event to all controllers in the layout. */
+- (void)announceEventToControllers:(CKComponentAnnouncedEvent)event;
+
+/** Computes the requested bounds animation given a previous version of the same item. */
+- (CKComponentBoundsAnimation)boundsAnimationFromPreviousItem:(CKTransactionalComponentDataSourceItem *)previousItem;
```
benlodotcom: Could we keep this item as a pure value object and have pure functions (static C function or class methods) to do the two operations above ?

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
adamjernst: I'd actually like to argue for userInfos.

- They allow you to pipe data through and have the data source hang onto it, freeing you from needing to maintain a lookaside table.
- They are idiomatic to Objective-C.
- It matches our internal "connection store" class.

We can debate this in person. :)

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceItem.h
adamjernst: Sure! Let me take a stab at it.

adamjernst:
@benlodotcom Done.

PR #148, opened 2015-04-23
Fix a bug that hash() doesn’t generate unique value

weijentu:
Hi there,
I just found a potential but probably minor issue. I'm assuming hash() function for CKSizeRange should always be able to generate a unique value as a fingerprint on iPhone simulator.

A.    NSLog(@"%ld", CKSizeRange(CGSizeMake(50, 90), CGSizeMake(60, 100)).hash());
B.    NSLog(@"%ld", CKSizeRange(CGSizeMake(50, 60), CGSizeMake(90, 100)).hash());
C.    NSLog(@"%ld", CKSizeRange(CGSizeMake(50, 91), CGSizeMake(60, 100)).hash());

For example, case A, B and C will generate a output below in console:
2015-04-23 16:14:37.286 WildeGuess[16181:1546643] 4636737291354636288
2015-04-23 16:14:37.287 WildeGuess[16181:1546643] 4636737291354636288
2015-04-23 16:14:37.287 WildeGuess[16181:1546643] 4636737291354636288

Contrary, output for iPhone 5 device will look like the below:
2015-04-23 16:16:29.911 WildeGuess[896:208289] 1008150498
2015-04-23 16:16:29.912 WildeGuess[896:208289] -1940518087
2015-04-23 16:16:29.913 WildeGuess[896:208289] -31152360

Therefore, I made a minor change to have unique hash value output on both iPhone simulator and device.

The following is a summary of my change. 
- Each element has to be NSUInteger type in CKIntegerArrayHash()
- Bitwise left shift according to the size of NSUInteger type

Cheers,

Allen

### DIFF
--- a/ComponentKit/Core/CKDimension.mm
+++ b/ComponentKit/Core/CKDimension.mm
@@ -131,12 +131,11 @@ NSString *CKSizeRange::description() const
 }
 size_t CKSizeRange::hash() const
 {
-  std::hash<CGFloat> hasher;
   NSUInteger subhashes[] = {
-    hasher(min.width),
-    hasher(min.height),
-    hasher(max.width),
-    hasher(max.height)
+      NSUInteger(min.width),
+      NSUInteger(min.height),
+      NSUInteger(max.width),
+      NSUInteger(max.height)
   };
   return CKIntegerArrayHash(subhashes, CK_ARRAY_COUNT(subhashes));
 }
--- a/ComponentKit/Utilities/CKInternalHelpers.mm
+++ b/ComponentKit/Utilities/CKInternalHelpers.mm
@@ -35,7 +35,7 @@ NSUInteger CKIntegerArrayHash(const NSUInteger *subhashes, NSUInteger count)
 {
   NSUInteger result = subhashes[0];
   for (int ii = 1; ii < count; ++ii) {
-    result = std::hash<unsigned long long>()((((unsigned long long)result) << 32 | subhashes[ii]));
+    result = std::hash<unsigned long long>()((((unsigned long long)result) << sizeof(NSUInteger) | subhashes[ii]));
   }
   return result;
 }


ComponentKit/Core/CKDimension.mm
```diff
@@ -131,12 +131,11 @@ _Range intersect(const _Range &other) const
...
-  std::hash<CGFloat> hasher;
   NSUInteger subhashes[] = {
-    hasher(min.width),
-    hasher(min.height),
-    hasher(max.width),
-    hasher(max.height)
+      NSUInteger(min.width),
+      NSUInteger(min.height),
+      NSUInteger(max.width),
+      NSUInteger(max.height)
```
adamjernst: Are you sure these changes are necessary? I agree the other one probably is, but `hasher` returns a `std::size_t` which should promote to `NSUInteger` without a problem.

Casting `CGFloat` to `NSUInteger` doesn't sit well with me (even though it probably works).

adamjernst:
Nice catch.

ComponentKit/Utilities/CKInternalHelpers.mm
```diff
@@ -35,7 +35,7 @@ NSUInteger CKIntegerArrayHash(const NSUInteger *subhashes, NSUInteger count)
 {
   NSUInteger result = subhashes[0];
   for (int ii = 1; ii < count; ++ii) {
-    result = std::hash<unsigned long long>()((((unsigned long long)result) << 32 | subhashes[ii]));
+    result = std::hash<unsigned long long>()((((unsigned long long)result) << sizeof(NSUInteger) | subhashes[ii]));
```
adamjernst: Should this be `NSUInteger` or `unsigned long long`? I honestly don't know.

ComponentKit/Core/CKDimension.mm
weijentu: Hi Adamjernst,
It does have little difference in precision. Let’s narrow down our comparison in iPhone simulator with the following two cases.
NSLog(@"%ld", CKSizeRange(CGSizeMake(50, 90), CGSizeMake(60, 100)).hash());
NSLog(@"%ld", CKSizeRange(CGSizeMake(50, 91), CGSizeMake(60, 100)).hash());

The only difference above is the min size range with 90 or 91 in width. Then let’s try our first configuration A:
- Have hasher() to construct subhashes
- Do bit shift with sizeof(unsigned long long)

And we get the following results and can’t differentiate our test samples.
2015-04-24 15:18:37.060 WildeGuess[2330:230394] -3577828428969148416
2015-04-24 15:18:37.060 WildeGuess[2330:230394] -3577828428969148416

So let’s try the second configuration B:
- Have NSUInteger to construct subhashes
- Do bit shift with sizeof(NSUInteger)

And we get the following results and can differentiate our test samples.
2015-04-24 15:20:54.276 WildeGuess[2372:237574] 844774500
2015-04-24 15:20:54.277 WildeGuess[2372:237574] 844840036

Also I found another thing. For example, if we search keyword "CKIntegerArrayHash" in Xcode, we can have two groups of results. Results in ComponentKit adopt NSUInteger to construct subhashes (ex. CKComponentDataSourceInputItem); however, results in ComponentTextKit adopt std::hash to construct subhashes (ex. CKTextKitRendererCache).

My assumption is that NSUInteger is more native to compiler and Apple pays more efforts on it for cast or internal conversion. In simple way to describe this, you can say I’m thinking CGFloat->NSInteger->std::hash is potentially better than CGFloat->std::hash directly. In order to avoid the situation of mixing up C++/C and type cast multiple times, I created another version of CKIntegerArrayHash in pure Swift and do another comparison . I paste the code here:

```
func CKIntegerArrayHash <H: Hashable, S: SequenceType where S.Generator.Element == H>(sequence: S, count: Int) -> Int {
    var result = 0
    for element in sequence {
        result = result.hashValue << sizeof(Int) | element.hashValue
    }
    return result
}
```
The result I got is as same as to the results of configuration B. 
Therefore, I chose to cast NSUInteger and sizeof(NSUInteger) to remind me I’m dealing with NSUInteger.

Cheers,

Allen

ComponentKit/Utilities/CKInternalHelpers.mm
weijentu: Yeah, I think you are right about it. It’s hard to say which one is better. In results, they can get same output here. The reason I chose sizeof(NSUInteger) just because of reminding me each element is NSUInteger. However, I have to honestly admit that sizeof(unsigned long long) looks much better when having the first sight of it :)

ComponentKit/Core/CKDimension.mm
weijentu: I just reviewed the original way and I think the issue could probably be that we throw a subhashes with huge values as elements and do four times of 8 bits shift left in CKIntegerArrayHash.

For example, let's say there is a width value which is 375 here. Next have it via hasher and get a new value 1136361472.

```
std::hash<float> hasher;
hasher(CGFloat(375)); // <--- The value is 1136361472
```

If the value is the first element of subhashes, let's imagine how big the final result will be after doing four times of bit shift left with 8 bits, especially comparing to NSUIntegerMax which is 4294967295.

Therefore, to cast geometry from CGFloat into NSUInteger as elements in subhashes can avoid to dealing with a value which is huge already; however, one thing has to be noted that the count of subhashes has to be under a certain limit. Ex. There are 10 elements in subhashes. The same problem will occur again even values are small, like 375 etc. 

For now, I guess current CKIntegerArrayHash can handle well because I didn’t find other subhashes with a huge value as the first element in componentKit. However, it could be better to take the variable "count" of CKIntegerArrayHash into account since it’s in function already.

Have a nice weekend!!

Allen

ComponentKit/Utilities/CKInternalHelpers.mm
darknoon: Doesn't left shifting the previous result completely destroy all of the results you computed in there. From testing in lldb, you just get the lsb smeared across all bits.

adamjernst:
We're working on an alternate fix for this issue — thanks for bringing to our attention!

Issue #153, opened 2015-04-28
Issue with appearance events when a visible indexPath is reloaded


benlodotcom:
When reloading a cell on screen :
- The collection view first send appearance events for the newly created cell.
- It then send disappearance events for the old cell.

If appearance events are funneled directly from the collection view (`
- collectionView:willDisplayCell:forItemAtIndexPath:`, `
- collectionView:didEndDisplayingCell:forItemAtIndexPath:`) to the corresponding lifecycle manager holding the component tree, then a disappearance event will automatically be sent after the appearance event while the corresponding component is still on screen. This situation should be dealt with in the controller layer.

*TODO*
- Put an example in the sample app that shows how it can be done.
- Maybe add an object that can be used to achieve those tasks and expose the UUID associated to a lifecycle manager.

ocrickard:
@benlodotcom is this fixed?

facebook-github-bot-9:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

ocrickard:
Ultimately this will also be resolved by https://github.com/facebook/componentkit/pull/373. Closing this out.

PR #154, opened 2015-04-28
[CDS] Fix appearance tracking

### LINKED ISSUE #129: Array out of bounds in announceDidDisappearForItemAtIndexPath
Currently, if you ask dataSource for not existing indexPath, it will crash since there is direct access to NSArray[index].
What happens for me is that I delete a cell and receive didEndDisplayingCell, the app crashes on announceDidDisappearForItemAtIndexPath because the model was deleted.
It might be better to simply return nil I guess.
Minor thing since I can actually check in my model, but still...

benlodotcom:
Fix #129, indexPath is not a reliable thing to use for disappearance announcements on deletion (either the index would be out of bounds if the last item is deleted or it will point to an invalid item).

Calling `- (void)announceWillAppearForItemInCell:(UICollectionViewCell *)cell atIndexPath:(NSIndexPath *)indexPath` and `- (void)announceDidDisappearForItemInCell:(UICollectionViewCell *)cell atIndexPath:(NSIndexPath *)indexPath` guarantees the announcement will be dispatched to the component tree associated to the cell. 

It is however non opinionated in the sense that it won't do deduplication of events, or check for imbalanced calls. 
For instance while reloading a cell on screen, the collection view first send appearance events for the newly created cell and then send disappearance events for the old cell, this situation should be handled properly in the controller layer. #153 has been opened to track these use cases and provide examples/code to deal with them.

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewDataSource.h
+++ b/ComponentKit/DataSources/CKCollectionViewDataSource.h
@@ -91,16 +91,16 @@ typedef void(*CKCellConfigurationFunction)(UICollectionViewCell *cell, NSIndexPa
 - (CGSize)sizeForItemAtIndexPath:(NSIndexPath *)indexPath;
 
 /**
- Sends -componentTreeWillAppear to all CKComponentControllers for the given indexPath.
- Call this from -collectionView:willDisplayCell:forItemAtIndexPath:
+ Sends -componentTreeWillAppear to all CKComponentControllers for the given cell.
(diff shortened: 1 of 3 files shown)


adamjernst:
Nice. Since we're now not even using the index path parameter from the `announce...` methods, mind removing it?

benlodotcom:
Kept it for API symmetry but totally fine removing it.

adamjernst:
I do like the symmetry but... #yagni

PR #159, opened 2015-04-29
Use new context in updateContextAndEnqueueReload

smeis:
Updates the context of all the items before the reload, currently the old context is used for the reload.


smeis:
Split from #132 @benlodotcom

benlodotcom:
Thanks ! @smeis

PR #160, opened 2015-04-29
Use reloadData for large updates

smeis:
Split of from #132 as per request @benlodotcom.

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewDataSource.mm
@@ -288,9 +288,9 @@ static void applyChangesetToCollectionView(const Output::Changeset &changeset, U
   };
   
   changeset.enumerate(sectionsEnumerator, itemEnumerator);
-  if(itemUpdateIndexPaths.count > 30){
+  if (itemUpdateIndexPaths.count > 30){
     [collectionView reloadData];
-  }else{
+  } else {
     [collectionView performBatchUpdates:^{
       if (itemRemovalIndexPaths.count > 0) {
         [collectionView deleteItemsAtIndexPaths:itemRemovalIndexPaths];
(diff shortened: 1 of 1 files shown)


ComponentKit/DataSources/CKCollectionViewDataSource.mm
```diff
@@ -290,14 +288,20 @@ static void applyChangesetToCollectionView(const Output::Changeset &changeset, U
...
-    [collectionView deleteItemsAtIndexPaths:itemRemovalIndexPaths];
-  }
-  if (itemUpdateIndexPaths.count > 0) {
-    [collectionView reloadItemsAtIndexPaths:itemUpdateIndexPaths];
-  }
-  if (itemInsertionIndexPaths.count > 0) {
-    [collectionView insertItemsAtIndexPaths:itemInsertionIndexPaths];
+  if(itemUpdateIndexPaths.count > 30){
+    [collectionView reloadData];
+  }else{
```
adamjernst: nit: spacing

smeis:
Thanks @adamjernst, fixed!

benlodotcom:
Can you give more details regarding why reloadData was needed ?

adamjernst:
I think this is a fine optimization, but we want to keep the "sample" data source simple. I think the best thing for you to do is to fork the data source and put this optimization there!

smeis:
I'm sorry for not getting back to this sooner @adamjernst @benlodotcom , I'm unusually busy right now. I had been working on building a sample project to clarify the problem but hadn't gotten around to finishing it and I don't really have time to finish it right now. Especially since it might take a lot more digging since I started suspecting this might only show up in pretty specific cases, so your suggestion is probably the best for now.

The problem was that when I reloaded the dataset to resize the components the app stalled the main thread. It spent a huge amount of time in _endAnimations as you can see in the screenshot below. I put in the call to reloadData because I remembered UICollectionView could sometimes crash if you enqueued too many updates at once. I expect to have more time to look at this in about a month or so.




ck-stall-screen-modified

PR #170, opened 2015-05-03
fix bug in Example project

andyqee:
when collectionView is loaded in viewDidLoad , the scrollViewDidScroll selector of UIScrollView delegate will also be triggered more than one time. <br>

In that case, the current contentSize is still (0, 0), so the scrolledToBottomWithBuffer will always return 'YES' !<br> 

So the data will be enqueued in unexpected situation, before you actually scroll the view
```
- (void)scrollViewDidScroll:(UIScrollView *)scrollView
{
  if (scrolledToBottomWithBuffer(scrollView.contentOffset, scrollView.contentSize, scrollView.contentInset, scrollView.bounds)) {
    [self _enqueuePage:[_quoteModelController fetchNewQuotesPageWithCount:8]];
  }
}
```


facebook-github-bot-9:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #179, opened 2015-05-05
fix hostingView cannot trigger resize issue

smalllixin:
Hi
 I found a issue that hosting view cannot receive invalidateSize message.

 Create a hosting view inside view controller. 
 And resize the frame of itself after layout the components.
```Objective-C
- (void)componentHostingViewDidInvalidateSize:(CKComponentHostingView *)hostingView {
    CGSize s = [hostingView sizeThatFits:CGSizeMake([UIScreen mainScreen].bounds.size.width, 8000)];
    hostingView.frame = (CGRect){.origin=hostingView.frame.origin, .size=s};
    [hostingView setNeedsLayout];//This is necessary because root view in hosting view should be modified too.
}
```
Inside hosting view the component is some simple labels.
StackLayout(Vertical)
   children:[label,label]

The problem is when I assign a new model to the hosting view. The invalidate size delegate do not triggered. And I found the reason is lifecyclemanager use self.bounds as size range (This is not the right one, and could be changed after update the hostingview's bounds )
```Objective-C
const CGRect bounds = self.bounds;
CKComponentLifecycleManagerState state = [_lifecycleManager prepareForUpdateWithModel:_model constrainedSize:CKSizeRange(bounds.size, bounds.size) context:_context];
```
The solutation is use the correct size range simply.

### DIFF
--- a/ComponentKit/HostingView/CKComponentHostingView.mm
+++ b/ComponentKit/HostingView/CKComponentHostingView.mm
@@ -140,7 +140,8 @@
   }
 
   const CGRect bounds = self.bounds;
-  CKComponentLifecycleManagerState state = [_lifecycleManager prepareForUpdateWithModel:_model constrainedSize:CKSizeRange(bounds.size, bounds.size) context:_context];
+  CKSizeRange constrainedSize = [_sizeRangeProvider sizeRangeForBoundingSize:bounds.size];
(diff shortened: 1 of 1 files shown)


adamjernst:
Sorry for the long delay! Nice find. Fixed by #283.

PR #184, opened 2015-05-08
Add CKComponentSubclass to umbrella header

MarcoSero:


adamjernst:
Thanks for the fix — however we specifically don't want to do this since we want these to stay private, available only to the subclasses that actually need them.

UIKit does the same thing with UIGestureRecognizerSubclass.h.

PR #185, opened 2015-05-08
Allow null action to be passed to `CKComponentActionAttribute`

adamjernst:
This should minimize the places we need to do conditional attributes.

http://componentkit.org/docs/avoid-push-back.html

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -147,10 +147,8 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
     {@selector(setSelected:), @(selected)},
     {@selector(setEnabled:), @(enabled)},
     {@selector(setAccessibilityIdentifier:), accessibilityConfiguration.accessibilityIdentifier},
+    CKComponentActionAttribute(action, UIControlEventTouchUpInside),
   });
-  if (action) {
-    attributes.insert(CKComponentActionAttribute(action, UIControlEventTouchUpInside));
-  }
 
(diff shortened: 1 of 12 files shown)


ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -147,10 +147,8 @@ + (instancetype)newWithTitles:(const std::unordered_map<UIControlState, NSString
     {@selector(setSelected:), @(selected)},
     {@selector(setEnabled:), @(enabled)},
     {@selector(setAccessibilityIdentifier:), accessibilityConfiguration.accessibilityIdentifier},
+    CKComponentActionAttribute(action, UIControlEventTouchUpInside),
   });
-  if (action) {
-    attributes.insert(CKComponentActionAttribute(action, UIControlEventTouchUpInside));
-  }
```
adamjernst: This is the main motivation. No more conditional attributes!

ComponentKitTests/CKComponentActionTests.mm
```diff
@@ -0,0 +1,151 @@
...
+  }
+  return c;
+}
+
+- (void)testAction:(CKComponent *)sender context:(id)context
+{
+  _block(sender, context);
+}
+
+@end
```
modocache: Maybe DRY up this "test action component"? Seems like you've written it twice, in both CKComponentActionAttributeTests and CKComponentActionTests.

ComponentKit/Components/CKButtonComponent.mm
ocrickard: Seems like a reasonable motivation

PR #193, opened 2015-05-14
[WIP] CKTransactionalCollectionViewDataSource

benlodotcom:


ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
```diff
@@ -0,0 +1,41 @@
...
+ Doing so this reverses the traditional approach for a `UICollectionViewDataSource`. Usually the controller layer will *tell* the `UICollectionView` to update and
+ then the `UICollectionView` *ask* the datasource for the data. Here the model is  more Reactive, from an external prospective, the datasource is *told* what
+ changes to apply and then *tell* the collection view to apply the corresponding changes.
+ */
+@interface CKCollectionViewTransactionalDataSource : NSObject
+
+- (instancetype)initWithCollectionView:(UICollectionView *)collectionView
+                     componentProvider:(Class<CKComponentProvider>)componentProvider
+                               context:(id<NSObject>)context
+                        itemsSizeRange:(const CKSizeRange &)itemsSizeRange;
```
adamjernst: > Should we make CKTransactionalComponentDataSourceConfiguration be only the configuration related to component generation (which it is right now) and rename it accordingly so that it is not specific to the CDS (e.g: CKComponentGenerationConfiguration, CKTransactionalDataSourceComponentGenerationConfiguration) ?

Probably. We'll likely want to use it in the HostingView too (although not sure what the size range will mean there).

For now, can you change this constructor to take a `CKTransactionalComponentDataSourceConfiguration`, but don't rename it just yet?

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
```diff
@@ -0,0 +1,41 @@
...
+
+- (void)applyChangeset:(CKTransactionalComponentDataSourceChangeset *)changeset
+                  mode:(CKTransactionalComponentDataSourceMode)mode
+              userInfo:(NSDictionary *)userInfo;
+
```
adamjernst: You'll probably want `updateConfiguration:mode:userInfo:` and `reloadWithMode:userInfo:` here, proxied through to the underlying data source?

adamjernst:
Great, let's knock out the TODOs and land this.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -0,0 +1,120 @@
...
+    CKTransactionalComponentDataSourceConfiguration *datasourceConfiguration = [[CKTransactionalComponentDataSourceConfiguration alloc] initWithComponentProvider:componentProvider
+                                                                                                                                                          context:context
+                                                                                                                                                        sizeRange:itemsSizeRange];
+    _componentDataSource = [[CKTransactionalComponentDataSource alloc] initWithConfiguration:datasourceConfiguration];
+    [_componentDataSource addListener:self];
+    
+    _collectionView = collectionView;
+    _collectionView.dataSource = self;
+    [_collectionView registerClass:[CKCollectionViewDataSourceCell class] forCellWithReuseIdentifier:kReuseIdentifier];
+  }
```
darknoon: Why do we need to use CKCollectionViewDataSourceCell / CKComponentRootView here, instead of being able to mount directly in the collection cell's root view?

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
```diff
@@ -0,0 +1,60 @@
...
+/**
+ @return The layout size of the component tree at a certain indexPath. Use this to access the component sizes for instance in a
+ `UICollectionViewLayout(s)` or in a `UICollectionViewDelegateFlowLayout`.
+ */
+- (CGSize)sizeForItemAtIndexPath:(NSIndexPath *)indexPath;
+
+/** See `CKTransactionalComponentDataSource` */
```
adamjernst: `@see` perhaps

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
```diff
@@ -0,0 +1,60 @@
...
+ Designated initializer
+ 
+ @param collectionView The collectionView is held strongly and its datasource property will be set to the receiver.
+ */
+- (instancetype)initWithCollectionView:(UICollectionView *)collectionView
+                         configuration:(CKTransactionalComponentDataSourceConfiguration *)configuration;
+
+/** 
+ Applies a changeset either synchronously or asynchronously to the collection view.
+ If a synchronous changeset is applied while asynchronous changesets are still pending or being applied, then all the pending changesets are (re)applied synchronously
```
adamjernst: Well they're not **re**-applied.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -0,0 +1,146 @@
...
+}
+
+#pragma mark - Changeset application
+
+- (void)applyChangeset:(CKTransactionalComponentDataSourceChangeset *)changeset
+                  mode:(CKTransactionalComponentDataSourceMode)mode
+              userInfo:(NSDictionary *)userInfo
+{
+  [_componentDataSource applyChangeset:changeset
+                                  mode:CKTransactionalComponentDataSourceModeSynchronous
```
adamjernst: Wrong mode

ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
```diff
@@ -0,0 +1,151 @@
...
+  self = [super init];
+  if (self) {
+    _scopeIdentifierToAttachedViewMap = [NSMutableDictionary dictionary];
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  for (UIView *attachedView in _scopeIdentifierToAttachedViewMap.objectEnumerator) {
```
adamjernst: Another option — instead of punting to the main thread here, you could require explicit main thread invalidation before dealloc.

adamjernst:
Neat!

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
benlodotcom: Good catch was doing some testing with this

ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
```diff
@@ -0,0 +1,170 @@
...
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  NSEnumerator *viewsEnumerator = _scopeIdentifierToAttachedViewMap.objectEnumerator;
+  void(^deallocBlock)(void) = ^{
+    for (UIView *attachedView in viewsEnumerator) {
+      [self _detachComponentLayoutFromView:attachedView];
```
ocrickard: We probably shouldn't capture self in a scope that will live longer than the life of dealloc.  Can we use a static C function instead?

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
```diff
@@ -0,0 +1,60 @@
...
+ apply the corresponding UI changes to the collection view leveraging automatically view reuse.
+ 
+ Doing so this reverses the traditional approach for a `UICollectionViewDataSource`. Usually the controller layer will *tell* the `UICollectionView` to update and
+ then the `UICollectionView` *ask* the datasource for the data. Here the model is  more Reactive, from an external prospective, the datasource is *told* what
+ changes to apply and then *tell* the collection view to apply the corresponding changes.
+ */
+@interface CKCollectionViewTransactionalDataSource : NSObject
+
+/**
+ Designated initializer
```
adamjernst: `NS_DESIGNATED_INITIALIZER`?

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -0,0 +1,146 @@
...
+{
+  CKCollectionViewDataSourceCell *cell = [_collectionView dequeueReusableCellWithReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
+  CKTransactionalComponentDataSourceItem *item = [_currentState objectAtIndexPath:indexPath];
+  [_attachController attachComponentLayout:item.layout withScopeIdentifier:item.scopeRoot.globalIdentifier toView:cell.rootView];
+  return cell;
+}
+
+- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView
+{
+  return _currentState ? [_currentState numberOfSections] : 0;
```
adamjernst: no need for the conditional, nil messaging will return 0

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -0,0 +1,146 @@
...
+}
+
+- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section
+{
+  return _currentState ? [_currentState numberOfObjectsInSection:section] : 0;
```
adamjernst: ditto

ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
adamjernst: Amen.

adamjernst:
Just what @ocrickard  noted. Otherwise, great.

ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
benlodotcom: clowntown, fixing

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
benlodotcom: Yeah, this is on purpose to make it explicit that 0 is a default value we choose. Can change if you want, will have to fix the way we deal with the initial state anyway.

benlodotcom:
- Added NS_DESIGNATED_INITIALIZER
- Regarding main thread deallocation I just made this object main thread affined for now, there is no good reason to pass it around multiple threads and actually accessing the internal datastructure that holds the views (which is needed to detach all the views) cannot be thread safe without extra locking. 
We could have added main thread invalidation as well but I don't really like this as it requires the client not to forget about calling invalidate.

Overall it is very fair to have an object that does mounting and unmounting affined to the main thread and I'm going to put a PR to actually attach a "lifecyle manager" to the root view that will ensure components gets unmounted when the view goes away. Which will remove the concern of dealing the lifecyle of the views within the attach controller.

ocrickard:
Looks like some tests failed.

benlodotcom:
Rebased on some changes that affected the mount functions, it should fix the tests

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -0,0 +1,146 @@
...
+              userInfo:(NSDictionary *)userInfo
+{
+  [_componentDataSource applyChangeset:changeset
+                                  mode:mode
+                              userInfo:userInfo];
+}
+
+static void applyChangesToCollectionView(CKTransactionalComponentDataSourceAppliedChanges *changes, UICollectionView *collectionView)
+{
+  [collectionView reloadItemsAtIndexPaths:[changes.updatedIndexPaths allObjects]];
```
adamjernst: Would be nice to eventually do this the nicer way, but OK

adamjernst: (Certainly not a regression since the existing open source DS does it this way too)

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
adamjernst: btw explicitness is good, _like_

adamjernst:
Let's land this thing.

PR #195, opened 2015-05-19
Work around the static destructor fiasco by using a pointer

adamjernst:
It's annoying that we need to do this, but such is life with C++.


darknoon:
For the record, what's the "static destructor fiasco?"

adamjernst:
@darknoon  destructors are called at `exit` time:
http://en.cppreference.com/w/cpp/utility/program/exit

If other threads are using the objects being destructed, you'll end up with weird crashes.

The ideal fix is to call `quick_exit` instead of `exit`, but we can't enforce that on iOS.

PR #198, opened 2015-05-20
Make CKRelativeDimension ctors inlineable

swolchok:
Helps with code size.


adamjernst:
Build fails — I think you're missing a change.

PR #200, opened 2015-05-21
[CDS] remove the old item enumeration API from CKArrayControllerChangeset

majak:
Another step closer to adding moves.


majak:
I'll convert tests to the new API instead or removing them.

majak:
Kept and converted few tests to the new enumeration API.

Issue #211, opened 2015-05-28
Issues with calling CKMountComponentLayout() without CKComponentLifecycleManager


darknoon:
1. Events do not fall off the components responder chain onto the UIView responder chain at the top because @property `rootComponentMountedView` is not set.

2. Components can never be unmounted because  [component unmount] is private to ComponentKit. What's the best way to manage this?

I'm going to try to tackle these these, but wanted to get them recorded in case I don't get around to it

adamjernst:
Yep, we definitely need to fix both.

adamjernst:
All that's left is to add some sort of `CKUnmountComponentLayout`. @benlodotcom will handle this as part of his transactional data source work I believe.

adamjernst:
Oops.

darknoon:
`CKUnmountComponents()` is available. I think this is mostly addressed.

PR #212, opened 2015-05-28
Fix responder chain for components mounted with CKMountComponentLayout()

darknoon:
Fix for first issue in #211


adamjernst:
@darknoon Merge at will.

PR #216, opened 2015-05-29
[rfc] Check in Cocoapods artifacts

dstnbrkr:
This PR makes it possible to build the ComponentKit library without needing to first install CocoaPods by checking in the Pods folder and generated xcworkspace. This way, CocoaPods will still be our dependency manager but is no longer a build tool.

I've left WildeGuess on Cocoapods for two reasons, see if you agree:
- we need some way to exercise the ComponentKit.podspec
- the WildeGuess Pods folder would contain a full copy of ComponentKit


darknoon:
Hey, I think this might make it harder to manage my fork of ComponentKit, since every time the project file is changed, I'll need to deal with merge conflicts. Also, slight differences between cocoapods versions could cause thrash on the project files as people update them.

Not blocking issues, but worth considering.

dstnbrkr:
Hm - possibly. Might not be that bad since Cocoapods thrashes the xcworkspace instead of the xcodeproj. Pod version changes thrash the Podfile.lock mostly. How come OS X support is a fork?

dstnbrkr:
Convinced that the repo thrash isn't worth it, closing...

Issue #222, opened 2015-06-03
Question: how to access properties of components


stampsrule:
If I have added a UITextField and want to do something with the text the user entered how do access the text?

darknoon:
Ah, yes. It's a bit hard to do editable text at the moment.

As far as I can tell, you need to use a controller on a custom component and then react to mounting by adjusting the first responder, etc. You can make the controller the delegate of the textField etc.

We'd be interested in taking patches here, and I might work on some stuff to make this easier. Do you know what API specifically you'd be interested in?

stampsrule:
Can I email you to discuss this further with you?

stampsrule:
would the controller be a subclass of CKComponentController?
would the custom component be a subclass of CKCompositeComponent?

darknoon:
You could email me (andrewpouliot@fb.com), but I don't check email that much and posting here would probably get more responses from the rest of the team.

Here's a snippet of what I'm doing for the first responder (yeah I have a component controller) https://gist.github.com/darknoon/f7832860f9f9d0f03f11. It's not a composite component, just directly wraps my text field. I have a custom text field class that sends actions to the component in a kind of hacky way. Again, looking to improve this, but maybe that's a start?

stampsrule:
does the file you linked to have a header file?

I am still junior at this but I can see what I can do with what you linked to.

darknoon:
There is more code, but it's wrapped up in the specifics of what I'm doing. You could figure out how to make a general solution.

stampsrule:
When I have finished the textfield component and UISwitch component how do I submit them to component kit?

adamjernst:
@stampsrule We actually try to keep the core really small, so please open source it as a separate cocoapod/repo instead! That way people can pull it in and use them.

stampsrule:
does issue/pull 223 solve this?

stampsrule:
yes it does, please close this

darknoon:
@stampsrule Glad I could help :)

PR #223, opened 2015-06-04
Add a view attribute to send a delegate protocol to components

darknoon:
Purpose:

Avoid always needing to have a controller just because you need events from a view. This goes along with the `CKComponentActionAttribute` where you can get events from `UIControl`s sent up to your component or controller subclass.

I could see this being used to allow editing of UITextFields with some work  #222); the obvious uses are scroll views for gestures, etc. that I take advantage of with the combiners example.

Test plan:

I've tried to write some basic unit tests, but I'm interested in hearing what cases are interesting / things that I've missed.

Limitations:

- only allows one delegate to be set per view (can't think of a situation where this is an issue)
- uses associated objects

### DIFF
--- a/ComponentKit/Utilities/CKComponentDelegateAttribute.mm
+++ b/ComponentKit/Utilities/CKComponentDelegateAttribute.mm
@@ -1,0 +1,130 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKComponentDelegateAttribute.h"
+
+#import <vector>
+#import <objc/runtime.h>
+
+#import "CKAssert.h"
+#import "CKComponentViewInterface.h"
+#import "CKComponentSubclass.h"
+
+std::string identifierFromDefinition(const CKComponentDelegateAttributeDefinition& first)
+{
+  std::string so = "Delegate";
+  for (auto& s : first) {
+    so = so + "-" + sel_getName(s);
+  }
+  return so;
+}
+
+@interface CKComponentDelegateForwarder : NSObject {
+  @package
+  // Weak ref to our view, to grab the component.
+  __weak UIView *_view;
+  CKComponentDelegateAttributeDefinition _defn;
+}
+
+@end
+
+@interface UIView (CKDelegateProxy)
+
+@property (nonatomic, strong, setter=ck_setDelegateProxy:) CKComponentDelegateForwarder *ck_delegateProxy;
+
+@end
+
+CKComponentViewAttributeValue CKComponentDelegateAttribute(SEL selector,
+                                                           CKComponentDelegateAttributeDefinition definition)
+{
+  if (selector == NULL) {
+    return {
+      {
+        std::string("Delegate-noop-") + sel_getName(selector) + "-",
+        ^(UIView *view, id value) {}, ^(UIView *view, id value) {}
+      },
+      @YES  // Bogus value, we don't use it.
+    };
+  }
+
+  return {
+    {
+      std::string(sel_getName(selector)) + identifierFromDefinition(definition),
+      ^(UIView *view, id value){
+
+        // Create a proxy for this set of selectors
+
+        CKCAssertNil(view.ck_delegateProxy,
+                     @"Unsupported: registered two delegate proxies for the same view: %@ %@", view, view.ck_delegateProxy);
+
+        CKComponentDelegateForwarder *proxy = [[CKComponentDelegateForwarder alloc] init];
+        proxy->_view = view;
+        proxy->_defn = definition;
+        view.ck_delegateProxy = proxy;
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-performSelector-leaks"
+        [view performSelector:selector withObject:proxy];
+#pragma clang diagnostic pop
+
+      },
+      ^(UIView *view, id value){
+
+        // When unapplied, remove association with the view
+        CKComponentDelegateForwarder *proxy = view.ck_delegateProxy;
+        proxy->_view = nil;
+        view.ck_delegateProxy = nil;
+
+#pragma clang diagnostic push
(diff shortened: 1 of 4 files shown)


darknoon:
To be clear, I mean having a controller on the direct component that has the scrollview (maybe in a stack layout or background component or whatever). Having a controller when you're handling complex events in general is good!

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -0,0 +1,143 @@
...
+#import <streambuf>
+#import <objc/runtime.h>
+
+#import "CKAssert.h"
+#import "CKInternalHelpers.h"
+#import "CKMutex.h"
+#import "CKComponentViewInterface.h"
+#import "CKComponentSubclass.h"
+
+std::string identifierFromDefinition(const CKComponentDelegateAttributeDefinition& first)
```
adamjernst: Silly nit: we use weird spacing, and might as well be consistent. `const CKFoo &foo`.

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -0,0 +1,143 @@
...
+{
+  std::string so = "Delegate";
+  for (auto& s : first) {
+    so = so + "-" + sel_getName(s);
+  }
+  return so;
```
adamjernst: Great.

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -0,0 +1,143 @@
...
+
+  return {
+    {
+      std::string(sel_getName(selector)) + identifierFromDefinition(definition),
+      ^(UIView *view, id value){
+
+        // Create a proxy for this set of selectors
+
+        CKCAssertNil(view.ck_delegateProxy,
+                     @"Unsupported: registered two delegate proxies for the same view: %@ %@", view, view.ck_delegateProxy);
```
adamjernst: OK, so you can only have one. Fair enough.

ComponentKitTests/CKComponentDelegateAttributeTests.mm
```diff
@@ -0,0 +1,123 @@
...
+  UIScrollView *view = [[UIScrollView alloc] init];
+
+  attr.first.applicator(view, attr.second);
+  XCTAssertNotNil(view.delegate, @"Expected delegate to be set");
+
+  attr.first.unapplicator(view, attr.second);
+  XCTAssertNil(view.delegate, @"Expected delegate to be unset");
+}
+
+static UIScrollView *findScrollView(UIView *v) {
```
adamjernst: Silly nit: { on new line

adamjernst:
I think this is brilliant. I'll leave it to @benlodotcom to give it a once-over, or otherwise accept tomorrow.

adamjernst:
Actually since @benlodotcom  is traveling I'll accept. Nit fixes would be appreciated, but I think this is good to go.

ComponentKitTests/CKComponentDelegateAttributeTests.mm
```diff
@@ -0,0 +1,123 @@
...
+  CKMountComponentLayout(layout, container);
+
+  XCTAssertFalse(noScrollHierarchy.receivedScroll, @"Should not have triggered yet");
+  hierarchy.receivedScroll = NO;
+
+  scroll.contentOffset = CGPointMake(0, 100);
+
+  XCTAssertFalse(noScrollHierarchy.receivedScroll, @"Should not have triggered because we don't want scroll events.");
+  XCTAssertFalse(hierarchy.receivedScroll, @"Should not have triggered on old hierarchy either.");
+
```
adamjernst: This test is great albeit just on the edge of being a bit monolithic. I prefer to break up into smaller tests even if that means a bit of repetition (someone beat into me that each test should have one and only one assertion, which is a nice rule to aspire at least).

However as the test is already written I think it's fine.

@modocache might have thoughts on more tests to write.

modocache: Totally. I think this test could be split up into several smaller ones:

```objc
- (void)testMountingComponentWithViewHierarchyDoesNotTriggerScrollEvent { /* ... */ }
- (void)testScrollingComponentWithViewHierarchyTriggersScrollEvent { /* ... */ }
- (void)testMountingComponentWithoutViewHierarchyDoesNotTriggerScrollEvent { /* ... */ }
- (void)testScrollingComponentWithoutViewHierarchyTriggersScrollEvent { /* ... */ }
```

From what I understand of the intent of the tests, splitting them up should make it possible to get rid of the `-makeObjectsPerformSelector:` line.

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -0,0 +1,143 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-performSelector-leaks"
+        [view performSelector:selector withObject:proxy];
+#pragma clang diagnostic pop
+
+      },
+      ^(UIView *view, id value){
+
+        // When unapplied, remove association with the view
+        CKComponentDelegateForwarder *proxy = view.ck_delegateProxy;
```
darknoon: @adamjernst This line assumes that the unapplicator can only be called once per application of the applicator (since it would crash if proxy is nil. I might put in additional safety here just to be sure.

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -0,0 +1,143 @@
...
+        view.ck_delegateProxy = nil;
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-performSelector-leaks"
+        [view performSelector:selector withObject:nil];
+#pragma clang diagnostic pop
+
+      }
+    },
+    @YES // Bogus value, we don't use it.
```
darknoon: Is there something better we can do here? Maybe standardize on one value that means to still run the applicator, but isn't anything in particular? Perhaps allow constructing without the last parameter, which would make a pair with that value?

ComponentKitTests/CKComponentDelegateAttributeTests.mm
darknoon: I'm trying to test the sequence "mount, unmount, mount different property," since that could be reasonably be broken by something changing.

PR #232, opened 2015-06-05
Add CKComponentFadeTransition

MarcoSero:
This can be used by CKNetworkImageComponent to specify a fade transition when the image is fetched from the network.


adamjernst:
This is really cool! We have something similar inside Facebook.

I want to keep the core infrastructure of ComponentKit relentlessly simple. Can you open source this as a fork of CKNetworkImageComponent in a separate repo/cocoapod so that people could use it if they wanted?

MarcoSero:
Yep I imagined that.
I created my own extension at https://github.com/MarcoSero/CKComponentFadeTransition

Issue #236, opened 2015-06-05
CKComponentGestureAttribute makes it hard to tell which gesture recognizer you're dealing with


darknoon:
When you're answering the "should begin" question about a gesture recognizer, it's hard to see what's what because all of the component-attribute gesture recognizers have the same -handleGesture: selector set and you didn't create them so it's hard to refer to them by pointer.

I'm going to put up a pull request to allow you to look up the selector for a gesture, but maybe we could preserve the original selector by making each action forwarder distinct to a view, akin to or using the delegate forwarder. Then it can answer the `-forwardingTargetForSelector:` question by looking at its view.

Alternately, we could go through the slower invocation-based api to look at the gesture parameter to grab its view.

PR #237, opened 2015-06-05
Provide a way to map back to the actual action selector on a gesture.

darknoon:
For #236

Example usage:

````objc++
    - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer 
    shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer
    {
      std::set<std::pair<SEL, SEL>> allowed{
        {@selector(mouseDown:gesture:), @selector(mouseDragged:gesture:)},
        {@selector(mouseDown:gesture:), @selector(doubleClick:gesture:)},
        {@selector(mouseDown:gesture:), @selector(doubleClickedRedacted:gesture:)},
      };

      return allowed.find({CKComponentGestureGetAction(gestureRecognizer), CKComponentGestureGetAction(otherGestureRecognizer)}) != allowed.end();
    }
````

### DIFF
--- a/ComponentKit/Utilities/CKComponentGestureActions.h
+++ b/ComponentKit/Utilities/CKComponentGestureActions.h
@@ -52,3 +52,9 @@ CKComponentViewAttributeValue CKComponentGestureAttribute(Class gestureRecognize
                                                           CKComponentGestureRecognizerSetupFunction setupFunction,
                                                           CKComponentAction action,
                                                           CKComponentForwardedSelectors delegateSelectors = {});
+
+/**
+ Allows mapping a UIGestureRecognizer back to the original CKComponentAction selector,
+ since ComponentKit internally changes the selector to be able send to the component responder chain.
+ */
+CKComponentAction CKComponentGestureGetAction(UIGestureRecognizer *gesture);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ComponentKit/Utilities/CKComponentGestureActions.h [hunk 1]
```diff
@@ -52,3 +52,9 @@ CKComponentViewAttributeValue CKComponentGestureAttribute(Class gestureRecognize
                                                           CKComponentGestureRecognizerSetupFunction setupFunction,
                                                           CKComponentAction action,
                                                           CKComponentForwardedSelectors delegateSelectors = {});
+
+/**
+ Allows mapping a UIGestureRecognizer back to the original CKComponentAction selector,
+ since ComponentKit internally changes the selector to be able send to the component responder chain.
+ */
+CKComponentAction CKComponentGestureGetAction(UIGestureRecognizer *gesture);
```
adamjernst: nit: file should end with newline

adamjernst:
Why does your example use a `std::pair` instead of just a `set<SEL>`? Just curious.

darknoon:
@adamjernst: true, in this example all of the first selectors are the same, but in other places they can vary, thus checking whether the pair is present is necessary.

PR #241, opened 2015-06-08
Split CKComponentSnapshotTestCase into a separate pod

adamjernst:
This breaks ComponentKitTestLib's dependency on XCTest.


modocache:
Nice!

Issue #238, opened 2015-06-08
Truncation attributed string cause crash.


tranngoclinh88:
I tried to set truncation attributed string for text in it causes crash.
Here is the detail:

![screen shot 2015-06-05 at 11 25 35 am](https://cloud.githubusercontent.com/assets/1511264/8027406/d910c45a-0dc1-11e5-8167-c032b3a16b11.png)

![screen shot 2015-06-08 at 9 34 15 am](https://cloud.githubusercontent.com/assets/1511264/8027407/db1d8ada-0dc1-11e5-8922-02b98f14b978.png)

ocrickard:
Can you please provide the code where you're configuring the truncation attributed string?

tranngoclinh88:
Here is the code that build truncation attributed string.
![screen shot 2015-06-09 at 8 53 33 am](https://cloud.githubusercontent.com/assets/1511264/8049140/2a3d1298-0e85-11e5-9501-d3a9eef054e6.png)

Here is the code that set the truncation attributed string for CKTextComponent.
![screen shot 2015-06-09 at 8 54 05 am](https://cloud.githubusercontent.com/assets/1511264/8049141/2bf81268-0e85-11e5-9fe4-362c9c7698c5.png)

Sometimes it does not crash right after showing the collection view, it crashes while scroll up and down.

ocrickard:
Can you please provide code for [IAWStringBuilder readMoreAttributedString]?

ocrickard:
Can you please provide the stack trace at the crash site?  I'd like to understand where this call is coming from.

ocrickard:
I tested this locally and can't repro the issue.  Can you try and get an isolated repro in a unit test?

tranngoclinh88:
@ocrickard OK. I'll try.

ocrickard:
Also, this may be an edge case with the actual string you're using here.  Can you take a screenshot of what the unit is supposed to look like when it doesn't crash?

tranngoclinh88:
Is that right?
![screen shot 2015-06-09 at 10 08 03 am](https://cloud.githubusercontent.com/assets/1511264/8049890/7b32e6e6-0e8f-11e5-9e1e-65fb5869e47c.png)

tranngoclinh88:
![screen shot 2015-06-09 at 10 10 15 am](https://cloud.githubusercontent.com/assets/1511264/8049922/fa754124-0e8f-11e5-9848-7bbe98e8bb5e.png)

[...]

PR #244, opened 2015-06-09
Fix tail truncation behavior when truncation rect is longer than last line rect

ocrickard:
This function assumed that the truncation token would be smaller than the last line.  This corrects the behavior for forced-truncation (maximum number of lines) cases where the truncation token is larger than the last line of text.

Fixes https://github.com/facebook/componentkit/issues/238


ocrickard:
Somehow it's not finding my snapshot.  Will update.

ocrickard:
Uggh, I had to download the iOS 8.1 simulator and re-record the snapshots for both iOS 8.1 and 8.3 in 64-bit and non-64-bit.

ComponentTextKit/TextKit/CKTextKitTailTruncater.mm
```diff
@@ -95,10 +95,12 @@ - (NSUInteger)_calculateCharacterIndexBeforeTruncationMessage:(NSLayoutManager *
   NSUInteger firstClippedGlyphIndex = [layoutManager glyphIndexForPoint:beginningOfTruncationMessage
                                                         inTextContainer:textContainer
                                          fractionOfDistanceThroughGlyph:NULL];
+  // If it didn't intersect with any text then it should just return the last visible character index, since the
+  // truncation rect can fully fit on the line without clipping any other text.
+  if (firstClippedGlyphIndex == NSNotFound) {
+    return [layoutManager characterIndexForGlyphAtIndex:lastVisibleGlyphIndex];
+  }
   NSUInteger firstCharacterIndexToReplace = [layoutManager characterIndexForGlyphAtIndex:firstClippedGlyphIndex];
-  CKAssert(firstCharacterIndexToReplace != NSNotFound,
```
ocrickard: It's actually fine for this to be NSNotFound, it just means we don't intersect with the last line.

ComponentTextKitApplicationTests/CKTextComponentTests.mm
```diff
@@ -384,4 +384,27 @@ - (void)testComplexLigatures
   CKSnapshotVerifyComponent(c, kUnrestrictedSize, @"");
 }
 
+- (void)testTruncationStringLargerThanLastLine
```
modocache: Can you rename this test to indicate what *should* happen when the truncation string is larger than the last line?

modocache:
Accepted; just the one nit for the test method name.

PR #245, opened 2015-06-10
Make announcers Obj-C friendly.

majak:
I have splitted `CKComponentAnnouncerBase` into several files in order to make possible to use announcers in pure Obj-C (as opposed to Obj-C++).
Previouly `CKComponentAnnouncerBase.h` contained C++ code, which is no longer true.

@benlodotcom


ComponentKit/DataSources/Common/CKComponentAnnouncerBaseInternal.h
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <vector>
+
+@interface CKComponentAnnouncerBase()
```
ocrickard: nit: space between class name and class extension parens

ComponentKit/DataSources/Common/CKComponentAnnouncerBaseInternal.h
```diff
@@ -0,0 +1,29 @@
...
+{
+@public
+  // we make this public, so it can be accessed from CK::AnnouncerHelper
```
ocrickard: nit: I realize this predates your changes here, but it'd be nice to make this comment match our style for multiline...

ComponentKit/DataSources/Common/CKComponentAnnouncerBase.mm
```diff
@@ -9,86 +9,7 @@
...
-    //  1) dealing with __weak id as a key in a hash is more complicated
-    //  2) we assume that the number of listeners is relatively small, and
-    //     add/remove is not a frequent event.
-    // n.b. using boost::shared_ptr might lead to faster code, since it's lockless
-    std::shared_ptr<const std::vector<__weak id>> _listenerVector;
-  }
-@end
+#import "CKComponentAnnouncerBaseInternal.h"
 
 @implementation CKComponentAnnouncerBase
```
ocrickard: Without the C++ this header is left with no docs at all.  At least a small docblock to explain what this is might be helpful?

ComponentKit/DataSources/Common/CKComponentAnnouncerHelper.h
```diff
@@ -0,0 +1,59 @@
...
+        auto frozenListeners = loadListeners(self);
+        if (frozenListeners) {
+          for (id listener : *frozenListeners) {
+            objc_msgSendTyped(listener, s, args...);
+          }
+        }
+      }
+      
+      template<typename... ARGS>
+      static void callOptional(CKComponentAnnouncerBase *self, SEL s, ARGS... args) {
```
ocrickard: Forgive my lack of knowledge on C++, but:

1. why is this static (I bet there's a good reason, just no idea).
2. Can we really not put the implementation in the .mm file?

ocrickard:
Oh boy am I excited to see less C++ in public headers!

ocrickard:
Seems fine to me, added nits and questions around the C++, but this doesn't actually change or add anything, so I can't hold you up on improving the existing code ;)

ComponentKit/DataSources/Common/CKComponentAnnouncerHelper.h
majak: 1. Not sure, digging into history didn't reveal any discussion around this. I assume it was done in order to make announcing as efficient as possible (cpu and memory wise).
2. Unfortunately not, because templates. I'm not an expert in C++ either, but I found this explanation: http://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file

PR #254, opened 2015-06-16
Don't animate the creation of new views

### LINKED ISSUE #202: Wrong bounds animation in 0.12
Did anything change in version `0.12` to animate bounds change between states?

I'm overriding `boundsAnimationFromPreviousComponent:` but the frame from which the animation starts in `0.12` is wrong.

From the diff I can see that a new function `CKComponentBoundsAnimationApply` has been added. It makes me wonder whether there was a regression perhaps?

adamjernst:
Also disables animation when recycling a view to show a logically distinct component (as determined by `scopeFrameToken`).

Fixes #202


ComponentKit/Core/CKComponent.mm
```diff
@@ -125,6 +126,7 @@ - (CKComponentViewContext)viewContext
 
   UIView *v = effectiveContext.viewManager->viewForConfiguration([self class], viewConfiguration);
   if (v) {
+    CKMountAnimationGuard g(v.ck_component, self);
```
modocache: Wow, this is super cool. C++ is neat.

ComponentKitApplicationTests/CKComponentBoundsAnimationTests.mm
```diff
@@ -0,0 +1,145 @@
...
+  NSSet *secondMountedComponents = CKMountComponentLayout(secondLayout, container);
+
+  CKBoundsAnimationRecordingView *v = (CKBoundsAnimationRecordingView *)secondResult.component.viewContext.view;
+  XCTAssertFalse(v.animatedLastBoundsChange);
+
+  [firstMountedComponents makeObjectsPerformSelector:@selector(unmount)];
+  [secondMountedComponents makeObjectsPerformSelector:@selector(unmount)];
+}
+
+@end
```
modocache: Are there no tests that actions are re-enabled once the mount animation guard goes out of scope? Should there be? Disabling the actions affects the global state, right?

adamjernst: True. Interesting! Let me consider.

modocache:
A few questions but LGTM! :+1:

adamjernst:
Follow-up test coverage for that sounds like a great idea.

PR #257, opened 2015-06-17
Simplify ComponentMountContext

adamjernst:
Since bleed-out has been simplified, this can be simplified too.


modocache:
Ah, OK. So since we no longer jump back to the root view manager, we can eliminate the code that would passed references to that root view manager, right? Good stuff.

PR #258, opened 2015-06-17
Blocking animations in the parent should block animations in the child

adamjernst:
Otherwise, we fix the issue of unwanted animations in a parent component but the children still animate willy-nilly.

### DIFF
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -126,7 +126,7 @@ struct CKComponentMountInfo {
 
   UIView *v = effectiveContext.viewManager->viewForConfiguration([self class], viewConfiguration);
   if (v) {
-    CKMountAnimationGuard g(v.ck_component, self);
+    CKMountAnimationGuard g(v.ck_component, self, context);
     if (_mountInfo->view != v) {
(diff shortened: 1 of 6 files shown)


ComponentKit/Core/CKComponent.mm
```diff
@@ -143,7 +143,7 @@ - (CKComponentViewContext)viewContext
     [v setBounds:{v.bounds.origin, size}];
 
     _mountInfo->viewContext = {v, {{0,0}, v.bounds.size}};
-    return {.mountChildren = YES, .contextForChildren = effectiveContext.childContextForSubview(v)};
+    return {.mountChildren = YES, .contextForChildren = effectiveContext.childContextForSubview(v, g.blockedAnimations)};
```
modocache: Über-nit, but when I read this line I thought for a second that `blockedAnimations` was a list of all the animations that were blocked--which I thought would be super cool. I had to look at the header file to remember that it was a boolean indicating whether animations were blocked. Maybe this could use a better name?

(I realize that I reviewed #254, which added it in the first place. :stuck_out_tongue:)

adamjernst:
Fix @modocache 's comment.

benlodotcom:
:+1:

PR #259, opened 2015-06-19
Adding CKComponentDelegateAttribute to umbrella header

nickfrey:


ocrickard:
Seems like a reasonable change to me.

PR #266, opened 2015-06-24
Allow specifying a reuse name

darknoon:
The idea is to provide a way such that when you update to a new component you can make sure views you are intending to animate don't get reused in an unfortunate way.

I'm really struggling at how to test this in a sane way. @majak @adamjernst ?

### DIFF
--- a/ComponentKit/Core/CKComponentViewConfiguration.h
+++ b/ComponentKit/Core/CKComponentViewConfiguration.h
@@ -37,6 +37,15 @@ struct CKComponentViewClass {
   CKComponentViewClass(Class viewClass);
 
   /**
+   Specifies that the component should have a view of the given class. The class will be instantiated with UIView's
+   designated initializer -initWithFrame:.
+   
+   This variant introduces a reuseName that additionally constrains reuse. So, if a component is being animated to another
+   component, you can be sure that the corresponding pieces animate to eachother instead of to a random view.
+   */
+  CKComponentViewClass(Class viewClass, std::string reuseName);
+
+  /**
    A variant that allows you to specify two selectors that are sent as a view is hidden/unhidden for future reuse.
    Note that a view can be reused but not hidden so never enters the pool (in which case these selectors won't be sent).
    @param didEnterReusePoolMessage Sent to the view just after it has been hidden for future reuse.
--- a/ComponentKitTests/CKComponentViewReuseTests.mm
+++ b/ComponentKitTests/CKComponentViewReuseTests.mm
@@ -14,9 +14,11 @@
 #import "ComponentViewReuseUtilities.h"
 #import "CKComponent.h"
 #import "CKComponentInternal.h"
+#import "CKStackLayoutComponent.h"
 #import "CKComponentLifecycleManager.h"
 #import "CKComponentProvider.h"
 #import "CKCompositeComponent.h"
+#import "CKComponentViewInterface.h"
 
 @interface CKComponentViewReuseTests : XCTestCase <CKComponentProvider>
 @end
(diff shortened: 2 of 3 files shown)


ComponentKit/Core/CKComponentViewConfiguration.h
```diff
@@ -37,6 +37,15 @@ struct CKComponentViewClass {
   CKComponentViewClass(Class viewClass);
 
   /**
+   Specifies that the component should have a view of the given class. The class will be instantiated with UIView's
+   designated initializer -initWithFrame:.
+   
+   This variant introduces a reuseName that additionally constrains reuse. So, if a component is being animated to another
+   component, you can be sure that the corresponding pieces animate to eachother instead of to a random view.
+   */
+  CKComponentViewClass(Class viewClass, std::string reuseName);
```
adamjernst: `reuseKey`?

ComponentKit/Core/CKComponentViewConfiguration.h
```diff
@@ -37,6 +37,15 @@ struct CKComponentViewClass {
...
+
+  /**
    A variant that allows you to specify two selectors that are sent as a view is hidden/unhidden for future reuse.
```
adamjernst: It's unfortunate that now you must choose between `didEnterReusePoolMessage`, function pointers, and reuse keys.

Could some use of default parameters get around this for us?

ComponentKitTests/CKComponentViewReuseTests.mm
```diff
@@ -330,5 +370,34 @@ + (instancetype)newWithComponent:(CKComponent *)component
...
+               newWithView:{
+                 {[UIView class]},
+               }
+               size:{}]};
+    }
+  })];
+}
+
+
+@end
```
adamjernst: Nice tests.

ComponentKitTests/CKComponentViewReuseTests.mm
```diff
@@ -261,6 +268,39 @@ - (void)testThatComponentThatInjectsAnIntermediateViewNotControlledByComponentsD
...
+  // Make some otherwise-indistinguishable views with one that is special
+
+  CKComponentLifecycleManager *clm = [[CKComponentLifecycleManager alloc] initWithComponentProvider:[CKTestIdentifierComponentProvider class]];
+  [clm updateWithState:[clm prepareForUpdateWithModel:@[@"", @"", @"", @"", @"A", @""] constrainedSize:{{0,0}, {100, 100}} context:nil]];
+  [clm attachToView:rootView];
+
+  auto findA = [](UIView *where) {
+    UIView *aView;
+    for (UIView *v in where.subviews) {
+      if (v.ck_component && v.ck_component.viewConfiguration.viewClass().getIdentifier() == "UIView/id/A") {
```
darknoon: Well, this part isn't so nice.

majak:
I don't have a full context on what is the unintended behavior you are trying to avoid, so my comment might not be really valid.
But I'm not terribly excited about adding a reuse name like this to "public" API if this is the only use case. Would it be possible to do/prevent reuse automatically when an animation is specified instead? My other concern is this is a pretty hidden feature, and even if someone would have an issue with animation I'm not sure how easy it would be to figure this out.

adamjernst:
For the reasons @majak gives, I'm going to pass on this right now. The need for this hasn't yet cropped up multiple times and I want to keep it out of the infra until that time. When we do need it, I'd prefer to rethink how we specify view classes to avoid the explosion of constructor overrides.

So — let's put it on ice and resurrect it if we need it later.

PR #267, opened 2015-06-24
Adding animationsOnInitialMount

abhinav24:
Once this is pulled, I'll update the relevant docs


adamjernst:
Great, but looks like we need a rebase?

PR #268, opened 2015-06-24
Adding canPerformAction to the component responder chain

deanmarsinelli:
I added the canPerformAction method to the component responder chain (in CKComponent and CKComponentController) so that subclasses of components can choose to implement this method or not. If they do not, the behavior is the same as it was before the methods were added. If subclasses do implement this they can return YES to handle the the event or NO to be skipped in the responder chain and the event will continue to traverse it.

I also added a test to make sure that the correct object is being returned when a component and its controller (the next responder) both implement the method with different conditions.

### DIFF
--- a/ComponentKit/Core/CKComponent.h
+++ b/ComponentKit/Core/CKComponent.h
@@ -54,4 +54,12 @@ struct CKComponentViewContext {
  */
 - (id)nextResponder;
 
+/**
+ When an event occurs, if a component implements the canPerformAction method it will be called in the normal responder
+ chain. If the method returns YES, the traversal will end and the action will be called on this component. If the
+ method returns NO, the component will be skipped and the event will continue to traverse the responder chain to the
+ component's next responder.
+ */
+- (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
+
 @end
--- a/ComponentKit/Core/CKComponent.h
+++ b/ComponentKit/Core/CKComponent.h
@@ -54,12 +54,4 @@ struct CKComponentViewContext {
  */
 - (id)nextResponder;
 
-/**
- When an event occurs, if a component implements the canPerformAction method it will be called in the normal responder
- chain. If the method returns YES, the traversal will end and the action will be called on this component. If the
- method returns NO, the component will be skipped and the event will continue to traverse the responder chain to the
- component's next responder.
- */
-- (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
-
 @end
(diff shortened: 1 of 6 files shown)


ComponentKit/Core/CKComponent.h
```diff
@@ -54,4 +54,12 @@ struct CKComponentViewContext {
  */
 - (id)nextResponder;
 
+/**
+ When an event occurs, if a component implements the canPerformAction method it will be called in the normal responder
+ chain. If the method returns YES, the traversal will end and the action will be called on this component. If the
+ method returns NO, the component will be skipped and the event will continue to traverse the responder chain to the
+ component's next responder.
+ */
+- (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
```
ocrickard: Oops, this probably belongs in CKComponentSubclass.h

ComponentKit/Core/CKComponentController.h
```diff
@@ -70,4 +70,12 @@
  */
 - (id)nextResponder;
 
+/**
+ When an event occurs, if a controller implements the canPerformAction method it will be called in the normal responder
```
ocrickard: "When an action is triggered"?

ComponentKit/Core/CKComponent.h
```diff
@@ -54,4 +54,12 @@ struct CKComponentViewContext {
  */
 - (id)nextResponder;
 
+/**
+ When an event occurs, if a component implements the canPerformAction method it will be called in the normal responder
```
ocrickard: Consider:

"When an action is triggered, a component may use this method to either capture, or ignore a given action.  The default implementation simply uses respondsToSelector: to determine if the component can perform the given action.

In practice this is useful only for integrations with UIMenuController, whose API walks the UIResponder chain to determine which menu items to display.  You should not override this for standard component actions."

ComponentKitTests/CKComponentControllerTests.mm
```diff
@@ -194,6 +208,14 @@ - (void)updateStateToIncludeNewAttribute
     return @YES;
   }];
 }
+
+- (BOOL)canPerformAction:(SEL)action withSender:(id)sender
+{
+  if (sender == self)
```
adamjernst: nit: always use braces

adamjernst:
Merge at will!

ComponentKitTests/CKComponentControllerTests.mm
deanmarsinelli: can't believe I forgot those!

PR #272, opened 2015-06-25
Add parameter for layout manager subclass to CKTextKitContext

itsthejb:
Hi all,

Me again :wink: Proposed minor feature addition here to support the following use case:

We have a "tokens" view component which displays tappable tokens with a rounded corner background like this:

screen shot 2015-06-25 at 15 30 00

`NSAttributedString` of course has `NSBackgroundColorAttributeName`, but this only allows colouring with sharp rectangles. In order to do this you need an `NSLayoutManager` subclass and override one of its drawing methods.

Since `CKTextComponent` doesn't provide enough customisation to its underlying renderer (which is perfectly fine), the approach I've taken is to create a `CKTextKitContext` within my component's custom view in order to create the `NSTextContainer` which is eventually used to create a `UITextView` right at the end of the layout stage. This is very important in order to run into thread safety problems with the rest of CK.

Soooo... In order to facilitate all of this I've added an optional (default to `NSLayoutManager`) `layoutManagerSubclass` parameter to `CKTextKitContext`.

Most likely adding *another* parameter isn't exactly desirably, but I couldn't see any other way to cleanly add this functionality. Since this is an internal class which is rarely used directly, this seems like NBD, but of course I welcome feedback.

Couple of simple test cases included, and others updated of course.

Thanks!


ocrickard:
Before I review, a quick comment:

In your description here you say that you're creating a `UITextView` at the end of the layout stage.  Do you mean that within the scope of `-computeLayoutThatFits:`?  When used with a CKComponentDataSource that method will be invoked on a bg thread, and so creation of a UITextView (while it may work 99% of the time) will probably crash.

Also, the reason we don't expose these classes is that there are tons of threading concerns when interacting with TextKit on a bg thread due to some unprotected global caches that they use under the hood.  We've tried to lock the usage where we know we will hit contention points, but it's just too easy to shoot yourself in the foot with these on a different thread.

Now I'm off to review!

itsthejb:
Hey there,

No; I defer the creation of the `UITextView` until the last moment (actually, `-layoutSubviews`, which is a bit ugly, but works). Up until this point I'm using `CKTextKitContext` to do all the sizing, which is of course of background threads - hence the need to use its shared locking facilities.

You can actually do this *without* overriding the type in `CKTextKitContext` - remove the layout manager created in the context and add your own. However this is quite untidy since you then need to retain the custom layout manager somewhere else. Nice to have everything encapsulated into a TextKit "stack" in `CKTextKitContext`.

So this seems like a good enough way to provide more flexibility to the framework without actually doing something like adding layout manager configuration to `CKTextComponent`. I assume you think that would be getting out of scope?

Please let me know what you think after a review!

ComponentTextKit/TextKit/CKTextKitContext.h
```diff
@@ -26,7 +26,8 @@
 - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
                            lineBreakMode:(NSLineBreakMode)lineBreakMode
                     maximumNumberOfLines:(NSUInteger)maximumNumberOfLines
-                         constrainedSize:(CGSize)constrainedSize;
+                         constrainedSize:(CGSize)constrainedSize
+                   layoutManagerSubclass:(Class)layoutManagerSubclass;
```
ocrickard: We have to add documentation for what this is and why you might use it.  There must also be a warning in this docblock indicating that whatever layout manager class you provide here (while only accessed from one thread at a time) multiple instances will be in use at the same time on different threads.  Any global caches must be strongly protected.

ocrickard: (I only say this since most large text systems heavily use global caches for performance and stable memory consumption)

ComponentTextKit/TextKit/CKTextKitRenderer.mm
```diff
@@ -54,7 +54,8 @@ - (instancetype)initWithTextKitAttributes:(const CKTextKitAttributes &)attribute
     _context = [[CKTextKitContext alloc] initWithAttributedString:attributes.attributedString
                                                     lineBreakMode:attributes.lineBreakMode
                                              maximumNumberOfLines:attributes.maximumNumberOfLines
-                                                  constrainedSize:shadowConstrainedSize];
+                                                  constrainedSize:shadowConstrainedSize
+                                            layoutManagerSubclass:nil];
```
ocrickard: If we're always passing nil here then how is an external consumer supposed to use this?  Using the CKTextKitContext directly is not advisable as you circumvent much of our caching infrastructure that allows this system to be even remotely performant for fast-scrolling views.

ocrickard:
I think that this approach is a necessary step considering TextKit's customization procedures (subclass everything, lol).  I would suggest that we add this as parameter on the CKTextKitAttributes struct so that users of the API can use the renderer caches, etc.

However, I'm not a huge fan of providing Classes for initialization in ComponentKit.  I would prefer a factory function that is defined in the API as a C function pointer.  A C function pointer is preferable to a block for this type of factory since these values must remain threadsafe and cache-able on arbitrary threads, and so capture of local state isn't a great pattern.

itsthejb:
Ok - surprised to hear that you'd prefer it in `CKTextKitAttributes`. Thought this was a little too niche for inclusion there. Seems like you'd prefer providing a pointer to a factory C function, as used in `CKComponentViewClass`? That shouldn't be too much work to change to when I have a few moments.

ocrickard:
lol, yeah, generally we're against adding niche features to core ComponentKit, but in this case, the answer of how you customize something in TextKit is: "Subclass NSLayoutManager", so for advanced usages, I think it's reasonable to include in CKTextKit itself.

itsthejb:
Indeed. I'll try to get this amended asap. Little snowed under here, though, so might be not until next week.

ocrickard:
Looking forward to it!  Thanks!

itsthejb:
Ok - I had to escalate this due to issues with our semi-native implementation. Definitely still requires review on a few points, which I'll line comment.

I created a snapshot test for the basic functionality. I've noticed the use of OS version-specific reference images. I'm not sure whether you're creating these in order to test for differences between the OS versions, or to allow for running of the test suite on local environments. I'm assuming it's (mostly) the latter, in which case it would seem adequate to just copy the reference image to the relevant folders. I don't think I can easily download the 8.0 simulator these days, in any case... Please let me know about this.

ComponentTextKit/CKTextComponent.mm
```diff
@@ -65,6 +65,7 @@ + (instancetype)newWithTextAttributes:(const CKTextKitAttributes &)attributes
                        viewAttributes:(const CKViewComponentAttributeValueMap &)viewAttributes
                  accessibilityContext:(const CKTextComponentAccessibilityContext &)accessibilityContext
 {
+  CKAssertTrue(!attributes.customLayoutManagerClass || [attributes.customLayoutManagerClass isSubclassOfClass:[NSLayoutManager class]]);
```
itsthejb: *Seems* worth having a sanity check assertion for this somewhere, and this seems like as good a place as any. Please let me know about this one.

ComponentTextKit/TextKit/CKTextKitAttributes.h
```diff
@@ -77,6 +77,10 @@ struct CKTextKitAttributes {
    The radius that should be applied to the shadow blur.  Larger values mean a larger, more blurred shadow.
    */
   CGFloat shadowRadius;
+  /**
+   A pointer to a function that that returns a custom layout manager subclass. If nil, defaults to NSLayoutManager.
+   */
+  NSLayoutManager *(*layoutManagerFactory)(void);
```
itsthejb: Switched to the function pointer approach as you suggested. This is nice since we get some type safety from the compiler and no need for internal assertions.

ComponentTextKit/TextKit/CKTextKitAttributes.h
```diff
@@ -104,6 +109,7 @@ struct CKTextKitAttributes {
     && maximumNumberOfLines == other.maximumNumberOfLines
     && shadowOpacity == other.shadowOpacity
     && shadowRadius == other.shadowRadius
+    && layoutManagerFactory == other.layoutManagerFactory
```
itsthejb: As noted, I moved this earlier in the order of evaluation since it's just a pointer comparison.

ComponentTextKitApplicationTests/CKTextComponentTests.mm
```diff
@@ -407,4 +426,31 @@ - (void)testTruncationStringLargerThanLastLine
...
+- (void)testShouldUseCustomLayoutManagerClass
+{
+	NSString *contentString = @"This is a string with a border behind every other word";
+  NSMutableAttributedString *attributed = [[NSMutableAttributedString alloc] initWithString:contentString attributes:@{NSFontAttributeName : [UIFont fontWithName:@"HelveticaNeue" size:12.5]}];
+  __block NSUInteger index = 0;
+  [contentString enumerateSubstringsInRange:[contentString rangeOfString:contentString]
+                                    options:NSStringEnumerationByWords
+                                 usingBlock:^(NSString *substring, NSRange substringRange, NSRange enclosingRange, BOOL *stop) {
+                                   if (index++ % 2) {
+                                     [attributed addAttribute:NSBackgroundColorAttributeName value:[UIColor greenColor] range:substringRange];
```
itsthejb: The verification here is that I specify `+[UIColor redColor]` as the background color attribute, which I then override with green in the layout manager subclass. Ergo, the subclass is used.

ComponentTextKit/TextKit/CKTextKitAttributes.h
ocrickard: Love it.

ComponentTextKit/TextKit/CKTextKitAttributes.h
ocrickard: Yep, seems fine.

ocrickard: Seems like we should probably include the layout manager factory in the hashing function? https://github.com/facebook/componentkit/blob/master/ComponentTextKit/TextKit/CKTextKitAttributes.mm#L20

ComponentTextKit/TextKit/CKTextKitContext.mm
```diff
@@ -33,7 +34,7 @@ - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
     std::lock_guard<std::mutex> l(__static_mutex);
     // Create the TextKit component stack with our default configuration.
     _textStorage = (attributedString ? [[NSTextStorage alloc] initWithAttributedString:attributedString] : [[NSTextStorage alloc] init]);
-    _layoutManager = [[NSLayoutManager alloc] init];
+    _layoutManager = layoutManagerFactory ? layoutManagerFactory() : [[NSLayoutManager alloc] init];
```
ocrickard: :P I like this version much better.

ComponentTextKitApplicationTests/CKTextComponentTests.mm
```diff
@@ -29,6 +29,25 @@
...
+
+@end
+
+@implementation CKTextComponentTestLayoutManager
+
+- (void)fillBackgroundRectArray:(const CGRect *)rectArray count:(NSUInteger)rectCount forCharacterRange:(NSRange)charRange color:(UIColor *)color
+{
+  CGContextRef ctx = UIGraphicsGetCurrentContext();
+  CGContextSetFillColorWithColor(ctx, [UIColor redColor].CGColor);
+  CGContextFillRects(ctx, rectArray, rectCount);
```
ocrickard: trolol, I still find this subclass-to-override pattern appalling... oh well.

ComponentTextKitApplicationTests/CKTextComponentTests.mm
ocrickard: Yep, seems reasonable to me.

ComponentTextKitApplicationTests/CKTextComponentTests.mm
itsthejb: *shrug*... That's Cocoa!

ocrickard:
Looks good, let's add the function to the hashing function and get this merged!

ComponentTextKitApplicationTests/CKTextComponentTests.mm
ocrickard: Yep, sure is.

ComponentTextKit/TextKit/CKTextKitAttributes.h
itsthejb: Ok - can do that in a moment. Would you like me to squash the branch while I'm at it?

ocrickard: Squashing would be great.

itsthejb:
Squashified

ocrickard:
Beautiful.

ocrickard:
I'll merge once the travis build completes.

itsthejb:
Great! Glad to contribute

PR #274, opened 2015-06-25
Fix completely broken hash function

darknoon:
Previously, it would often return 0 since left shifting all previous results by 32-bits deletes a lot of information. Most importantly, the actual text hash!


ocrickard:
Yep, this was a really bad hash function.  Thanks for fixing.

PR #278, opened 2015-06-29
Improve the mount/unmount APIs

adamjernst:
No more set arithmetic required to use component mounting.

### DIFF
--- a/ComponentKit/Core/CKComponentLayout.h
+++ b/ComponentKit/Core/CKComponentLayout.h
@@ -44,5 +44,19 @@ struct CKComponentLayoutChild {
   CKComponentLayout layout;
 };
 
-/** Recursively mounts the layout in the view, returning a set of the mounted components. */
-NSSet *CKMountComponentLayout(const CKComponentLayout &layout, UIView *view, CKComponent *supercomponent = nil);
+/**
+ Recursively mounts the layout in the view, returning a set of the mounted components.
+ @param layout The layout to mount, usually returned from a call to -layoutThatFits:parentSize:
+ @param view The view in which to mount the layout.
+ @param previouslyMountedComponents If a previous layout was mounted, pass the return value of the previous call to
+        CKMountComponentLayout; any components that are not present in the new layout will be unmounted.
+ @param supercomponent Usually pass nil; if you are mounting a subtree of a layout, pass the parent component so the
+        component responder chain can be connected correctly.
+ */
+NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
+                              UIView *view,
+                              NSSet *previouslyMountedComponents,
+                              CKComponent *supercomponent);
+
+/** Unmounts all components returned by a previous call to CKMountComponentLayout. */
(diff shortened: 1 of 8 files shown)


ComponentKit/Core/CKComponentLayout.h
```diff
@@ -44,5 +44,19 @@ struct CKComponentLayoutChild {
...
+ @param supercomponent Usually pass nil; if you are mounting a subtree of a layout, pass the parent component so the
+        component responder chain can be connected correctly.
+ */
+NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
+                              UIView *view,
+                              NSSet *previouslyMountedComponents,
+                              CKComponent *supercomponent);
+
+/** Unmounts all components returned by a previous call to CKMountComponentLayout. */
+void CKUnmountComponents(NSSet *componentsToUnmount);
```
darknoon: Now that you're adding unmounting implicitly to mount, is there any sense in passing CKComponentLayout{} into CKMountComponentLayout to unmount components?

CKMountComponentLayout({}, nil, previouslyMountedComponents, nil);

darknoon:
Looks good, though.

ComponentKit/Core/CKComponentLayout.h
```diff
@@ -44,5 +44,19 @@ struct CKComponentLayoutChild {
...
 
-/** Recursively mounts the layout in the view, returning a set of the mounted components. */
-NSSet *CKMountComponentLayout(const CKComponentLayout &layout, UIView *view, CKComponent *supercomponent = nil);
+/**
+ Recursively mounts the layout in the view, returning a set of the mounted components.
+ @param layout The layout to mount, usually returned from a call to -layoutThatFits:parentSize:
+ @param view The view in which to mount the layout.
+ @param previouslyMountedComponents If a previous layout was mounted, pass the return value of the previous call to
+        CKMountComponentLayout; any components that are not present in the new layout will be unmounted.
+ @param supercomponent Usually pass nil; if you are mounting a subtree of a layout, pass the parent component so the
```
darknoon: Just a thought: is this parameter necessary, since component responder chain bubbles onto view responder chain—could then be directed to the component responder chain again if there is one above the current level right?

benlodotcom:
I like it !

IMO we should also eventually attach a "lifecycle manager" on the root view that will tie the lifecycle of the component tree to the one of its root view. Right now we are in this strange in between state were each view is tied to its component but you still explicitly need to unmount when the view goes away (for example from a component collection view data source). If I have time I'll put up a quick RFC in this sense.

The other solution is to completely detach the lifecyle of views and components but even though it sounds cleaner this is not practical at all.

ComponentKit/Core/CKComponentLayout.h
adamjernst: Yeah that's true. Maybe a followup could just do that.

ComponentKit/Core/CKComponentLayout.h
adamjernst: Not a bad idea for a followup. I'll land this as an incremental solution.

PR #279, opened 2015-06-29
Refactor ComponentHostingView

adamjernst:
- No longer uses `CKComponentLifecycleManager`, freeing us from that cruft.
- If only the size of the view changes, but not the model or context, the component is not regenerated. This is more efficient.
- Similarly we cache the computed layout, so if for some reason `layoutSubviews` is called twice with no change to the component or size, we won't expensively re-layout the component.
- Added thread safety asserts.

Next up is to add asynchronous modes for model, state, and context updates.

### DIFF
--- a/ComponentKit/HostingView/CKComponentHostingView.mm
+++ b/ComponentKit/HostingView/CKComponentHostingView.mm
@@ -16,18 +16,26 @@
 
 #import "CKComponentAnimation.h"
 #import "CKComponentHostingViewDelegate.h"
-#import "CKComponentLifecycleManager.h"
-#import "CKComponentLifecycleManager_Private.h"
 #import "CKComponentRootView.h"
+#import "CKComponentScopeRoot.h"
 #import "CKComponentSizeRangeProviding.h"
+#import "CKComponentSubclass.h"
 
-@interface CKComponentHostingView () <CKComponentLifecycleManagerDelegate>
+@interface CKComponentHostingView () <CKComponentStateListener>
 {
-  CKComponentLifecycleManager *_lifecycleManager;
+  Class<CKComponentProvider> _componentProvider;
   id<CKComponentSizeRangeProviding> _sizeRangeProvider;
-  CKComponentRootView *_containerView;
+
+  CKComponentScopeRoot *_scopeRoot;
+  CKComponentStateUpdateMap _pendingStateUpdates;
+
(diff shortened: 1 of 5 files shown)


ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -40,108 +48,127 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+  CKAssertMainThread();
   if (_context != context) {
     _context = context;
-    [self setNeedsLayout];
+    [self _setComponentNeedsUpdate];
   }
 }
 
-- (UIView *)containerView
+- (const CKComponentLayout &)mountedLayout
```
benlodotcom: What's the reason to expose that ?

darknoon: This is used in diamond; I made my own equivalent of CKComponentHostingView to be able to grab the layout. I do queries like hit test / bounding boxes / etc on this data structure.

ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -40,108 +48,127 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+- (void)_setComponentNeedsUpdate
 {
+  if (!_componentNeedsUpdate) { // Avoid thrashing delegate
+    _componentNeedsUpdate = YES;
+    [self setNeedsLayout];
+    [_delegate componentHostingViewDidInvalidateSize:self];
+  }
+}
+
+- (void)_updateComponentIfNeeded
```
darknoon: In diamond, I have the equivalent of this being public API, because I need to be able to "force" an update before doing something else.

ComponentKit/HostingView/CKComponentHostingView.mm
benlodotcom: :+1:

adamjernst: It's also used one place in Facebook for iOS.

ComponentKit/HostingView/CKComponentHostingView.mm
adamjernst: Yeah, `CKTransactionalComponentDataSource` does have such an API as well. I'll probably add it in a follow-on then.

PR #280, opened 2015-06-29
Mac support via #defines, some #ifdeffing

darknoon:
This is preliminary, intented for commentary / review of basic approach.

A lot of the CKM classes are non-final, and some merging of API / functionality should really occur between iOS and mac text labels, buttons, etc.


ComponentKit.podspec
```diff
@@ -1,19 +1,28 @@
...
   s.version = "0.12"
-  s.summary = "A React-inspired view framework for iOS"
+  s.summary = "A React-inspired view framework for iOS and OS X"
   s.homepage = "https://componentkit.org"
-  s.authors = 'adamjernst@fb.com'
+  s.authors = 'adamjernst@fb.com', 'andrewpouliot@fb.com', 'beefon@fb.com',
   s.license = 'BSD'
   s.source = {
-    :git => "https://github.com/facebook/ComponentKit.git",
+    :git => "https://github.com/darknoon/ComponentKit-Mac.git",
```
adamjernst: Oops, revert

ComponentKit/Accessibility/CKComponentAccessibility.mm
```diff
@@ -24,7 +24,11 @@ static CKViewComponentAttributeValueMap ViewAttributesFromAccessibilityContext(c
     accessibilityAttributes[@selector(setAccessibilityIdentifier:)] = accessibilityContext.accessibilityIdentifier;
   }
   if (accessibilityContext.isAccessibilityElement) {
+#if TARGET_OS_IPHONE
     accessibilityAttributes[@selector(setIsAccessibilityElement:)] = accessibilityContext.isAccessibilityElement;
+#else
```
adamjernst: I might prefer else-if's and then an #else with #error unrecognized platform... but that's probably overkill

darknoon: That's reasonable. Apple's decision to make iOS also `TARGET_OS_MAC` defines the order, but I can make the else simply `TARGET_OS_MAC`.

adamjernst:
Looks good to me in general; closing until it builds properly, or I have time to fix builds.

PR #283, opened 2015-06-30
Asynchronous support for CKComponentHostingView

adamjernst:
State updates, model updates, and context updates can now be asynchronous. Layout will also be performed asynchronously on a speculative basis.


ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -130,33 +137,97 @@ - (void)componentScopeHandleWithIdentifier:(CKComponentScopeHandleIdentifier)glo
...
+
+  // Wait until the end of the run loop so that if multiple async updates are triggered we don't thrash.
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [self _scheduleAsynchronousUpdate];
+  });
+}
+
+- (void)_scheduleAsynchronousUpdate
+{
+  if (_componentNeedsUpdateType != CKComponentHostingViewUpdateTypeAsynchronous) {
```
natansh: I think the better way to check is `>`? If the update type is `none` then we're in bad territory and should assert.

natansh: Ah, maybe not. Sorry, I think I got the point of this... after a sync update we'll likely reset the flag.

ComponentKit/HostingView/CKComponentHostingView.h
```diff
@@ -16,6 +16,11 @@
 @protocol CKComponentHostingViewDelegate;
 @protocol CKComponentSizeRangeProviding;
 
+typedef NS_ENUM(NSUInteger, CKComponentHostingViewMode) {
```
darknoon: `CKComponentHostingViewUpdateMode` might be clearer. Could this be unified with the enum in the transactional data source?

adamjernst: @darknoon  Yes! I in fact do have plans to unify the enums in a follow-on diff.

adamjernst: (As `CKUpdateMode` or something similar.)

ComponentKit/Core/CKComponent.mm
```diff
@@ -275,21 +275,11 @@ + (id)initialState
...
-  [self _updateState:updateBlock tryAsynchronousUpdate:NO];
-}
-
-- (void)updateStateWithExpensiveReflow:(id (^)(id))updateBlock
-{
-  [self _updateState:updateBlock tryAsynchronousUpdate:YES];
-}
-
-- (void)_updateState:(id (^)(id))updateBlock tryAsynchronousUpdate:(BOOL)tryAsynchronousUpdate
+- (void)updateState:(id (^)(id))updateBlock mode:(CKUpdateMode)mode
```
benlodotcom: Should we have a stateUpdate method with default behavior (ie everything is async by default) ?

== Discussion around the API within components ==

Specifying sync and async from the component is definitely the most straightforward approach but it makes me a bit sad because:
- It couples components to a certain implementation of an hosting view and data source.
- It can be easy to take strange assumptions around the fact that we guarantee some state updates will be sync (already the case today though).
- It will prevent enforcing different policies than sync and async (e.g we could do our best for a state update to be sync but still scatter it across different run loops if needed). This point is far down the road I agree though. 

It feels like the decision between sync and async lies around two semantic factors :
- Prioritity of the interaction (should be defined by the component)
- Complexity of the state update (should be inferred by the infra)

What if we exposed in components land an API based on priority ? Inferring complexity is tricky (number of components in the tree) and depends on the context the component is presented in (e.g button in an h-scroll vs same button in a simple story) so we could assume homogeneous complexity for now and just reason in terms of priority of the interaction (e.g button tap high pri, state update due to incoming data coming from the server low-pri).

Thoughts ?

ComponentKit/Core/CKComponentSubclass.h
```diff
@@ -88,7 +89,7 @@ extern CGSize const kCKComponentParentSizeUndefined;
    return [nextState copy]; // immutable! :D
  }];
  */
-- (void)updateState:(id (^)(id))updateBlock;
+- (void)updateState:(id (^)(id))updateBlock mode:(CKUpdateMode)mode;
```
benlodotcom: A bit of doc about the mode, maybe just `@see CKUpdateMode`

ComponentKit/Core/CKComponent.mm
adamjernst: Yeah, this is a super interesting question. I've discussed it with @modocache, @ocrickard, and many others (including a Facebook-internal post). There is absolutely no consensus on what to do.

Having two separate methods is a straightforward option; e.g. `updateState:` (async) and `synchronouslyUpdateState:` (sync). But now we have bifurcating implementations and it's not entirely obvious to the caller *how* the two are different.

I don't like the idea of exposing a new semantic concept like "priority." This is just another indirect mapping that the caller has to understand. So, mark me down against that one.

In the end, I don't like the idea of forcing everyone to think about it at every point either, but at least the mode has exact symmetry with the updates that you give to a ComponentHostingView or a transactional data source. They behave identically, which makes it entirely obvious what is happening.

ComponentKit/Core/CKComponentSubclass.h
adamjernst: Sure.

ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -126,49 +128,114 @@ - (void)setContext:(id<NSObject>)context
...
+    const CKBuildComponentResult result = CKBuildComponent(scopeRootCopy, *stateUpdatesToApply, ^CKComponent *{
+      return [_componentProvider componentForModel:modelCopy context:contextCopy];
+    });
+    dispatch_async(dispatch_get_main_queue(), ^{
+      if (!_componentNeedsUpdate) {
+        // A synchronous update snuck in and took care of it for us.
+        return;
+      }
+
+      if (_currentModel == modelCopy
```
benlodotcom: I'd had a bit more comments regarding the fact that state updates will run concurrently and hence we could get out of order execution and this is why we need this check.

adamjernst: Adding comments sounds great to me

ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -126,49 +128,114 @@ - (void)setContext:(id<NSObject>)context
...
+        _component = result.component;
+        _scopeRoot = result.scopeRoot;
+        _currentModel = modelCopy;
+        _currentContext = contextCopy;
+        _componentNeedsUpdate = NO;
+        _scheduledAsynchronousComponentUpdate = NO;
+        [self setNeedsLayout];
+        [_delegate componentHostingViewDidInvalidateSize:self];
+      } else {
+        [self _scheduleAsynchronousUpdate];
```
benlodotcom: I think we can theoretically keep retrying forever if for instance a model update is done at the right rate. The likelihood of this happening is very low though.

adamjernst: Sure, but if you're doing a model update at that rate and we can't compute components fast enough, there's not much we can do right?

PR #285, opened 2015-07-02
Add hashing functions to core component things

darknoon:
Needed to memoize on them, will not compile until memoization diff lands.

Fixes #285

### DIFF
--- a/ComponentKit/Core/CKDimension.mm
+++ b/ComponentKit/Core/CKDimension.mm
@@ -16,6 +16,8 @@
 
 #import "ComponentLayoutContext.h"
 #import "CKMacros.h"
+#import "CKInternalHelpers.h"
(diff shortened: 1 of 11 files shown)


ComponentKit/Core/CKDimension.mm
```diff
@@ -56,6 +57,15 @@
   }
 }
 
+size_t std::hash<CKRelativeDimension>::operator ()(const CKRelativeDimension &size) {
+  size_t subhashes[] = {
+    (size_t)(size._type),
+    std::hash<CGFloat>()(size._value),
+  };
+  return CKIntegerArrayHash(subhashes, std::end(subhashes) - std::begin(subhashes));
```
darknoon: eventually, I want to replace CKIntegerArrayHash with a c++ macro to avoid this

PR #317, opened 2015-07-20
Facilitate the inclusion of additional information during layout.

eczarny:
Adds an optional `NSDictionary` field called `extra` to CKComponentLayout that can be used to contain additional information during layout.

This change also pins CocoaPods and Slather to specific versions. Previously Travis CI would install the latest versions which may not always work well together (for some reason or another). Using specific versions should lead to a more deterministic build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: objective-c
 
 before_install:
   - gem install activesupport
-  - gem update cocoapods
-  - gem install slather
+  - gem install cocoapods -v 0.37.2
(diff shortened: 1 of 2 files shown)


adamjernst:
Weird, there's some obscure error from travis. sorry @eczarny  can you investigate? We should get a clean build.

adamjernst:
Actually nm this is affecting the main build too.

eczarny:
I think the upgrade to CocoaPods 0.38.0 did it. I can try explicitly installing 0.37.2 instead.

Issue #271, opened 2015-06-25
didSelectItemAtIndexPath not called


dylanslewis:
I've asked my original question here:
https://github.com/facebook/componentkit/issues/149

I have tried to create a super basic component with a view with no userInteractionEnabled, which just has a red background and a size of {100, 100}, meaning it is not a problem of another view blocking the touch.
On the ViewController which creates the component, I have implemented the method:
- (void)dataSource:(CKCollectionViewDataSource *)dataSource didSelectItemAtIndexPath:(NSIndexPath *)indexPath

However, this is never called when I tap on the component. Has anyone else had experience of this?

Thanks

smiLLe:
```
- (void)collectionView:(UICollectionView *)collectionView didSelectItemAtIndexPath:(NSIndexPath *)indexPath
{
  [_dataSource modelForItemAtIndexPath:indexPath];
}
```

dylanslewis:
Amazing, it's working now. Strange that the CKCollectionViewDataSource method didn't work (which was recommended in the docs)...

PR #320, opened 2015-07-23
fixed UICollectionViewDelegate API usage

leoschweizer:
As mentioned in #271, the API shown in the docs (`dataSource:didSelectItemAtIndexPath:`) does not actually exist. I've updated the docs to use the `UICollectionViewDelegate`-API.


adamjernst:
thanks!

PR #328, opened 2015-07-30
Add support for capping the maximum size of the stateful view reuse pool

dyesno:
... to a value defined by the CKStatefulViewComponentController class. Add related
unit test.

### DIFF
--- a/ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
@@ -7,6 +7,7 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  *
  */
+#import "CKStatefulViewComponentController.h"
 
 #import "CKStatefulViewReusePool.h"
 
@@ -30,6 +31,11 @@ public:
       [allViews removeObject:view];
     }
     return view;
+  };
+  
+  NSUInteger viewCount()
+  {
+    return [allViews count];
   };
 
   void addView(UIView *view)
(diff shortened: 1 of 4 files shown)


ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -94,8 +100,24 @@ - (void)enqueueStatefulView:(UIView *)view
...
+    // only add the stateful view to the pool if a maximum limit has not been set
+    return _pool[std::make_pair(controllerClass, context)].addView(view);
+  }
+}
 
-  return _pool[std::make_pair(controllerClass, context)].addView(view);
+- (NSInteger)_sizeForControllerClass:(Class)controllerClass
+                             context:(id)context
+{
+  const auto it = _pool.find(std::make_pair(controllerClass, context));
```
adamjernst: This is a little wasteful because we're making the pair and looking it up in the pool twice.

Better might be to get rid of this helper method entirely and take an iterator into the pool like so:

```
auto it = _pool[std::make_pair(controllerClass, context)];
if (maximumPoolSize < 0 || it->second.viewCount() < maximumPoolSize) {
  it->second.addView(view);
}
```

But, it's your call.

ComponentKitTests/StatefulViews/CKStatefulViewReusePoolTests.mm
```diff
@@ -120,6 +131,35 @@ - (void)testEnqueueingViewThenDequeueingWithDifferentContextReturnsNewView
   XCTAssertTrue(firstView != dequeuedView, @"Expected different view to be vended.");
 }
 
+- (void)testMaximumPoolSizeOfOneByEnqueueingTwoViewsThenDequeueingTwoViewsReturnsNewViwe
```
adamjernst: Tests! You're my hero.

PR #337, opened 2015-08-05
Add appending to CKTransactionalComponentDataSourceChangesetBuilder

itsthejb:
Hi all,

Have just been looking at the recently merged transactional data source changes, and from a third-party point of view, a few things popped out. I have a bit of time to propose some ideas, so please let me know what you think.

In the case of `CKTransactionalComponentDataSourceChangesetBuilder`, I noticed that whilst the implementation supports chaining, since all the properties are overwritten in the convenience methods it's not possible to append inserts/removals/moves, etc. This would mean that when looping over a collection of models one would have to create a mutable container themselves and then pass it to the builder, which rather detracts from the convenience it should provide. Also, if users assume that appends are possible they will do a lot of head-scratching wondering why their updates aren't applied.

I wanted to avoid unnecessary use of `-mutableCopy`, in case the user doesn't make any appends. This, means the methods being a little longer than they might be otherwise, and contain a lot of repeated boiler-plate. I considered using a macro to simplify this, but thought perhaps you would prefer them written out in full.

This also means that the properties could be either mutable or immutable, depending on whether an append has been performed. As such, I had to change the ivars to `id` in order to satisfy the compiler. Since the initialiser for `CKTransactionalComponentDataSourceChangeset` makes copies, I think this is the most efficient way to add the mutable functionality without doing any unnecessary copying.

In any case, please let me know what you think!

PS. No documentation regarding this feature added *as yet*, but could be if you think it necessary.

### DIFF
--- a/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceChangeset.m
+++ b/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceChangeset.m
@@ -35,21 +35,75 @@
 
 @implementation CKTransactionalComponentDataSourceChangesetBuilder
 {
-  NSDictionary *_updatedItems;
-  NSSet *_removedItems;
-  NSIndexSet *_removedSections;
-  NSDictionary *_movedItems;
-  NSIndexSet *_insertedSections;
-  NSDictionary *_insertedItems;
+  id _updatedItems;
+  id _removedItems;
+  id _removedSections;
+  id _movedItems;
+  id _insertedSections;
+  id _insertedItems;
 }
 
 + (instancetype)transactionalComponentDataSourceChangeset { return [[self alloc] init]; }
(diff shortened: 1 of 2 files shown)


adamjernst:
Interesting choice. On one hand, this makes the value object a little less straightforward, but I agree it also is more in line with what I'd expect.

Let me think it over for a day and see what we should do!

itsthejb:
:+1:

itsthejb:
Thinking about this, what would be *really great* would be to implement a builder DSL in the spirit of something like [Masonry](https://github.com/SnapKit/Masonry), so you could express something like:

```
[builder buildChangeSet:^(Builder *changeSet) {
    do.insert.section(0);
    do.insert.section(0).item(1).with(object);
    do.move.section(1).to.section(2);
    ...
}];
```

That would be great.

adamjernst:
I want to keep the actual value objects simple. I would definitely welcome a PR to create some sort of append-builder though!

itsthejb:
Gotcha. Builder is nice. When we migrate to the transactional data source that's a possible

PR #338, opened 2015-08-05
Add -modelForItemAtIndexPath to CKCollectionViewTransactionalDataSource

itsthejb:
Hi all,

I noticed that `CKCollectionViewTransactionalDataSource` lacks an equivalent to `CKCollectionViewDataSource`'s `-modelForItemAtIndexPath:` method. Seems like an accidental omission to me, since the implementation is trivial, and it would certainly be a necessity for us to migrate to the new class. So... here it is here.


adamjernst:
Seems reasonable!

PR #339, opened 2015-08-05
Snapshot test case updates

itsthejb:
Hi all,

Some minor updates to the podpsecs that currently cause pain on `master` (which we're using...). The podfile doesn't limit the version of `FBSnapshotTestCase`, so some minor changes required.

### DIFF
--- a/ComponentSnapshotTestCase/ComponentSnapshotTestCase.podspec
+++ b/ComponentSnapshotTestCase/ComponentSnapshotTestCase.podspec
@@ -8,6 +8,6 @@ Pod::Spec.new do |s|
   s.platform     = :ios, '7.0'
   s.requires_arc = true
   s.source_files = '**/*.h', '**/*.m', '**/*.mm'
-  s.dependency 'FBSnapshotTestCase'
+  s.dependency 'FBSnapshotTestCase/Core'
   s.frameworks = 'UIKit', 'XCTest'
 end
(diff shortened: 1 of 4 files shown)


ComponentSnapshotTestCase/ComponentSnapshotTestCase.podspec
```diff
@@ -8,6 +8,6 @@ Pod::Spec.new do |s|
   s.platform     = :ios, '7.0'
   s.requires_arc = true
   s.source_files = '**/*.h', '**/*.m', '**/*.mm'
-  s.dependency 'FBSnapshotTestCase'
+  s.dependency 'FBSnapshotTestCase/Core'
```
itsthejb: Subspec available for Swift since https://github.com/facebook/ios-snapshot-test-case/commit/5508b49217857bd3c01eef84f6015c87a333872a or thereabouts. This fails without `use_frameworks!`. Since ComponentKit is a Swift-free zone (and likely to remain so), so seems to make life a lot easier to just use the core (ObjC-only) subspec.

ComponentTextKitApplicationTests/CKTextKitTests.mm
```diff
@@ -75,7 +75,7 @@ static BOOL checkAttributes(const CKTextKitAttributes &attributes, const CGSize
   FBSnapshotTestController *controller = [[FBSnapshotTestController alloc] init];
   UIImage *labelImage = UITextViewImageWithAttributes(attributes, constrainedSize);
   UIImage *textKitImage = CKTextKitImageWithAttributes(attributes, constrainedSize);
-  return [controller compareReferenceImage:labelImage toImage:textKitImage error:nil];
+  return [controller compareReferenceImage:labelImage toImage:textKitImage tolerance:0 error:nil];
```
itsthejb: After breaking change in https://github.com/facebook/ios-snapshot-test-case/pull/86, the `tolerance` argument needs to be added in a few places.

adamjernst:
Thanks.

PR #341, opened 2015-08-05
Fixed dcomponents to work with CKComponentHostingView

Goom11:
+ Previously dcomponents looked for a lifecycleManager but since componentkit is currently transitioning away from this and towards using CKComponentHostingView, dcomponents no longer worked in certain cases.
+ This pull request introduces a fix for those cases where a lifecycleManager does not exist but a CKComponentHostingView does.

The way the fix works is that if there is a CKComponentHostingView, this DebugController calls setNeedsLayout on the CKComponentHostingView. This may not be the best way to do this. @adamjernst may know a better one.

### DIFF
--- a/ComponentKit/Debug/CKComponentDebugController.mm
+++ b/ComponentKit/Debug/CKComponentDebugController.mm
@@ -14,9 +14,12 @@
 
 #import "CKComponent.h"
 #import "CKComponentAnimation.h"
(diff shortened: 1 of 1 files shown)


natansh:
Seems OK to me to do it this way, especially since it doesn't use any private API's.

PR #343, opened 2015-08-07
Fixed a bug with pcomponents

Goom11:
+ Fixed a bug with pcomponents where if showViews was NO then it wouldn't print anything at all.
+ It now traverses down the view hierarchy without printing out the views and then continues with normal behavior once it hits CKComponentRootView.
+ Also updated some of the indentation on computeDescription and recursiveDescriptionForLayout
+ Replaced all instances of {0, 0} with CGSizeZero/CGPointZero

### DIFF
--- a/ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
+++ b/ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
@@ -73,7 +73,12 @@ static NSString *ancestorComponentHierarchyDescriptionForView(UIView *view, BOOL
                                  withString:indentString
                             startingAtIndex:0];
     } else {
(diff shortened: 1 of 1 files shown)


ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
```diff
@@ -73,7 +73,12 @@ + (NSString *)componentHierarchyDescriptionForView:(UIView *)view searchUpwards:
                                  withString:indentString
                             startingAtIndex:0];
     } else {
-      NSString *rootViewDescription = computeDescription(nil, rootView, {0, 0}, {0, 0}, @"", YES);
+      NSString *rootViewDescription = computeDescription(nil,
```
natansh: Feels a bit off. Was it actually going beyond 120 characters?

Issue #316, opened 2015-07-20
Not clear how to account for state of sub-components in memoization


darknoon:
This is probably the biggest drawback of the system so far. Filing to discuss with @adamjernst.

ocrickard:
I'll be taking this on.

PR #348, opened 2015-08-17
Try to allow memoization of components with state

darknoon:
Work in progress, via discussions with @adamjernst. Need to clean up the `id scopeIdentifier` parameter, which shouldn't be necessary.

### DIFF
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -300,4 +300,9 @@ static void *kRootComponentMountedViewKey = &kRootComponentMountedViewKey;
   return _scopeHandle ? @(_scopeHandle.globalIdentifier) : nil;
 }
 
+- (CKComponentScopeHandleIdentifier)globalIdentifier
+{
+  return _scopeHandle.globalIdentifier;
+}
+
 @end
--- a/ComponentKit/Core/CKComponentMemoizer.mm
+++ b/ComponentKit/Core/CKComponentMemoizer.mm
@@ -14,10 +14,46 @@
 #import "CKComponentSubclass.h"
 #import "CKMacros.h"
 #import "CKInternalHelpers.h"
+#import "CKThreadLocalComponentScope.h"
+#import "CKComponentScopeHandle.h"
 
 #include <map>
 
+
 static NSString *CKComponentMemoizerThreadKey = @"CKComponentMemoizer";
+
+static bool CKCanUseMemoizedComponent(CKComponent *c) {
+  CKThreadLocalComponentScope *s = CKThreadLocalComponentScope::currentScope();
+
+  NSCAssert(s != nullptr, @"No thread local scope!");
+  if (!s) {
+    return false;
+  }
+
+  auto stackTop = s->stack.top();
+  CKComponentScopeHandleIdentifier oldIdent = c.globalIdentifier;
+
+  // Look for the equivalent child frame
+  auto oldFrameRet = [stackTop.equivalentPreviousFrame childScopeFrameForOldIdentifier:oldIdent];
+  CKComponentScopeFrame *oldScopeFrame = oldFrameRet.frame;
+
+  // Make sure none of the descendents of the oldScopeFrame have pending state updates
+  if (oldScopeFrame && s->stateUpdates.size() > 0) {
+    if ([oldScopeFrame anyDescendentHasPendingUpdates:s->stateUpdates]) {
+      //NSLog(@"Cannot reuse component %@ because state has been updated.", c);
+      return false;
+    }
+  }
+
+  // Copy oldScopeFrame into new scope frame
+  if (oldScopeFrame) {
+    [stackTop.frame injectScopeHierarchy:oldFrameRet];
+  }
+
+  //NSLog(@"Ok to reuse component %@.", c);
+
+  return true;
+}
 
 // Define hash as just pulling out the precomputed hash field
 namespace std {
@@ -149,6 +185,11 @@ id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void))
 {
   _CKComponentMemoizerImpl *impl = [_CKComponentMemoizerImpl currentMemoizer];
   CKComponent *component = [impl dequeueComponentForKey:memoizationKey];
+  if (component) {
+    if (!CKCanUseMemoizedComponent(component)) {
+      component = nil;
+    }
+  }
   if (!component && block) {
     component = block();
   }
--- a/ComponentKit/Core/Scope/CKComponentScopeFrame.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeFrame.h
@@ -25,7 +25,7 @@ struct CKComponentScopeFramePair {
 
 + (CKComponentScopeFramePair)childPairForPair:(const CKComponentScopeFramePair &)pair
                                       newRoot:(CKComponentScopeRoot *)newRoot
-                               componentClass:(Class)aClass
+                               componentClass:(Class)componentClass
                                    identifier:(id)identifier
                           initialStateCreator:(id (^)(void))initialStateCreator
                                  stateUpdates:(const CKComponentStateUpdateMap &)stateUpdates;
@@ -33,5 +33,17 @@ struct CKComponentScopeFramePair {
 - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle;
 
 @property (nonatomic, strong, readonly) CKComponentScopeHandle *handle;
+
+struct _CKChildScopeFrameRet {
+  Class cls;
+  id identifier;
+  CKComponentScopeFrame *frame;
+};
+
+- (_CKChildScopeFrameRet)childScopeFrameForOldIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier;
+
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame;
+
+- (BOOL)anyDescendentHasPendingUpdates:(const CKComponentStateUpdateMap &)stateUpdates;
 
 @end
--- a/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
@@ -23,14 +23,14 @@
 #import "CKMacros.h"
 #import "CKThreadLocalComponentScope.h"
 
-typedef struct _CKStateScopeKey {
+struct _CKStateScopeKey {
   Class __unsafe_unretained componentClass;
   id identifier;
 
   bool operator==(const _CKStateScopeKey &v) const {
     return (CKObjectIsEqual(this->componentClass, v.componentClass) && CKObjectIsEqual(this->identifier, v.identifier));
   }
-} _CKStateScopeKey;
+};
 
 namespace std {
   template <>
@@ -86,6 +86,42 @@ namespace std {
     _handle = handle;
   }
   return self;
+}
+
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
+{
+  Class componentClass = oldFrame.cls;
+  id identifier = oldFrame.identifier;
+  CKAssert([oldFrame.frame isKindOfClass:[CKComponentScopeFrame class]], @"Can't inject with incorrect frame class.");
+
+  auto result = _children.insert({{componentClass, identifier}, oldFrame.frame});
+
+  CKAssert(result.second, @"Scope collision: attempting to create duplicate scope %@:%@", componentClass, identifier);
+}
+
+- (_CKChildScopeFrameRet)childScopeFrameForOldIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier
+{
+  for (auto &pr : _children) {
+    if (pr.second->_handle.globalIdentifier == globalIdentifier) {
+      return _CKChildScopeFrameRet{pr.first.componentClass, pr.first.identifier, pr.second};
+    }
+  }
+  return {nil, nil, 0};
+}
+
+- (BOOL)anyDescendentHasPendingUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+{
+  auto it = stateUpdates.find(_handle.globalIdentifier);
+  if (it != stateUpdates.end()) {
+    return YES;
+  }
+  // Look in children
+  for (auto &pr : _children) {
+    if ([pr.second anyDescendentHasPendingUpdates:stateUpdates]) {
+      return YES;
+    }
+  }
+  return NO;
 }
 
 @end
--- a/ComponentKit/Core/CKComponentInternal.h
+++ b/ComponentKit/Core/CKComponentInternal.h
@@ -13,6 +13,7 @@
 #import <ComponentKit/ComponentMountContext.h>
 #import <ComponentKit/CKComponent.h>
 #import <ComponentKit/CKComponentLayout.h>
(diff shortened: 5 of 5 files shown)


darknoon:
Related to #316.

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,41 @@
...
 #import "CKMacros.h"
 #import "CKInternalHelpers.h"
+#import "CKThreadLocalComponentScope.h"
+#import "CKComponentScopeHandle.h"
 
 #include <map>
 
 static NSString *CKComponentMemoizerThreadKey = @"CKComponentMemoizer";
 
+static bool CKCanUseMemoizedComponent(CKComponent *c, id scopeFrameIdentifier) {
```
adamjernst: Instead, use `c.scopeHandle.rootIdentifier` which is a unique identifier for its scope frame that persists across state updates.

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,41 @@
...
+  NSCAssert(s != nullptr, @"No thread local scope!");
+
+  if (!s) {
+    return false;
+  }
+
+  auto stackTop = s->stack.top();
+
+  // Look for the equivalent child frame
+  CKComponentScopeFrame *oldScopeFrame = [stackTop.equivalentPreviousFrame scopeFrameForClass:c.class identifier:scopeFrameIdentifier];
```
adamjernst: You can just make this `scopeFrameForHandleWithGlobalIdentifier:` then. Don't even need to pass in `c.class`.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,29 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
   return self;
 }
 
+- (CKComponentScopeFrame *)scopeFrameForClass:(Class)componentClass identifier:(id)identifier;
+{
+  auto it = _children.find({componentClass, identifier});
```
adamjernst: This is going to be relatively less efficient when you're doing it based on global identifiers instead of class/identifiers; now it'll be an O(N) search instead of O(1). But given that the number of children will typically be ~5 at most, that should be just fine.

darknoon: Updating the diff now; slightly concerned about complexity but it seems to work for my use case at the moment.

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,47 @@
...
 
 #include <map>
 
+
 static NSString *CKComponentMemoizerThreadKey = @"CKComponentMemoizer";
 
+static bool CKCanUseMemoizedComponent(CKComponent *c) {
+  CKThreadLocalComponentScope *s = CKThreadLocalComponentScope::currentScope();
+
+  NSCAssert(s != nullptr, @"No thread local scope!");
```
darknoon: I should use CKAssert or whatever here

ComponentKit/Core/Scope/CKComponentScopeFrame.h
```diff
@@ -34,4 +34,16 @@ struct CKComponentScopeFramePair {
 
 @property (nonatomic, strong, readonly) CKComponentScopeHandle *handle;
 
+struct _CKChildScopeFrameRet {
```
darknoon: This is somewhat inelegant, but I need to return more than just one of these items… thoughts?

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
   return self;
 }
 
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
+{
+  Class componentClass = oldFrame.cls;
+  id identifier = oldFrame.identifier;
+  CKAssert([oldFrame.frame isKindOfClass:[CKComponentScopeFrame class]], @"Can't inject with incorrect frame class.");
+
+  auto result = _children.insert({{componentClass, identifier}, oldFrame.frame});
```
darknoon: Please review this carefully, I'm not familiar enough with the code to know it it's really safe.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
...
+    if (pr.second->_handle.globalIdentifier == globalIdentifier) {
+      return _CKChildScopeFrameRet{pr.first.componentClass, pr.first.identifier, pr.second};
+    }
+  }
+  return {nil, nil, 0};
+}
+
+- (BOOL)anyDescendentHasPendingUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+{
+  auto it = stateUpdates.find(_handle.globalIdentifier);
```
darknoon: Is there a way we could use immutable ptrs to avoid this recursion in the no-change case, which should be very common?

adamjernst:
Reviewing now...

ComponentKit/Core/CKComponent.mm
```diff
@@ -300,4 +300,9 @@ - (id)scopeFrameToken
   return _scopeHandle ? @(_scopeHandle.globalIdentifier) : nil;
 }
 
+- (CKComponentScopeHandleIdentifier)globalIdentifier
+{
+  return _scopeHandle.globalIdentifier;
```
adamjernst: This already exists as `scopeFrameToken`, can we just use that?

ComponentKit/Core/CKComponentMemoizer.mm
adamjernst: Yup

ComponentKit/Core/CKComponent.mm
adamjernst: Ah I see why you didn't and exposed this instead. Will consider...

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
adamjernst: Not as far as I can see because `stateUpdates` is just a flat map of ids -> updates.

We could annotate each node in the state tree with a hash-set of all the IDs below it, which would make this an O(1) operation here, but given the typical size of a state tree (10-15 elements) I don't think it's worth avoiding the recursion. For now let's do this, and come back to it if it's a hotspot (doubt it)

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
...
+  if (it != stateUpdates.end()) {
+    return YES;
+  }
+  // Look in children
+  for (auto &pr : _children) {
```
adamjernst: const auto &pr

will avoid copying, not that it matters much since they're just pairs

adamjernst: (the copy is necessary since `stateUpdates` is const, so to give you a non-const pr it must copy the pair)

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
...
+{
+  for (auto &pr : _children) {
+    if (pr.second->_handle.globalIdentifier == globalIdentifier) {
+      return _CKChildScopeFrameRet{pr.first.componentClass, pr.first.identifier, pr.second};
+    }
+  }
+  return {nil, nil, 0};
+}
+
+- (BOOL)anyDescendentHasPendingUpdates:(const CKComponentStateUpdateMap &)stateUpdates
```
adamjernst: descendant not descendent

ComponentKit/Core/Scope/CKComponentScopeFrame.h
adamjernst: I think it's fine.

- Move it to the top of the file
- Rename to `CKChildScopeHierarchy` (drop underscore, lose abbreviation, and I like your term "scope hierarchy" below better than "child scope frame")

ComponentKit/Core/Scope/CKComponentScopeFrame.h
```diff
@@ -34,4 +34,16 @@ struct CKComponentScopeFramePair {
...
+  Class cls;
+  id identifier;
+  CKComponentScopeFrame *frame;
+};
+
+- (_CKChildScopeFrameRet)childScopeFrameForOldIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier;
```
adamjernst: `childScopeHierarchyForScopeFrameToken:(id)scopeFrameToken`

Inside the implementation just check if the scope frame token is an NSNumber and unwrap it if so. Kinda sucks as it's a hidden coupling with CKComponent's `scopeFrameToken`, but it is better than exposing another method on CKComponent (even if it's internal-only)

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,47 @@
...
 #import "CKInternalHelpers.h"
+#import "CKThreadLocalComponentScope.h"
+#import "CKComponentScopeHandle.h"
 
 #include <map>
 
+
 static NSString *CKComponentMemoizerThreadKey = @"CKComponentMemoizer";
 
+static bool CKCanUseMemoizedComponent(CKComponent *c) {
```
adamjernst: nit: { on new line

nit: don't need "CK" prefix if it's static

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,47 @@
...
+  NSCAssert(s != nullptr, @"No thread local scope!");
+  if (!s) {
+    return false;
+  }
+
+  auto stackTop = s->stack.top();
+  CKComponentScopeHandleIdentifier oldIdent = c.globalIdentifier;
+
+  // Look for the equivalent child frame
+  auto oldFrameRet = [stackTop.equivalentPreviousFrame childScopeFrameForOldIdentifier:oldIdent];
```
adamjernst: might as well `const auto`

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -14,11 +14,47 @@
...
+  CKComponentScopeFrame *oldScopeFrame = oldFrameRet.frame;
+
+  // Make sure none of the descendents of the oldScopeFrame have pending state updates
+  if (oldScopeFrame && s->stateUpdates.size() > 0) {
+    if ([oldScopeFrame anyDescendentHasPendingUpdates:s->stateUpdates]) {
+      //NSLog(@"Cannot reuse component %@ because state has been updated.", c);
```
adamjernst: clean up these logs of course…

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -149,6 +185,11 @@ id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void))
 {
   _CKComponentMemoizerImpl *impl = [_CKComponentMemoizerImpl currentMemoizer];
   CKComponent *component = [impl dequeueComponentForKey:memoizationKey];
+  if (component) {
+    if (!CKCanUseMemoizedComponent(component)) {
```
adamjernst: can we just move this check inside `dequeue`? Or do you prefer to keep it here for cleanliness of the queue? your call

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
adamjernst: Yup it's safe. You're injecting a scope frame from the *old* tree, which by now is immutable.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
...
 
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
+{
+  Class componentClass = oldFrame.cls;
+  id identifier = oldFrame.identifier;
+  CKAssert([oldFrame.frame isKindOfClass:[CKComponentScopeFrame class]], @"Can't inject with incorrect frame class.");
+
+  auto result = _children.insert({{componentClass, identifier}, oldFrame.frame});
+
+  CKAssert(result.second, @"Scope collision: attempting to create duplicate scope %@:%@", componentClass, identifier);
```
adamjernst: yup, good.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
   return self;
 }
 
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
+{
+  Class componentClass = oldFrame.cls;
+  id identifier = oldFrame.identifier;
+  CKAssert([oldFrame.frame isKindOfClass:[CKComponentScopeFrame class]], @"Can't inject with incorrect frame class.");
```
adamjernst: why is this here? it's typechecked, right?

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
   return self;
 }
 
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
+{
+  Class componentClass = oldFrame.cls;
```
adamjernst: probably unnecessary; the compiler will do this for you in C++ if you're doing it for efficiency reasons. if you just prefer it for readability, carry on

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -88,4 +88,40 @@ - (instancetype)initWithHandle:(CKComponentScopeHandle *)handle
   return self;
 }
 
+- (void)injectScopeHierarchy:(_CKChildScopeFrameRet)oldFrame
```
adamjernst: pass const reference:

  - (void)injectChildScopeHierarchy:(const CKChildScopeHierarchy &)hierarchy;

adamjernst:
Perfect, let's clean it up and land it.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
darknoon: Passing const reference instead of by value is actually less efficient since it can't move in. Do you have a specific reason for this request?

adamjernst: Nope, carry on then

ComponentKit/Core/CKComponentMemoizer.mm
darknoon: Good question. I wrote some tests, so going to try changing it and re-running them. One thing is that we could probably leave the old component in the queue if it's not usable if I move the call into `dequeue`.

PR #350, opened 2015-08-20
Remove record mode left enabled in spec

itsthejb:
Hi all,

Just found this `recordMode` left enabled (by myself in #272) :blush: Removed.


adamjernst:
Good find — but it looks like a snapshot needs re-recording. Take a look?

itsthejb:
Ah! Ok, will do...

ocrickard:
Looks like the build is still failing.

itsthejb:
Yep - sorry wrestling a little with creating ref images for the right platform versions... Think I have it now

itsthejb:
Ok, got it :wink:

ocrickard:
Thanks @itsthejb !

PR #363, opened 2015-09-02
Fix assert inside block

mattjgalloway:
This assert is inside a block, so we should use the `CKCFailAssert` instead of plain `CKFailAssert`.


ocrickard:
My bad :P

ocrickard:
Will merge once the CI build finishes.

PR #366, opened 2015-09-03
Fix building with Xcode 7

mattjgalloway:
We need the override for building with Xcode 7, because it's more strict.


ocrickard:
Awesome, thanks.  Will merge once CI build finishes.

adamjernst:
Just did this manually. Sorry.

PR #371, opened 2015-09-09
Support alignment rect insets

darknoon:
Sometimes components like buttons want space for their frame that extends beyond the frame they wish to use for layout. This diff extends the component layout system to allow for such things. A classic example is NSTextField (~UILabel) on OS X, which requires 2pt of space on either side that isn't part of its nominal size.

On iOS you have the `alignmentRectInsets` property available on UIView that performs the same effect, thus I have named it likewise.


ComponentTextKit/CKTextComponent.mm
```diff
@@ -99,11 +99,13 @@ - (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
 
 - (CK::Component::MountResult)mountInContext:(const CK::Component::MountContext &)context
                                         size:(const CGSize)size
+                         alignmentRectInsets:(UIEdgeInsets)alignmentRectInsets
                                     children:(std::shared_ptr<const std::vector<CKComponentLayoutChild>>)children
                               supercomponent:(CKComponent *)supercomponent
 {
   CK::Component::MountResult result = [super mountInContext:context
                                                        size:size
+                                        alignmentRectInsets:alignmentRectInsets
```
darknoon: Is there any reason this doesn't just take the layout?
I wasn't sure if that would be breaking some sort of abstraction.

darknoon:
Context on the Mac side: https://github.com/darknoon/componentkitx/pull/6

adamjernst:
Let's discuss this in person.

adamjernst:
Gonna veto this for now to keep CK's layout super-simple, but I'm open to the concept as part of a wider-ranging wholistic change to the layout algorithm.

PR #379, opened 2015-09-15
CollectionView maintain context during large bounds change animation

suvarshibhadra:
During large bounds change animations (such as for large inline expansion collapses) the user completely loses context. The fix is to adjust the content offset of the collection view by an appropriate amount to keep sufficient context.

The user gets his sense of context/place from the 'element taking up the most screen real-estate prior to animation'. After the animation completes the user should -
a) see this element on screen
b) see this element take up as much screen real-estate as it was taking up previously. (as close as possible given its bounds may have changed)

This fix should work for horizontal/vertical scrolling collection views.


suvarshibhadra:
@adamjernst anything actionable here on my part or can this be merged in?

PR #381, opened 2015-09-16
[TDS][fix] remove the correct indexPath for moves

majak:
We have to remove the row at the index path the move originates at, not the index path it ends at.


adamjernst:
Can you add a test for this?

majak:
Funny thing is there was already a test using moves, but it has passed by dumb luck.

PR #382, opened 2015-09-16
[TDS] compute final index paths for updates

majak:
So far updates in `CKTransactionalComponentDataSourceAppliedChanges` have contained only index paths in pre-changeset-application world. This makes sense, since that's what you use to tell tableview/collection view what to update.

It's useful to know where updated rows ended up after the whole changeset has been applied (for reasons we don't let CV do reloads but we perform them manually). The only way to do gain that information is basically by performing a simulation of what TDS does internally.
That is an duplication of work and another opportunity to introduce a bug. Therefore I created this diff which makes TDS compute these post-changeset-application index paths for updates as well.

How does it work? We just record initial index path for each updated item. Then once we apply the whole changeset we look at every item in the new state whether we have an initial position recorded for it.

Caveat: This means a backwards incompatible change in `CKTransactionalComponentDataSourceAppliedChanges`'s interface.

cc @benlodotcom

### DIFF
--- a/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetModification.mm
+++ b/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetModification.mm
@@ -12,6 +12,8 @@
 
 #import <map>
 
+#import "CKArgumentPrecondition.h"
+#import "CKFunctor.h"
 #import "CKTransactionalComponentDataSourceConfiguration.h"
 #import "CKTransactionalComponentDataSourceStateInternal.h"
 #import "CKTransactionalComponentDataSourceChange.h"
@@ -56,6 +58,8 @@
   }];
 
   // Update items
+  // This structure is used for computing post-changeset-application index paths for updated items.
+  __block std::unordered_map<CKTransactionalComponentDataSourceItem *, NSIndexPath *, CK::HashFunctor<NSObject *>> itemToInitialIndexPathMap;
   [[_changeset updatedItems] enumerateKeysAndObjectsUsingBlock:^(NSIndexPath *indexPath, id model, BOOL *stop) {
     NSMutableArray *section = newSections[indexPath.section];
     CKTransactionalComponentDataSourceItem *oldItem = section[indexPath.item];
@@ -65,8 +69,9 @@
     });
     const CKComponentLayout layout = [result.component layoutThatFits:sizeRange parentSize:sizeRange.max];
 
-    [section replaceObjectAtIndex:indexPath.item withObject:
-     [[CKTransactionalComponentDataSourceItem alloc] initWithLayout:layout model:model scopeRoot:result.scopeRoot]];
+    CKTransactionalComponentDataSourceItem *newItem = [[CKTransactionalComponentDataSourceItem alloc] initWithLayout:layout model:model scopeRoot:result.scopeRoot];
+    [section replaceObjectAtIndex:indexPath.item withObject:newItem];
+    itemToInitialIndexPathMap[newItem] = indexPath;
   }];
 
   __block std::unordered_map<NSUInteger, std::map<NSUInteger, CKTransactionalComponentDataSourceItem *>> insertedItemsBySection;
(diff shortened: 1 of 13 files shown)


ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetModification.mm
```diff
@@ -56,6 +58,8 @@ - (CKTransactionalComponentDataSourceChange *)changeFromState:(CKTransactionalCo
   }];
 
   // Update items
+  // This structure is used for computing post-changeset-application index paths for updated items.
+  __block std::unordered_map<CKTransactionalComponentDataSourceItem *, NSIndexPath *, CK::HashFunctor<NSObject *>> itemToInitialIndexPathMap;
```
adamjernst: Does this really need to be `__block`?

adamjernst: Also looks like `CKFunctor<NSObject *>` is just casting the pointer, which is a fairly poor hash function…

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetModification.mm
```diff
@@ -142,6 +147,26 @@ - (CKTransactionalComponentDataSourceChange *)changeFromState:(CKTransactionalCo
...
+/**
+ This function computes a pre-changeset-application -> post-changeset-application index path map for updated items.
+ It does a linear pass over all items (in `newState`), and each item is looked up in a previously constructed map (`itemToInitialIndexPathMap`) of item -> pre-changeset-application index path.
+ If it exists we know it was updated and can easily create the mapping (since `newState` is post-changeset-application).
+ */
+static NSDictionary *computeUpdateIndexPaths(NSArray *newState, const std::unordered_map<CKTransactionalComponentDataSourceItem *, NSIndexPath *, CK::HashFunctor<NSObject *>> &itemToInitialIndexPathMap)
+{
+  __block NSMutableDictionary *updatedItemIndexPaths = [NSMutableDictionary dictionary];
+  [newState enumerateObjectsUsingBlock:^(NSArray *items, NSUInteger sectionIdx, BOOL *sectionStop) {
+    [items enumerateObjectsUsingBlock:^(CKTransactionalComponentDataSourceItem *item, NSUInteger index, BOOL *stop) {
```
adamjernst: A nested loop like this looks awfully expensive. I thought you said we were already computing this information? Why should core do this? :(

majak: Hmm sorry I wasn't clear enough, the main reason why I wanted to put it here is that we basically did the same changeset application work on level above just to get this information. But considering this solution I came up with  here is really not much reason why it should happen here in core. I'll think about it a bit more (state update fast passing in suspension might benefit from similar information) and will update this PR accordingly.

benlodotcom:
We don't really need the final indexPath for the update but mostly the datasource items corresponding to the updated indexPath (so that we can mount the new component tree on the updated cell) . We have two solutions here :
1) Require that any operation actually return this in a change, which I think would be fair since it is a computation that has to be done anyway and this just matches the change that we are passing to the transactional DS. In that case, we should then have each operation generate a changeset that is then applied by the infra to the state, instead of having the operation modifying the state itself we don't want to rely on the implementation of the operation to both compute the right final state and associate the right item to a change. Pretty big change overall and only necessary because of the way we deal with updates (a.k.a not using reloadItemAtIndexPath, but directly mounting the view on the cell).
2) Write a function that for every changeset get the updated item given the insertions, removals and moves that are in the same changeset. It is not entirely trivial and is an operation that is in the end redundant with the work the transactional datasource is already doing and at least O(nlogn) with the size of the changeset but it has the benefit to keep the transactional datasource simple.
3) For each update we find the right item in the datasource state (pretty much what is being done in this PR to find the "final indexPath"), safe, easy, but definitely not cheap.

I'm leaning heavily towards 1/, @adamjernst, @majak thoughts ?

majak:
1) seems the best to me, since there is no extra work involved and exactly solves our needs. We really care only about the updated item. Having the final index path would help us get to it, but if we can get the item without it we don't need it.

majak:
Outdated. Feel free to revive in case it would be necessary.

PR #386, opened 2015-09-24
Fix CKNetworkImageComponent flash when defaultImage changes

mcohnen:
If we already showed the image in CKNetworkImageComponent, if we download the same URL but the defaultImage is the same, the current code will assume it is a different image and set the defaultImage again and redownload the URL again.
There is no need to do that if the image is already downloaded. All we have to do is update the specifier to make sure if we need to show the defaultImage, we will use the new one when needed.

It is hard to snapshot test this since this is behavior inside CKNetworkComponentView. It will require creating a new 2nd component and make it use the same view. Happy to add a unit test if this is something possible.

### DIFF
--- a/ComponentKit/Components/CKNetworkImageComponent.mm
+++ b/ComponentKit/Components/CKNetworkImageComponent.mm
@@ -125,6 +125,12 @@
   if (CKObjectIsEqual(specifier, _specifier)) {
     return;
   }
+  
+  BOOL isShowingCurrentImageURL = self.image != _specifier.defaultImage && self.image != nil;
+  if (isShowingCurrentImageURL && CKObjectIsEqual(_specifier.url, specifier.url)) {
+    _specifier = specifier;
+    return;
+  }
 
   if (_download) {
     [_specifier.imageDownloader cancelImageDownload:_download];


ComponentKit/Components/CKNetworkImageComponent.mm
```diff
@@ -125,6 +125,12 @@ - (void)setSpecifier:(CKNetworkImageSpecifier *)specifier
   if (CKObjectIsEqual(specifier, _specifier)) {
     return;
   }
+  
+  BOOL isShowingCurrentImageURL = self.image != _specifier.defaultImage && self.image != nil;
+  if (isShowingCurrentImageURL && CKObjectIsEqual(_specifier.url, specifier.url)) {
+    _specifier = specifier;
+    return;
```
adamjernst: Breaks if cropRect changes.

I'd rather we model this with an explicit state enum and state machine.

mcohnen: CropRect is not taken into consideration in the isEqual method right now, so it is not breaking any existing functionality. I don't think we need a state machine, but instead we could check every property individually to decide what actions we should take in the setSpecifier: method:
- If cropRect changes, call setNeedsLayout.
- If the URL changes, cancel and start a new download.
- If the defaultImage changes, update the image if we were showing a defaultImage.

adamjernst:
Closing this for now, reopen upon editing!

PR #387, opened 2015-09-24
Delete legacy datasource

adamjernst:


benlodotcom:
In think we are still missing support for appearance in the transactional DS world. Wanna give it a go ?

conradev:
Besides appearance, I also found #469 and #477 to be missing in the in the transactional data source.

Is there anything else missing that needs to be implemented?

adamjernst:
closing for now, someone needs to finish this up so we can delete it

PR #388, opened 2015-09-29
Make FBSnapshotTestCase tolerance property accessible

itsthejb:
Hi all,

Just a simple one here - `FBSnapshotTestCase` has had the `tolerance` feature available since baae047998f6c3f5ace0d38daad5653bc3aee385. However, it's not easily accessibly from `CKComponentSnapshotTestCase`. We certainly do a lot of re-recording of reference images for things like major OS updates (we had loads for 8 -> 9) where inconsequential minor pixel changes happen, so it would be great to make use of this.

Seems fine to just have it as a property of the test case since it would seem to me that one would set this to a value that's a "policy" of sorts across the board.

Anyway, let me know!


adamjernst:
Nice one!

PR #391, opened 2015-10-04
DSL changeset builder

itsthejb:
Hi all,

As discussed in https://github.com/facebook/componentkit/pull/337, I finally got some time to do an implementation of blocks-based DSL changeset builder, in the vein of https://github.com/SnapKit/Masonry.

*This also adds support for moved sections, since I noticed this was missing*

Reasonably happy with the [syntax](https://github.com/eyeem/componentkit/blob/change-set-builder/ComponentKitTests/TransactionalDataSource/CKTransactionalComponentDataSourceChangesetBuilderTests.m#L33). It's a bit *wordy*, although some of the prepositions are optional. I only needed to add a little bit of macro assistance for `ck_removeItem` and `ck_moveItem` since I have no facility for optional arguments in the ObjC methods (AFAIK?), and the object argument for `-item:` here would naturally be `nil`. I added `ck_indexPath()` as well to make the indexPath method nicer to work with.

Tests aren't terribly exhaustive since I went with runtime assertions, mainly. This is perhaps more useful for the user, and certainly helped in the writing. In the tests I wanted to confirm that all expressions are valid and readable, and also provide a reference to the syntax. As such, they're mostly just examples of the possible expressions for the operations, and some shallow checking of the resulting changeset. I can certainly add more if necessary.

I think this isn't *too* complex. There are a couple of large-ish `switch` statements in there, which I don't see any way around. Compared to the previous builder there's a little more logic involved, but I think it's all basic conditionals and performance shouldn't be a concern since the syntax is pretty nice.

Currently I made this a separate class to `CKTransactionalComponentDataSourceChangesetBuilder`. This seemed like the simplest approach at first. Open for discussion, of course!

Thanks!


adamjernst:
OK, I've had some time to think about this and discuss it internally a bit. I'm a big fan of the terseness and readability this solution has, but I also want to keep the core of ComponentKit relentlessly small.

Are you willing to pull it into a separate cocoapod that others could depend on? I think that'd be the ideal end state — you can opt into this syntax, but the core stays small and simple.

If you create the cocoapod, I'd love to drop a link to it on the componentkit.org docs!

itsthejb:
Ok, fair enough. Will do that at some point :+1:

itsthejb:
@adamjernst Alright, now published an initial version of this as a separate pod: https://cocoapods.org/?q=CKCollectionViewDataSourceChangesetBuilder :+1:

PR #392, opened 2015-10-05
Some Xcode 7-oriented project fixes

itsthejb:
Hi all,

Just a few unglamourous fixes necessitated by Xcode7:

1. Key window with `nil` `rootViewController` now throws an exception, so threw a vanilla `UIViewController` in there.
2. Running tests now gives me `ObjectiveC.gcno` and `ObjectiveC.gcda` files in the project directory, so added those to `.gitignore`. This seems like a fine solution? Although I'm not sure why the change.
3. Applied Xcode's usual "recommended project updates" rigamarole. 

One that I haven't been able to fix, though, and which could be added here is `-[CALayer ck_isAsyncTransactionContainer]`: the method is missing when running with Xcode7 (irrespective of iOS 8 or 9), so *every* test in `ComponentTextKitApplicationTests` fails. I would fix it, but I can't for the life of me find out where this method is added for `CALayer`. There's no category, and I don't see any runtime use such as associated object or adding the method itself. If someone can point me in the right direction there, I can see about getting it fixed. Perhaps you guys aren't migrated yet?

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Example/WildeGuess/Pod
 appledoc
 *.gcov
 .DS_Store
+ObjectiveC.gc*
--- a/ComponentKit.xcodeproj/project.pbxproj
+++ b/ComponentKit.xcodeproj/project.pbxproj
@@ -506,7 +506,7 @@
 		B3EECEBA1AC2366500BFC5DA /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 0610;
+				LastUpgradeCheck = 0700;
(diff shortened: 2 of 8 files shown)


adamjernst:
Ah yes, the CALayer one is interesting—we were relying on the fact that CALayer used to be implicit key/value coding containers, and would just respond to `setFoo:` methods by updating its internal K/V map for foo. That's no longer true in iOS 9, so we need to use Obj-C associated objects or something.

The rest is very nice—thanks!

itsthejb:
Ok, cool. I might look into that at some point if I have a moment :+1:

PR #397, opened 2015-10-07
Use associated object storage for ck_asyncTransactionContainer

itsthejb:
As discussed in https://github.com/facebook/componentkit/pull/392, using `CALayer` as a key-value store no longer works under iOS 9. So simply provided explicit associated object storage for this property.


adamjernst:
Beautiful. Let's wait for travis and merge.

PR #398, opened 2015-10-07
Add moved sections support to CKTransactionalComponentDataSourceChangeset

itsthejb:
As discussed in https://github.com/facebook/componentkit/pull/391. I'll do the DSL builder as a separate project, but section moving support is definitely missing here, so added it.

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -70,6 +70,10 @@ static void applyChangesToCollectionView(CKTransactionalComponentDataSourceAppli
     NSIndexPath *to = changes.movedIndexPaths[from];
     [collectionView moveItemAtIndexPath:from toIndexPath:to];
   }
+  for (NSNumber *from in changes.movedSections) {
+    NSNumber *to = changes.movedSections[from];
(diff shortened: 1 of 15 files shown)


itsthejb:
Coveralls is insisting I add some tests, but this is a rather mundane addition, and there aren't any existing tests for similar behaviour, ie. that the change set passes on the expected modifications to a data source. Thoughts?

adamjernst:
I'm fine with skipping coveralls generally, although I do really want good test coverage overall.

itsthejb:
Sure. Well I can add something, but it would actually be rather a lot for such a small change here.

benlodotcom:
It doesn't seem like it will work right now, the state computed by CKTransactionalComponentDataSource is not taking into account moved sections, so we will end up with a UICollectionView that moves the sections but the backing state that doesn't so it won't work. See `CKTransactionalComponentDataSourceChangesetModification` to make the necessary changes.

Moving sections is a tricky beast so we should definitely add test coverage for it.

itsthejb:
Ah! Silly me for thinking this was trivial. Well, considering this is hardly a primary use case, perhaps it would be best to just close this and perhaps it's something that could come in the future? Admittedly, I've need seen a use case for moving sections!

itsthejb:
Ok, gonna close this :wink:

PR #402, opened 2015-10-09
Fixed dangling datasource pointer

eepetrash:
Prior iOS9 UIKit does not use ARC, we have to clear weak ref manually.

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewDataSource.mm
@@ -73,6 +73,11 @@ CK_FINAL_CLASS([CKCollectionViewDataSource class]);
   CK_NOT_DESIGNATED_INITIALIZER();
 }
 
+- (void)dealloc
+{
+    _collectionView.dataSource = nil;
+}
+
 #pragma mark - Changesets
 
 - (void)enqueueChangeset:(const CKArrayControllerInputChangeset &)changeset constrainedSize:(const CKSizeRange &)constrainedSize


ComponentKit/DataSources/CKCollectionViewDataSource.mm
```diff
@@ -73,6 +73,11 @@ - (instancetype)init
   CK_NOT_DESIGNATED_INITIALIZER();
 }
 
+- (void)dealloc
+{
+    _collectionView.dataSource = nil;
```
modocache: I may be missing something, but `-[UICollectionView dataSource]` is defined as a weak pointer:

```objc
@property (nonatomic, weak, nullable) id <UICollectionViewDataSource> dataSource;
```

Which means that when this instance is deallocated, the reference to it will be nil-ed out. Isn't that right? Or am I misunderstanding this change? :question:

modocache: > Prior iOS9 UIKit does not use ARC, we have to clear weak ref manually.

Ah, I see--sorry I missed this! Is this really the case, though? Can you link me to some documentation? I'm investigating myself as well.

modocache: Whoops! Sorry for the confusion--yup, this used to be `assign`, not `weak`:

```sh
cat Xcode-6.0.app/Contents/Developer/Documentation/DocSets/com.apple.adc.documentation.AppleiOS8.0.iOSLibrary.docset/Contents/Resources/Tokens/Objective-C/instp/UICollectionView/dataSource.xml

<?xml version="1.0" encoding="UTF-8"?>
<Property-ObjC AppleRefID="//apple_ref/occ/instp/UICollectionView/dataSource">
 <Name>dataSource</Name>
 <Abstract>
  <Para>The object that provides the data for the collection view.</Para>
 </Abstract>
 <Declaration>
  <CodeLine>@property(nonatomic, assign) id&lt;UICollectionViewDataSource&gt; dataSource</CodeLine>
 </Declaration>
 <Discussion>
  <Para>The data source must adopt the <uAPI url="documentation/UIKit/Reference/UICollectionViewDataSource_protocol/index.html#//apple_ref/occ/intf/UICollectionViewDataSource">UICollectionViewDataSource</uAPI> protocol. The collection view maintains a weak reference to the data source object.</Para>
 </Discussion>
 <Availability>
  <AvailabilityItem>Available in iOS 6.0 and later.</AvailabilityItem>
 </Availability>
 <HeaderFile>UICollectionView.h</HeaderFile>
</Property-ObjC>
```

modocache:
Thanks, @eepetrash! :bouquet:

ComponentKit/DataSources/CKCollectionViewDataSource.mm
eepetrash: Oh, you were faster :) https://developer.apple.com/library/prerelease/ios/releasenotes/General/iOS90APIDiffs/Objective-C/UIKit.html

PR #395, opened 2015-10-07
Add CKSupplementaryViewDataSource support to transactional data source

itsthejb:
Resolves https://github.com/facebook/componentkit/issues/394.

Ok, we found that problem! `CKCollectionViewTransactionalDataSource` doesn't forward supplementary view methods à la `CKCollectionViewDataSource`. Since we use section headers implemented previously with `CKCollectionViewDataSource` we return non-zero sizes for the expected supplementary views. `CKCollectionViewTransactionalDataSource` doesn't implement the supplementary view method and hence the collectionView starts playing with `nil` views.

Problem solved and QED. Trivial to implement. Please let me know any suggested adjustments.

PR #410, opened 2015-10-21
weak for component provider, and supplementary view data source

itsthejb:
Hi all :smiley: 

Looks like some retain cycles fixed here. The first is my bad: I prefer `@property` over direct ivar access, so forgot to make the supplementary view data source `__weak` in https://github.com/facebook/componentkit/pull/395. Fixed here. However, pretty sure that the `componentProvider` also needs to be weak since we have the following basic ownership:

untitled diagram

Therefore the obvious place to break the cycle is to have a `weak` `componentProvider`. Doesn't seem to be any danger here, and fixes the cycles in our case.

Let me know what you think!

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -26,7 +26,7 @@ CKTransactionalComponentDataSourceListener
 >
 {
   CKTransactionalComponentDataSource *_componentDataSource;
-  id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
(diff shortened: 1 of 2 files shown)


adamjernst:
Component providers are classes, and classes are singletons, so `__weak` would have no effect right?

adamjernst:
(The supplementary view data source on the other hand… that one definitely might make sense, cc @benlodotcom )

itsthejb:
Ah! Yes indeed. We're actually abusing that at the moment and passing instances, and I'd forgotten. Hence the problem. That's something we'll have to resolve at our end. The supplementary views, though, that definitely will be an instance and needs correcting. I can separate that out and make another PR.

PR #411, opened 2015-10-23
Weak supplementary view

itsthejb:
As discussed in https://github.com/facebook/componentkit/pull/410, this is the actual error that should be fixed. The other is our own integration problem.

Thanks!

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -26,13 +26,14 @@ CKTransactionalComponentDataSourceListener
 >
 {
   CKTransactionalComponentDataSource *_componentDataSource;
-  id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
+  __weak id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
   CKTransactionalComponentDataSourceState *_currentState;
   CKComponentDataSourceAttachController *_attachController;
 }
 @end
 
 @implementation CKCollectionViewTransactionalDataSource
+@synthesize supplementaryViewDataSource = _supplementaryViewDataSource;
 
 - (instancetype)initWithCollectionView:(UICollectionView *)collectionView
            supplementaryViewDataSource:(id<CKSupplementaryViewDataSource>)supplementaryViewDataSource
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.h
@@ -26,6 +26,7 @@
 
 /**
  @param collectionView The collectionView is held strongly and its datasource property will be set to the receiver.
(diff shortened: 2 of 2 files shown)


adamjernst:
@benlodotcom  can you review?

benlodotcom:
Yep sounds good to me, can you add a comment in the initializer that the supplementary view datasource will be held weakly. This is a bit unusual since it's passed in the intializer, I may also be open to making this a property. Thoughts ?

itsthejb:
Can do. Actually, since doing this, I had the thought that *technically* really should be using `__unsafe_unretained` rather than either `strong` or `weak` (http://arigrant.com/blog/2014/3/14/dont-retain-that-class). Granted, we're unfortunately still abusing the component provider at our end, but that would still be more "correct", and would at least work around our problem as well until I have a chance to make the necessary changes.

itsthejb:
Oh wait - I'm getting myself mixed up. The `__unsafe_unretained` comment related to https://github.com/facebook/componentkit/pull/410, not this. I still think that change might be a *tiny* bit warranted, though?

For this, the supplementary view data source would be an instance, so `weak` is definitely appropriate. As for documenting, certainly can do, although I would have thought the behaviour was expected since datasources are never retained in Cocoa....?

benlodotcom:
The reason for the comment is that it is a bit unusual to have a datasource passed in the initializer regarding standard cocoa practices (but immutability is better eventually), and usually dependencies passed in the initializer are retained strongly.

So because overall it is a bit unusual I think a small comment wouldn't hurt.

benlodotcom:
As for the component provider as it is a type Class<CKComponentProvider> clang will ensure that _unsafe_unretained is used and not strong, so the class won't be retained (we would have had a bad time otherwise).

itsthejb:
Ok - commented added. In doing so, noticed that a readonly property is lacking to mirror `CKCollectionViewDataSource`, so added it - the cloned comment led me in this direction. The propery doesn't seem terribly useful, but is harmless and would be in line with the total replacement of the `CKCollectionViewDataSource` API. Let me know what you think.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -26,13 +26,14 @@ @interface CKCollectionViewTransactionalDataSource () <
...
   CKTransactionalComponentDataSource *_componentDataSource;
-  id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
+  __weak id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
   CKTransactionalComponentDataSourceState *_currentState;
   CKComponentDataSourceAttachController *_attachController;
 }
 @end
 
 @implementation CKCollectionViewTransactionalDataSource
+@synthesize supplementaryViewDataSource = _supplementaryViewDataSource;
```
adamjernst: nit: do you really need to create the ivar given that the property is declared? I think you can remove both the ivar and the synthesize completely.

itsthejb: I *believe* this was necessary, since the ivar is explicitly defined....... But this was a while ago...

PR #414, opened 2015-10-24
Add move support to CKArrayController

conradev:
Both `UITableView` and `UICollectionView` support moving cells and sections. I want to support moving cells and sections within ComponentKit. This feature is not relevant for the news feed use case as much, but is for other user interfaces.

Moves operate like both an insertion and a deletion:
- You cannot delete an item at index path 1 and move an item from index path 1 at the same time
- You cannot insert an item at index path 2 and move an item to index path 2 at the same time

But a move is not like an insertion/deletion because it exhibits a different animation and preserves the cell that is being moved (if it's on screen).

I wanted to put this here as a working draft to see if there are any elements/approaches that should be changed. Right now it is missing the following:

- Throwing an exception for simultaneous move/delete/insert of items, and tests for that
- Throwing an exception for simultaneous move/delete/insert of sections, and tests for that

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewDataSource.mm
@@ -215,20 +215,26 @@ static void applyChangesetToCollectionView(const Output::Changeset &changeset, U
       case CKArrayControllerChangeTypeUpdate:
         [itemUpdateIndexPaths addObject:change.sourceIndexPath.toNSIndexPath()];
         break;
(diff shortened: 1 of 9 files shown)


benlodotcom:
Looks good to me, although with the new transactional datasource this structure won't be used anymore. 

Would be great to have the extra exceptions and tests, happy to merge this afterwards !

adamjernst:
Since this will die when the transactional data source becomes default anyway, I decided to just merge it.

PR #418, opened 2015-10-26
Fix hostingview initial size issue.

smalllixin:
The issue is if update model in first time the method componentHostingViewDidInvalidateSize is not called. So there's no chance to adjust the hostingview's size.
Simply change the update initial value for flag variable.

### DIFF
--- a/ComponentKit/HostingView/CKComponentHostingView.mm
+++ b/ComponentKit/HostingView/CKComponentHostingView.mm
@@ -71,7 +71,7 @@ struct CKComponentHostingViewInputs {
     _containerView = [[CKComponentRootView alloc] initWithFrame:CGRectZero];
     [self addSubview:_containerView];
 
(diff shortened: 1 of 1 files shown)


adamjernst:
Let me consider this — thanks!

facebook-github-bot-5:
It's been a while since the last commit was reviewed and the labels show this pull request needs review. Based on the blame information for the files in this pull request we identified @eczarny as a potential reviewer. Could you take a look please or cc someone with more context?

ocrickard:
CI failed, can we look at why?

ocrickard:
Alright, looking at the logic, it is unclear to me why this would improve the situation. At the minimum we would need tests, comments, and a clear logical explanation of why this is better before we would consider merging. Since this is such an old PR (sorry, I promise we're going to get better at this), I'm going to close. If you still want to do this, then feel free to submit another PR, and we'll consider.

PR #419, opened 2015-10-28
debugQuickLookObject categories

itsthejb:
Hi all,

Just a quick simple debugging aid for Xcode users here. Very trivial addition, but looks potentially very helpful. Up until this point I've been using Chisel for this myself, but direct Quicklook access via LLDB seems much more user friendly:

screen shot 2015-10-28 at 10 44 36

Let me know what you think.

Thanks!


adamjernst:
Neat!

PR #422, opened 2015-11-02
Add configurable watchdog timer

adamjernst:
The process of building a component may deadlock. Add a class that will detect this case and fire a configurable C function pointer when it occurs.

### DIFF
--- a/ComponentKit/Utilities/CKWatchdogTimer.mm
+++ b/ComponentKit/Utilities/CKWatchdogTimer.mm
@@ -1,0 +1,93 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKWatchdogTimer.h"
+
+#import <UIKit/UIKit.h>
+#include <mach/mach_time.h>
+
+#import "CKMutex.h"
+
+/** Protects timeoutNanoseconds and handler */
+static CK::StaticMutex mutex = CK_MUTEX_INITIALIZER;
+static uint64_t timeoutNanoseconds = 0;
+static void (*handler)(void) = nullptr;
+
+static dispatch_source_t newTimer()
+{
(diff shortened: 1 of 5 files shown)


ComponentKit/Utilities/CKWatchdogTimer.mm
```diff
@@ -0,0 +1,94 @@
...
+                                                  usingBlock:^(NSNotification * _Nonnull note) {
+                                                    lastBackgroundingTimestamp = mach_absolute_time();
+                                                  }];
+  });
+
+  const uint64_t timerCreationTimestamp = mach_absolute_time();
+  dispatch_source_t timer = dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER, 0, 0, dispatch_get_main_queue());
+  dispatch_source_set_event_handler(timer, ^{
+    CKCAssertMainThread();
+    // Don't fire the handler if the app backgrounded after scheduling the timer since that suspends app's CPU
```
adamjernst: It would be fancier to reschedule the timer upon app re-foregrounding, but that makes this much more complicated. I think I'd prefer to start simple.

ComponentKit/Utilities/CKWatchdogTimer.mm
```diff
@@ -0,0 +1,94 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKWatchdogTimer.h"
+
+#import <UIKit/UIKit.h>
+#import <vector>
```
adamjernst: Oops, can remove this.

ComponentKit/Utilities/CKWatchdogTimer.mm
eczarny: I prefer the simplicity of this approach as well.

eczarny:
This looks solid. Is this something we could reliably add test coverage for?

natansh:
Looks good to me.

adamjernst:
@eczarny I'm going to skip test coverage. I think this isn't well-suited to unit testing; it would require a lot of timers and waits, which never makes for reliable tests.

eczarny:
@adamjernst :+1:

PR #424, opened 2015-11-05
Include queue label in watchdog timer callback

adamjernst:


ComponentKit/Utilities/CKWatchdogTimer.mm
```diff
@@ -18,9 +18,15 @@
 /** Protects timeoutNanoseconds and handler */
 static CK::StaticMutex mutex = CK_MUTEX_INITIALIZER;
 static uint64_t timeoutNanoseconds = 0;
-static void (*handler)(void) = nullptr;
+static void (*handler)(NSString *) = nullptr;
 
-static dispatch_source_t newTimer()
+static NSString *currentQueueLabel(void)
+{
+  const char *label = dispatch_queue_get_label(DISPATCH_CURRENT_QUEUE_LABEL);
```
eczarny: Would this always be the preparation queue's label?

adamjernst: Yes, so I was hoping we could customize the label for each?

eczarny: :+1:

eczarny:
LGTM

Issue #428, opened 2015-11-20
Help using CKTextKitEntityAttribute


gudjao:
I'm using CKTextKitEnitityAttribute to make the link clickable. but the problem is that when using CKTextComponent, the app crashes when you click on a trancated link.

<img width="328" alt="screen shot 2015-11-20 at 12 24 47 pm" src="https://cloud.githubusercontent.com/assets/4822801/11291937/b6958c7a-8f81-11e5-965e-a55c49dec24b.png">

```objective-c
[attributedString addAttribute:CKTextKitEntityAttributeName value:[[CKTextKitEntityAttribute alloc] initWithEntity:URL] range:result.range];
```

```objective-c
                contentComponent = [CKTextComponent
                                    newWithTextAttributes:{
                                        .attributedString = [self checkURLString:content.content],
                                        .truncationAttributedString = continueReading(),
                                        .maximumNumberOfLines = 4,
                                        .lineBreakMode = NSLineBreakByWordWrapping
                                    }
                                    viewAttributes:{
                                        {@selector(setBackgroundColor:), [UIColor clearColor]},
                                        {@selector(setUserInteractionEnabled:), @YES},
                                        CKComponentActionAttribute(@selector(stringTapped:withEvent:), CKUIControlEventTextViewDidTapText)
                                    }
                                    accessibilityContext:{
                                    }];
```

ocrickard:
Can you provide the exception and call stack please?

gudjao:
@ocrickard here it is.

*** Assertion failure in __63-[CKTextKitRenderer(Tracking) rectsForTextRange:measureOption:]_block_invoke(), /.../Pods/ComponentKit/ComponentTextKit/TextKit/CKTextKitRenderer+Positioning.mm:31

*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: ''
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010c506f45 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x000000010bf7edeb objc_exception_throw + 48
	2   CoreFoundation                      0x000000010c506daa +[NSException raise:format:arguments:] + 106
	3   Foundation                          0x0000000108c9c6b2 -[NSAssertionHandler handleFailureInFunction:file:lineNumber:description:] + 169
	4   Petgago                             0x00000001078de4b6 __63-[CKTextKitRenderer(Tracking) rectsForTextRange:measureOption:]_block_invoke + 646
[...]

PR #431, opened 2015-11-25
[text] Fix crash when highlighted range exceeds truncation range

ocrickard:
This was a needless assert. Remove it and just clamp the range to the string range. Add a test to confirm this works.

Resolves https://github.com/facebook/componentkit/issues/428

### DIFF
--- a/ComponentTextKitApplicationTests/CKTextKitTests.mm
+++ b/ComponentTextKitApplicationTests/CKTextKitTests.mm
@@ -13,8 +13,10 @@
 
 #import <FBSnapshotTestCase/FBSnapshotTestController.h>
 
+#import "CKTextKitEntityAttribute.h"
 #import "CKTextKitAttributes.h"
 #import "CKTextKitRenderer.h"
+#import "CKTextKitRenderer+Positioning.h"
 
 @interface CKTextKitTests : XCTestCase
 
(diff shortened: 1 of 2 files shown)


ComponentTextKitApplicationTests/CKTextKitTests.mm
```diff
@@ -135,4 +137,22 @@ - (void)testStringsWithVariableAttributes
...
+  CKTextKitRenderer *renderer =
+  [[CKTextKitRenderer alloc]
+   initWithTextKitAttributes:{
+     .attributedString = attributedString,
+     .maximumNumberOfLines = 1,
+     .truncationAttributedString = [[NSAttributedString alloc] initWithString:@"... Continue Reading"]
+   }
+   constrainedSize:{ 100, 100 }];
+
+  XCTAssert([renderer rectsForTextRange:NSMakeRange(0, attributedString.length) measureOption:CKTextKitRendererMeasureOptionBlock].count > 0);
```
ocrickard: This test would have fail asserted before this diff.

eczarny: Should we add test cases for the updated conditional? Basically if the clamped range location is not found or its length is 0.

ocrickard: I don't really care either way. If you think it's important, sure.

ocrickard: Please request changes if you do think I need to do this.

eczarny: I don't think it's important for this change.

eczarny:
LGTM

Issue #429, opened 2015-11-24
Question - component (controller) lifecycle - "birth" and "death" hooks


leoschweizer:
I often encounter situations where I want my components to be delegates of some kind of event sources (e.g. a countdown timer, which should trigger an update of the component on every tick). So I have associated component controllers which subscribe and unsubscribe to these sources.

However, none of the provided lifecycle methods seem to suit this purpose. `willMount/didMount` and `willUnmount/didUnmount` are e.g. invoked for every `updateState:`, not when the component (controller) gets created for the first time, or respectively when it gets deallocated.

So the second obvious idea is to override `init` and `dealloc` in the component controller class, but that doesn't help either, because access to the associated component (and thus the associated model instance) is not yet possible in `init` (and is probably not guaranteed in `dealloc` either).

What is a robust way to handle situations like these?

adamjernst:
Yep! Short answer is that there is no good solution just yet.

I've been thinking about it recently for a similar use case of my own.

I'd be open to PRs of course.

leoschweizer:
@adamjernst I'd be willing to contribute one. Do you already have a rough idea how it could work?

I just skimmed the `CKComponentController` implementation and it seems that a `willMountInitially/didMountInitially` lifecycle method could be straightforward (based on `_performedInitialMount`?).

However I don't see how an end-of-life hook would work.

ocrickard:
I'm going to close this for now. There isn't a good way for us to do this today, though it's on @adamjernst's radar and we're going to try and fix soon.

PR #434, opened 2015-11-27
introduced willMountInitially and didMountInitially lifecycle methods

leoschweizer:
this contributes half of what is required for #429


adamjernst:
Thanks! Will review it.

adamjernst:
Sorry, I'm going to vote against this one actually. You can achieve almost the same thing with an ivar on your component controller.

I *would* be interested in a "death" hook for component controllers that is guaranteed to execute on the main thread.

PR #437, opened 2015-12-03
Allow primitive values in view attribute maps

darknoon:
Instead of forcing people to manually box values for setters, we can use c++ to deduce the right way to box it for them. Still backwards compatible with manual boxing.

Shouldn't cause additional overhead, but we should test just in case.


adamjernst:
Neat trick.

Woody17:
:+1: nice!

PR #450, opened 2015-12-23
[ComponentKit] Fix animation leaks

vreis:
Retaining animations on CKComponentController causes a reference cycle:
the component is retained by the animation, which is retained by the
controller, which is retained by the scope handle, which is retained by
the component.

The cycle gets broken when the component is unmounted, but that happens
too late in some cases. This change makes the cycle get broken when the
animation finishes.

### DIFF
--- a/ComponentKit/Core/CKComponentController.mm
+++ b/ComponentKit/Core/CKComponentController.mm
@@ -34,6 +34,9 @@ typedef NS_ENUM(NSUInteger, CKComponentControllerState) {
   CKComponentControllerStateUnmounting,
 };
 
+typedef size_t CKComponentAnimationID;
+typedef std::unordered_map<CKComponentAnimationID, CKAppliedComponentAnimation> CKAppliedComponentAnimationMap;
+
 static NSString *componentStateName(CKComponentControllerState state)
 {
   switch (state) {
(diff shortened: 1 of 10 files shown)


ComponentKit/Core/CKComponentController.mm
```diff
@@ -191,15 +193,31 @@ - (void)_relinquishView
...
   }
   _appliedAnimations.clear();
 }
 
+- (void)_applyAnimation:(const CKPendingComponentAnimation &)pendingAnimation toMap:(CKAppliedComponentAnimationMap &)map
+{
+  [CATransaction begin];
+  CKComponentAnimationID animationID = _nextAnimationID++;
+  [CATransaction setCompletionBlock:^() {
+    auto it = map.find(animationID);
```
adamjernst: alas, this won't work because you're capturing a C++ reference in a block. that's akin to capturing a raw pointer; if the component controller deallocates before the block is executed, you'll dereference garbage memory and crash.

I'm a bit flummoxed (as i imagine you are) by how to make this code reusable. Hrm.

vreis: Good catch! I hadn't noticed that. We can make the block capture the controller, just to make sure this will work. It's a bit obscure, though, and I'd rather have the code duplicated.
Other possibility is to turn CKAppliedComponentAnimationMap into a std::shared_ptr and pass it by copy. Not sure I like that either.

adamjernst: Any way is fine to me.

vreis:
I rebased and screwed up everything. Following up on https://github.com/facebook/componentkit/pull/451

PR #458, opened 2016-01-07
Fix flash in CKNetworkImageComponent when defaultImage is different.

mcohnen:
This is the 2nd attempt to fix this. This one is respecting cropRect and is only doing the necessary steps in setSpecifier based on what fields changed in the specifier.


adamjernst:
woohoo, will review shortly

adamjernst:
Done!

PR #464, opened 2016-01-16
Create willDealloc CKComponentController lifecycle method

dbworku:
Similar to #429, I have a case where a component's controller is managing a timer. 

This PR, introduces a new api, willDealloc, on `CKComponentController`. 

The `CKComponentScopeRoot` implementation lead me to believe that a component's controller's lifecycle ends when the scope root is release. Therefore, my approach was to trigger `willDealloc`, on the main thread,  just before the `CKComponentLifecycleManager` destroys the root scope.

TODO: Document willDealloc in API and Getting Started guide (should this PR get merged)

### DIFF
--- a/ComponentKit/Core/CKComponentController.h
+++ b/ComponentKit/Core/CKComponentController.h
@@ -39,6 +39,9 @@
 /** The controller's component was previously mounted, but now it no longer is. */
 - (void)didUnmount NS_REQUIRES_SUPER;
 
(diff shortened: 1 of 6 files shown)


ocrickard:
Thanks for the contribution! Unfortunately I don't think this is something we would want in the core framework. Notification on dealloc is a dangerous pattern that we avoid after seeing all the problems it causes.

PR #467, opened 2016-01-20
Add "remove all" support

itsthejb:
Hi all,

Here's one I should have thought to submit a while ago. Basically, CK is very good at inserting items, and removing individual items, but lacks any support for a one-shot *reset* or *clear all* feature. We actually need this to support a kind of "reset and refresh" functionality, and it requires a bit of external overhead in order to support this otherwise - we need access to a reliable set of already inserted item indexPaths and section indexes. Since `CKTransactionalComponentDataSourceState` already has all of this info internally, it's trivial to add.

I would perhaps have added this in a category, but it's necessary to add in the main implementation since `CKTransactionalComponentDataSourceState` is a direct ivar usage. Also unsure of threading considerations here. In any case, basically implemented with a simple test. Happy to hear feedback, of course.

Thanks!


adamjernst:
Super cool. Can we implement this as a helper function in a separate repo entirely which generates a changeset though, instead of building it into the transactional data source itself?

- I want the core data source to remain as simple as possible, with the bare minimum of operations there.
- I think it shouldn't be too hard to implement this as a category on the changeset class.

I absolutely agree that changesets are clunky to create and deal with directly! Both this and your previous "changeset builder" API are great ideas. I just want to keep them outside the core, since my overriding goal is to keep the core tiny and then point people to great utilities to make that core easier to use.

Shout if you disagree. I don't want to discourage contributions.

itsthejb:
Yes, no problem. I would have done this as a category already, in fact, but due to use of ivars over properties, I would have to use KVC to get to `_currentState` so that's why I went with the main implementation approach. 

I could in fact include this is something larger than I'm working on, although it would be a slightly strange fit.

itsthejb:
Now that I think about it, perhaps this would be a good addition to the change set generator... That has legs. Otherwise this feature alone as its own Pod would be a bit tiny. I'll have a look into that.

itsthejb:
Alright, I finally got around to bundling this and a few other things into [a modular pod](http://www.cocoapods.org/?q=CKToolbox). Check it out :wink:

adamjernst:
So awesome! Thanks @itsthejb!

PR #473, opened 2016-01-22
Replace the mapIndex method on CKArrayControllerSections

eczarny:
The `mapIndex` method is public and may still be used. Let's add it back for now.

I'm not quite familiar with the reasoning behind removing it in the first place. We should make sure that it isn't incompatible with the changes introduced in #414.


adamjernst:
ship it

PR #476, opened 2016-01-25
Remove unused CKComponentDataSourceAttachController helper method

eczarny:
This should have been removed in #462.


adamjernst:
@eczarny  ship at will

PR #479, opened 2016-02-05
Flex-wrap in CKStackLayoutComponent

antigp:
This is implement of flex-wrap in CKStackLayoutComponent
Work with Horizontal and Vertical Stack

unknown-1

### DIFF
--- a/ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
+++ b/ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
@@ -45,9 +45,9 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
   const auto maxCrossSize = crossDimension(style.direction, constrainedSize.max);
   const auto maxStackSize = stackDimension(style.direction, constrainedSize.max);
 
-  const CGFloat crossSize = MIN(MAX(minCrossSize, largestChildCrossSize), maxCrossSize);
+  CGFloat crossSize = MIN(MAX(minCrossSize, largestChildCrossSize), maxCrossSize);
 
-  CGFloat maxLineHieght = 0;
+  CGFloat maxCrossItemSize = 0;
   CGFloat crossPosition = 0;
   CGPoint p = directionPoint(style.direction, offset, 0);
   BOOL first = YES;
@@ -55,9 +55,18 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
     if(!first && style.flexWrap == CKStackLayoutWrapFlexWrap) {
       CGFloat stackSize = (style.direction == CKStackLayoutDirectionVertical) ? p.y : p.x;
       if (stackSize + l.child.spacingBefore + style.spacing + stackDimension(style.direction, l.layout.size) + l.child.spacingAfter > maxStackSize) {
-        crossPosition = crossPosition + maxLineHieght;
-        maxLineHieght = 0;
+        // Add maximum item size for new line cross position
+        crossPosition = crossPosition + maxCrossItemSize;
+        // Reset maximum cross size for new line
+        maxCrossItemSize = 0;
+        // Reset point to start of the line
         p = directionPoint(style.direction, offset, crossPosition);
+        // Indicate the first item of line
+        first = YES;
+      }
+      if(crossDimension(style.direction, l.layout.size) > maxCrossItemSize) {
+        // Calculating maximum cross size for current line
+        maxCrossItemSize = crossDimension(style.direction, l.layout.size);
       }
     }
     p = p + directionPoint(style.direction, l.child.spacingBefore, 0);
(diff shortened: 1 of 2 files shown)


ocrickard:
Great! Really excited about this. To start off we're going to need comprehensive test coverage to make sure this matches the standard.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
```diff
@@ -43,11 +43,23 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
   const auto largestChildCrossSize = it == unpositionedLayout.items.end() ? 0 : crossDimension(style.direction, it->layout.size);
   const auto minCrossSize = crossDimension(style.direction, constrainedSize.min);
   const auto maxCrossSize = crossDimension(style.direction, constrainedSize.max);
+  const auto maxStackSize = stackDimension(style.direction, constrainedSize.max);
+
   const CGFloat crossSize = MIN(MAX(minCrossSize, largestChildCrossSize), maxCrossSize);
 
+  CGFloat maxLineHieght = 0;
```
ocrickard: *maxLineHeight

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
```diff
@@ -59,9 +71,12 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
       l.layout,
     };
     p = p + directionPoint(style.direction, stackDimension(style.direction, l.layout.size) + l.child.spacingAfter, 0);
+    if(crossDimension(style.direction, l.layout.size) > maxLineHieght) {
```
ocrickard: This is computed even if flexWrap isn't enabled.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
ocrickard: Also, line height I don't think is a very good name. This algorithm is built to work in either horizontal or vertical directions, so I doubt this is a height in every case right? Height implies vertical directionality of the value.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
```diff
@@ -43,11 +43,23 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
...
+      CGFloat stackSize = (style.direction == CKStackLayoutDirectionVertical) ? p.y : p.x;
+      if (stackSize + l.child.spacingBefore + style.spacing + stackDimension(style.direction, l.layout.size) + l.child.spacingAfter > maxStackSize) {
+        crossPosition = crossPosition + maxLineHieght;
+        maxLineHieght = 0;
+        p = directionPoint(style.direction, offset, crossPosition);
+      }
+    }
     p = p + directionPoint(style.direction, l.child.spacingBefore, 0);
     if (!first) {
       p = p + directionPoint(style.direction, style.spacing, 0);
```
ocrickard: What happens here with your algorithm? When we proceed to break the line are we still adding spacing before the element? Probably need a test for this.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
```diff
@@ -43,11 +43,23 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
...
+  CGFloat maxLineHieght = 0;
+  CGFloat crossPosition = 0;
   CGPoint p = directionPoint(style.direction, offset, 0);
   BOOL first = YES;
   auto stackedChildren = CK::map(unpositionedLayout.items, [&](const CKStackUnpositionedItem &l) -> CKComponentLayoutChild {
+    if(!first && style.flexWrap == CKStackLayoutWrapFlexWrap) {
+      CGFloat stackSize = (style.direction == CKStackLayoutDirectionVertical) ? p.y : p.x;
+      if (stackSize + l.child.spacingBefore + style.spacing + stackDimension(style.direction, l.layout.size) + l.child.spacingAfter > maxStackSize) {
+        crossPosition = crossPosition + maxLineHieght;
+        maxLineHieght = 0;
```
ocrickard: You probably need a comment here explaining why you're setting to 0. It takes a bit of reading to realize you're resetting the max line height for the next line.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
```diff
@@ -59,9 +71,12 @@ static CKStackPositionedLayout stackedLayout(const CKStackLayoutComponentStyle &
       l.layout,
     };
     p = p + directionPoint(style.direction, stackDimension(style.direction, l.layout.size) + l.child.spacingAfter, 0);
+    if(crossDimension(style.direction, l.layout.size) > maxLineHieght) {
+        maxLineHieght = crossDimension(style.direction, l.layout.size);
+    }
     return c;
   });
-  return {stackedChildren, crossSize};
+  return {stackedChildren, style.flexWrap == CKStackLayoutWrapFlexWrap ? (maxLineHieght + crossPosition) : crossSize};
```
ocrickard: You are dealing with a positioned layout here, but the sizes of the items has already been determined by the unpositioned layout algorithm. Don't you need to update that algorithm as well? Maybe I'm missing something though... it's been a long time since I worked on this code.

ComponentKit/LayoutComponents/CKStackPositionedLayout.mm
antigp: SpacingBefore always add even in the first item of line. 
style.spacing will not be added in new line. I added a flag setting "first" to YES

ocrickard: Ahh, I see you're setting first above. Got it.

ocrickard:
As I mentioned, I believe this is an incomplete implementation:

https://www.w3.org/TR/css-flexbox-1/#multi_line

At the point you're injecting this code in the algorithm the sizes have been determined according to a single-line layout. The standard specifies that each line should be laid out independently according to the flex-box rules.

What we actually need here is a multi-pass sizing/positioning algorithm. In the unpositioned layout phase we need to determine how many lines we are going to break into, and size each of the lines using the normal algorithm. This will probably take two passes (one sizing pass to line break, one to size independently once we know how many elements will fit on each line).

ocrickard:
I think it'd be awesome to approach this from a TDD perspective. Write a whole bunch of snapshot tests where we would hope to get all the edge-cases covered. Compare the output behavior with the behavior of Chrome to confirm the expected behavior, then tweak the algorithm until we get the expected results.

ocrickard:
I'm going to close the PR for now. Please re-issue a PR after addressing feedback.

wy19901227:
@antigp why not to write a new component named CKFlowLayoutComponent?

PR #485, opened 2016-02-17
Fixing UITableview's demo code

vizlabxt:
It seem like the CKArrayControllerOutputChange API has changed. The [UITableView's demo](http://componentkit.org/docs/datasource-dive-deeper.html) doesn't compile.


eczarny:
Good catch @vizlabxt!

PR #489, opened 2016-02-20
Changed iphone simulator number in the build script to 9.2

MDSilber:
This is so CI will compile code with generics

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ before_install:
   - gem install cocoapods -v 0.37.2
 
 script: ./build.sh ci
+osx_image: xcode7.3
 
 # Whitelist for branches
 branches:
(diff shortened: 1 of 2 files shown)


adamjernst:
Nice. Can yo figure out how to make Travis work?

natansh:
@MDSilber Have a look at https://docs.travis-ci.com/user/languages/objective-c/

Based on this, I think you'll have to specify both `iphonesimulator9.3` (I'd suggest going straight to 9.3) as well as `osx_image: xcode7.3`  in the `.travis.yml` file to allow it to work.

natansh:
It's important to note that adding generics might break compatibility with previous SDK versions. I don't see that as a blocker, but something to be mindful of.

MDSilber:
Weird git things, going to submit a new pull request

PR #503, opened 2016-03-15
Add the designated initializer style best practice

eczarny:


adamjernst:
@eczarny One unusual exception is that I usually support omitting `.component = ` for `CKStackLayoutComponent` in particular, since there aren't two distinct fields that accept `CKComponent`s and it really cuts down on clutter.

I don't think it is necessarily worth documenting this though, since I'm fine with people using `.component = ` if they really want to.

PR #496, opened 2016-02-26
Subspec test helper libs

dbworku:
Split from #495 to only contain migration subspec changes.

- Migrate ComponentKitTestLib and ComponentSnapshotTestCase podspec to
subspecs on the ComponentKit.podspec. The public now has access -- pod
'ComponentKit/ComponentSnapshotTestCase' and your done.
- Migrate ComponentKitTestHelpers static lib to subspec.
- Migrate podspec to pod_target_xcconfig and os.deployment_target.
(Support introduce 0.38.2)

Motivation: #483

PR #506, opened 2016-03-21
Use Subspec for static libs

dbworku:
#496 builds were failing. Repeated tested build script locally for this PR. The Travis build should succeed.

- Migrate ComponentKitTestLib and ComponentSnapshotTestCase podspec to
subspecs on the ComponentKit.podspec.
- Migrate ComponentKitTestHelpers static lib to subspec.
- Migrate podspec to pod_target_xcconfig and os.deployment_target.
(Support introduce 0.38.2)
- Use Cocoapods v0.38.2 with Travis CI builds


ComponentSnapshotTestCase/CKComponentSnapshotTestCase.mm
```diff
@@ -8,7 +8,7 @@
  *
  */
 
-#import <ComponentSnapshotTestCase/CKComponentSnapshotTestCase.h>
+#import <ComponentKit/CKComponentSnapshotTestCase.h>
```
adamjernst: Hm, this won't work with our internal build process (which isn't cocoapods based).

How is it that `CKComponentSnapshotTestCase` lives in ComponentKit proper, as opposed to a separate sub-library? We specifically want to avoid linking it into production builds.

Can you explain this a little more? I don't know a lot about Cocoapods.

adamjernst:
See note on commit @dbworku — I'm enthusiastic about this change, but there is a problem with our internal build process, would love to brainstorm solutions.

ComponentSnapshotTestCase/CKComponentSnapshotTestCase.mm
dbworku: Think of [subspec](https://guides.cocoapods.org/syntax/podspec.html#subspec) as an optional expansion library upon the parent spec. A subspec inherits the value of the attributes of the parent. 

You can add a specific subspec to a podfile for their project with the form:
`pod 'PARENT/SUBSPEC'`

Otherwise, Cocoapods uses the `default_subspec` defined in the Podspec. For example
`pod 'PARENT'` == `pod 'PARENT/default_subspec'`

The `default_spec`, of ComponentKit, is "Core". I defined Core to simply link the .h, .m and .mm files, recursively contained, in `./ComponentKit` and `./ComponentTextKit`. 

The podfile for WildeGuess imports the plain old:
`pod 'ComponentKit', :path => '.'`
In this case, cocoapods uses the default subspec of `ComponentKit/Core`. Therefore, the files from the other subspecs are not linked it into builds (production or debug).

Podfile
```diff
@@ -3,19 +3,19 @@ source 'https://github.com/CocoaPods/Specs.git'
...
+target 'ComponentKitTests' do
+	pod 'OCMock', '~> 3.2'
+	pod 'ComponentKit/ComponentKitTestHelpers', :path => '.'
+	pod 'ComponentKit/ComponentKitTestLib', :path => '.'
+end
 
+target 'ComponentKitApplicationTests' do
+	pod 'ComponentKit/ComponentKitTestHelpers', :path => '.'
+	pod 'ComponentKit/ComponentSnapshotTestCase', :path => '.'
+end
```
dbworku: Here's an example of subspec usage... The 'ComponentKitApplicationTests' target needs access to the ComponentSnapshotTestCase and ComponentKitTestHelpers libraries. Cocoapods will only link those libraries to the  'ComponentKitApplicationTests' target.

dbworku:
@adamjernst Yes, I'm sure that we can up with a solution! 

It looks like [FBSnapshotTestCase uses subspec](https://github.com/facebook/ios-snapshot-test-case/blob/master/FBSnapshotTestCase.podspec) without build issues. 

Can you provide more information surrounding the specific stage in the build process that these changes fail? That will help narrow our options for solutions.

adamjernst:
@dbworku We use buck for builds and buck expects each library (aka subspec) to have a distinct header path prefix (the part before / in #import statements).

FBSnapshotTestCase is a different matter since we simply exclude SwiftSupport entirely; we don't use it yet...

I'm not sure what to propose here. Is there a way to tell Cocoapods to use a different header prefix for a subspec? Are there subspec docs I can read through?

adamjernst:
It sounds like subspecs are mostly used to *add* features to a base spec, while in our opinion the snapshot test case is a logically distinct library. Does that make sense?

dbworku:
@adamjernst Thanks for clarifying with more detail! 

> Is there a way to tell Cocoapods to use a different header prefix for a subspec?

I couldn't find a way to instruct cocoapods to use a custom header of a subspec.

> Are there subspec docs I can read through?

[This is it](https://guides.cocoapods.org/syntax/podspec.html#subspec) for documentation. :disappointed: 
 
> It sounds like subspecs are mostly used to add features to a base spec, while in our opinion the snapshot test case is a logically distinct library. Does that make sense?

Yes, I didn't share that view in my approach. It is now much more clear why ComponentSnapshotTestCase originally had it's own Podspec. 

We can bring back multiple pod specs, to preserve the import path prefix `ComponentSnapshotTestCase/*`. You need to push the other pods in into the CocoaPods Trunk (`pod trunk push ComponentSnapshotTestCase.podspec`), to give public access to ComponentSnapshotTestCase. The solution should preserve Buck builds and allow devs to make their own snapshot tests with `pod 'ComponentSnapshotTestCase'`


What is your view on ComponentKitTestHelpers and ComponentKitTestLib?

adamjernst:
I *think* we definitely want ComponentKitTestHelpers and ComponentKitTestLib to remain as logically distinct libraries — they're designed to be reused by sublibraries you create in your app that may need helpers to test components.

dbworku:
@adamjernst Okay, thanks for clarifying the approach. The podspecs, for those libraries, also need to be pushed into the CocoaPods Trunk, to allow the community to use them in their apps.

No changes are required to the source files.

PR #529, opened 2016-04-06
Changes definition of equality for CKTransactionalComponentDataSource…

MDSilber:
…State so only the models are checked in CKTransactionalComponentDataSourceItems (instead of the layouts, scope roots, components inside the layouts, etc)

### DIFF
--- a/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceState.mm
+++ b/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceState.mm
@@ -13,6 +13,7 @@
 #import "CKEqualityHashHelpers.h"
 #import "CKMacros.h"
 #import "CKTransactionalComponentDataSourceConfiguration.h"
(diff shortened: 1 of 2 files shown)


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceState.mm
```diff
@@ -88,4 +89,15 @@ - (NSUInteger)hash
...
+{
+  NSMutableArray *models = [NSMutableArray new];
+  for (NSArray *section in sections) {
+    for (CKTransactionalComponentDataSourceItem *item in section) {
+      [models addObject:item.model];
+    }
+  }
+  return models;
+}
+
```
natansh: If you have a case of

Section 1: Item 1
Section 2: Item 2, Item 3

and 

Section 1: Item 1, Item 2
Section 2: Item 3

Then this would consider them equal. Are you sure you want that behavior?

PR #533, opened 2016-04-12
Adds equality method for TDS item and tests

MDSilber:


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceItem.mm
```diff
@@ -12,6 +12,8 @@
 #import "CKTransactionalComponentDataSourceItemInternal.h"
 
 #import "CKComponentLayout.h"
+#import "CKEqualityHashHelpers.h"
```
adamjernst: unused?

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceItem.mm
```diff
@@ -12,6 +12,8 @@
...
+#import "CKMacros.h"
```
adamjernst: what here is used?

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceItem.mm
```diff
@@ -37,4 +39,18 @@ - (instancetype)initWithLayout:(const CKComponentLayout &)layout
   return _layout;
 }
 
+- (BOOL)isEqual:(id)object
+{
+  if (![object isKindOfClass:[CKTransactionalComponentDataSourceItem class]]) {
+    return NO;
+  } else {
+    return [_model isEqual:((CKTransactionalComponentDataSourceItem *)object).model];
```
adamjernst: making this based on model equality is a bit odd to me.

adamjernst:
Is there a good reason why we couldn't simply create a helper that does this externally (e.g. using a `NSComparator`) and leave the core untouched?

MDSilber:
@adamjernst this is used by OCMock to compare object equality, so unfortunately we need to override isEqual:

adamjernst:
@MDSilber  so don't use AssertEqualObjects. Use your own comparator and assert the result is YES.

PR #534, opened 2016-04-12
Switch from CocoaPods to Carthage for development

conradev:
This PR adds support for Carthage and makes building the project possible without any external build tool. Carthage is only required to build the testing dependencies.

The framework target in the `xcodeproj` is a superset of the `podspec` in that it annotates each header with Public/Private/Internal. Some headers that are named Internal are currently marked as Private because the tests require them. The `xcodeproj` does *not* contain any build settings – those are all in an `xcconfig` file.

The only code change is the addition of `NS_EXTENSION_UNAVAILABLE` notation to the `CKComponentDebugController`.

The tests run fine locally, and the CI *should* also work (but `build.sh` might need further modification).


modocache:
> makes building the project possible without any external build tool

This is fantastic. I'm curious, what would it take to be able to build and test the project out of the box? I know many new contributors encounter the most difficulty with just getting everything set up, especially since they don't use CocoaPods and Carthage very often at Facebook.

conradev:
To build the project, you need to do clone the project with submodules, open the project in Xcode and press `⌘+B`.

To run the tests, you need to run the following command once:
```
carthage build --platform iOS
```
and then all you need to do is open the project in Xcode and press `⌘+U`. You also need the iPhone Simulator 8.x installed because the snapshot tests have not been updated for iOS 9.

adamjernst:
Pretty sweet, but as someone who is blissfully unaware of the Carthage vs Cocoapods debate, is there a reason we can't support both? Removing the podfile entirely for people already using it via Cocoapods makes me nervous.

conradev:
The reason I removed it was because I didn't see a clear way to keep both in the same Xcode project (it's not just the `Podfile`) :(

The changes I made to the Xcode project:
– Added the ComponentKit framework target
– Deleted the existing scheme, created one for the framework, and attached the tests to that scheme
– Added `ComponentSnapshotTestCase` and `ComponentKitTestLib` into the test targets
– Added the snapshots into the test targets
– Added the testing dependencies as file references
– Removed the `libPods.a` reference

I *could* make it so that there are two separate schemes, one for the CocoaPods target, and one for the framework target if that's an approach that sounds reasonable!

conradev:
I went ahead and updated the PR – now it doesn't even require `carthage` to run the tests, @modocache! I realized I could just use subprojects in Xcode for the testing dependencies.

I also fixed the build script such that all of the tests run successfully on Travis.

modocache:
Looks fantastic! One last thing--we've had so much pain onboarding people onto ComponentKit in the past, since they didn't know how to use CocoaPods (or Carthage, for that matter). Could you include some instructions somewhere (maybe in the README.md or CONTRIBUTING.md) that explain why and how one would update the Carthage checkouts?

I'll be the first to admit that I don't know. I guess if I need a newer version of one of the dependencies, I should update the `Cartfile.private`, then install `carthage`, then run `carthage update` or something?

conradev:
I added a note on *how* to update the testing dependencies to `CONTRIBUTING.md`, but I'm not sure about any of the reasons why someone would want to do that. The main reason I can imagine is a warning being introduced in a new version of Xcode that an updated version of the dependency addresses.

modocache:
Awesome. This looks excellent to me, I'll defer to @adamjernst to merge.

eczarny:
If we merged this pull request would that mean we're abandoning folks who are using ComponentKit via CocoaPods?

conradev:
No, we would only be abandoning developers who use CocoaPods to *test* ComponentKit. The `podspec` would continue to work as it does now.

adamjernst:
Cool, let me review this soon.

eczarny:
@conradev would you mind updating the quick start section in the README as well?

conradev:
I went and added Carthage to the `README`, let me know if there are any other changes I should make!

eczarny:
Thanks @conradev! I'll make sure to spend some time reviewing/trying this out tomorrow.

eczarny:
This change is great. It was wonderful just being able to open the Xcode project and get started.

Overall the changes here seem fine to me. @adamjernst mind taking a quick look to make sure everything works for you? Is there a way to verify the CocoaPod release workflow still works?

adamjernst:
This deletes the `podspec` file for `ComponentSnapshotTestCase` and `ComponentKitTestLib`, but we never published those in the first place. Oops! So, let's do this.

adamjernst:
@conradev Thanks for pushing this over the line. Much appreciated!

PR #535, opened 2016-04-15
Expose a way to get state updater from CKComponentScope

adamjernst:
This is useful when you want to pass a state updater off to some element that is a sub-element of your component (and thus you can't wait until the component is created to pass it off).

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScope.h
+++ b/ComponentKit/Core/Scope/CKComponentScope.h
@@ -10,7 +10,12 @@
 
 #import <Foundation/Foundation.h>
 
+#import <ComponentKit/CKUpdateMode.h>
+
 class CKThreadLocalComponentScope;
+@class CKComponentScopeHandle;
+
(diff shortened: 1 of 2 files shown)


ComponentKit/Core/Scope/CKComponentScope.h
```diff
@@ -49,11 +54,14 @@ class CKComponentScope {
   ~CKComponentScope();
 
   /** @return The current state for the component being built. */
-  id state() const;
+  id state(void) const;
+
+  /** @return A block that schedules a state update. Usually, use [CKComponent -updateState:mode:] instead. */
```
eczarny: It would be nice to explain why this exists. I'd also recommend explaining why using `-[CKComponent updateState:mode:]` is preferable.

eczarny:
👍

ComponentKit/Core/Scope/CKComponentScope.h
adamjernst: Cool, i'll do in a follow on

PR #537, opened 2016-04-19
Remove the removeTransactionContainer: API

adamjernst:
It's actually no longer necessary after I removed flushing in an earlier change.


eczarny:
👍

PR #539, opened 2016-04-27
Use xcodebuild instead of xctool for CI

eczarny:
`xctool` has been reported to run into trouble with Xcode 7.3. For some context read through some of the more recent comments in https://github.com/travis-ci/travis-ci/issues/3986.

Let's just stick with `xcodebuild` instead.


modocache:
Yes please! 👏  Changes look good to me.

PR #543, opened 2016-05-03
Removed Carthage/checkouts and added it to the gitignore

MDSilber:


modocache:
I believe these *should* be in the source control, no? Aren't they git submodules that keep track of which version we're using? I think by not checking them into source control, each person who clones the repository would end up using different versions of the dependencies.

For example, Carthage itself keeps its dependencies checked in: https://github.com/Carthage/Carthage/tree/master/Carthage/Checkouts

eczarny:
@modocache I think `Cartfile.resolved` tracks this information for us (i.e. the specific version of a dependency to use). The Carthage documentation mentions that `Carthage/Checkouts` aren't required to be committed:

> You are not required to commit this folder to your repository, but you may wish to, if you want to guarantee that the source checkouts of each dependency will always be accessible at a later date.

https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#carthagecheckouts

The only real downside here is that `carthage bootstrap` could fail if GitHub, or the dependency's underlying repository, becomes unavailable.

modocache:
Ha! Glad to hear it wasn't just me that was having trouble loading GitHub.

> You are not required to commit this folder to your repository, but you may wish to, if you want to guarantee that the source checkouts of each dependency will always be accessible at a later date.

This seems like a good thing to guarantee. What's the upside of merging this change?

eczarny:
The upside is less complicated diffs when synchronizing the upstream repository with the internal mirror.

PR #545, opened 2016-05-04
Treat warnings as errors

eczarny:
Only a few places were using unused variables or functions in release builds. We can suppress warnings for those cases and change the project to treat warnings as errors.

The easiest way to test this for release builds is to use Carthage:

```
carthage build --no-skip-current --platform iOS
```


ComponentKit/Utilities/CKEqualityHashHelpers.h
```diff
@@ -102,7 +101,7 @@ namespace CKTupleOperations
       size_t prev = _hash_helper<Tuple, Index-1>::hash(tuple);
       using TypeForIndex = typename std::tuple_element<Index,Tuple>::type;
       size_t thisHash = CK::hash<TypeForIndex>()(std::get<Index>(tuple));
-      return CKHashCombine(prev, thisHash);
+      return CKHash64ToNative(CKHashCombine(prev, thisHash));
```
eczarny: `CKHashCombine` returns a `uint64_t` hash yet we're expected to return a `size_t` hash. This isn't particular safe for 32-bit devices so I try and downscale here if necessary.

PR #549, opened 2016-05-16
Changes CKComponentDeciding to have only a class method that returns …

MDSilber:
…a boolean

This also removes CKComponentConstantDecider and adds CKComponentConstantApprovingDecider and CKComponentConstantDenyingDecider as a replacement so that we don't need to keep the '_enabled' state like we did when it was an instance.


MDSilber:
Latest push fixes broken test

MDSilber:
Oh whoops, forgot to actually include the fix the first time...too much mercurial :)

MDSilber:
Missed an exclamation mark

PR #553, opened 2016-05-19
Maximum number of lines is not correctly enforced

conradev:
I added a failing test to illustrate the issue. Will update the PR with a fix for the issue if I come up with one!


adamjernst:
Good catch! Let me know if you come up with a solution.

ocrickard:
Awesome, thanks for finding this!

conradev:
Fixed by #566, should this test case (which I assume passes now) be merged in?

adamjernst:
@conradev looks like @nscoding included your test in #566, so it's merged. Credit goes to you for finding this. Thank you!

PR #556, opened 2016-05-24
[CK] Add async render mode to CKTextComponent

Msegan:
This should be exposed to give developers the ability to control async rendering more finely

### DIFF
--- a/ComponentTextKit/CKTextComponent.h
+++ b/ComponentTextKit/CKTextComponent.h
@@ -10,6 +10,7 @@
 
 #import <ComponentKit/CKComponent.h>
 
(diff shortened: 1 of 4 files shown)


ComponentTextKit/CKTextComponent.h
```diff
@@ -23,11 +24,17 @@ struct CKTextComponentAccessibilityContext
   CKComponentAccessibilityTextAttribute accessibilityLabel;
 };
 
+struct CKTextComponentOptions
+{
+  CKAsyncLayerDisplayMode displayMode;
```
ocrickard: Docs would be awesome.

PR #563, opened 2016-05-27
[CK] Add async render mode to CKTextComponent

Msegan:
This should be exposed to give developers the ability to control async rendering more finely

For some reason updating https://github.com/facebook/componentkit/pull/556 is giving me trouble so making a new one


ocrickard:
Nice, thanks.

PR #566, opened 2016-06-02
Fix maximumNumberOfLines on CKTextKitRenderer

nscoding:
This is an attempt to address the issue that @conradev illustrated with #553.

The issue is that our loop is looking at `[layoutManager numberOfGlyphs]` which counts all glyphs and not just what is visible. I am adding a check to ensure that `lineFragmentRectForGlyphAtIndex` returns non empty rect before increasing the count. This was modeled as the documentation [describes](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/TextLayout/Tasks/CountLines.html) and it worked as expected when `maximumNumberOfLines` was not set.

Any suggestions would be appreciated.

### DIFF
--- a/ComponentTextKit/TextKit/CKTextKitRenderer.mm
+++ b/ComponentTextKit/TextKit/CKTextKitRenderer.mm
@@ -125,10 +125,16 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
 {
   __block NSUInteger lineCount = 0;
   [_context performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
-    for (NSRange lineRange = { 0, 0 }; NSMaxRange(lineRange) < [layoutManager numberOfGlyphs]; lineCount++) {
-      [layoutManager lineFragmentRectForGlyphAtIndex:NSMaxRange(lineRange) effectiveRange:&lineRange];
+    NSUInteger numberOfGlyphs = [layoutManager numberOfGlyphs];
+    for (NSRange lineRange = { 0, 0 }; NSMaxRange(lineRange) < numberOfGlyphs;) {
+      CGRect lineRect = [layoutManager lineFragmentRectForGlyphAtIndex:NSMaxRange(lineRange) effectiveRange:&lineRange];
+      if (CGRectIsEmpty(lineRect)) {
+        break;
+      }
+      lineCount++;
     }
   }];
+
   return lineCount;
 }
 
--- a/ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
+++ b/ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
@@ -140,4 +140,30 @@
                                                    constrainedSize:constrainedSize]);
 }
 
+- (void)testNoLimitOfMaximumNumberOfLines
+{
+  const NSUInteger maximumNumberOfLines = 0;
+  NSAttributedString *attributedString = [self _sentenceAttributedString];
+  CKTextKitRenderer *renderer = [[CKTextKitRenderer alloc]
+                                 initWithTextKitAttributes:{
+                                   .attributedString = attributedString,
+                                   .maximumNumberOfLines = maximumNumberOfLines,
+                                 }
+                                 constrainedSize:CGSizeMake(50, INFINITY)];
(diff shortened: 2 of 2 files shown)


ComponentTextKit/TextKit/CKTextKitRenderer.mm
```diff
@@ -125,10 +125,16 @@ - (NSUInteger)lineCount
 {
   __block NSUInteger lineCount = 0;
   [_context performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
-    for (NSRange lineRange = { 0, 0 }; NSMaxRange(lineRange) < [layoutManager numberOfGlyphs]; lineCount++) {
-      [layoutManager lineFragmentRectForGlyphAtIndex:NSMaxRange(lineRange) effectiveRange:&lineRange];
+    NSUInteger numberOfGlyphs = [layoutManager numberOfGlyphs];
+    for (NSRange lineRange = { 0, 0 }; NSMaxRange(lineRange) < numberOfGlyphs;) {
+      CGRect lineRect = [layoutManager lineFragmentRectForGlyphAtIndex:NSMaxRange(lineRange) effectiveRange:&lineRange];
+      if (CGRectIsEmpty(lineRect)) {
```
ocrickard: My memory of these methods is really fuzzy... what does lineFragmentRectForGlyphAtIndex: return for a non-visible glyph?

ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
```diff
@@ -140,4 +140,30 @@ - (void)testHandleZeroHeightConstrainedSize
...
+  const NSUInteger maximumNumberOfLines = 3;
+  NSAttributedString *attributedString = [self _sentenceAttributedString];
+  CKTextKitRenderer *renderer = [[CKTextKitRenderer alloc]
+                                 initWithTextKitAttributes:{
+                                   .attributedString = attributedString,
+                                   .maximumNumberOfLines = maximumNumberOfLines,
+                                 }
+                                 constrainedSize:CGSizeMake(50, INFINITY)];
+  XCTAssertEqual(renderer.lineCount, maximumNumberOfLines);
+}
```
ocrickard: Interesting, so this fixes the renderer reporting the wrong number of lines. How about rendering? Was our numberOfLines support there broken too?

ocrickard:
Code makes sense, depending on the output of the lineFragment... method. Thanks for fixing this @nscoding!

ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
conradev: If I recall correctly, it rendered two lines

ComponentTextKit/TextKit/CKTextKitRenderer.mm
nscoding: It appears to be `CGRectZero`

ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
nscoding: `lineCount` is just a helper function that is not used when rendering, rendering works because we set `maximumNumberOfLines` on `NSTextContainer`.

ComponentTextKit/TextKit/CKTextKitRenderer.mm
nscoding: At least that is how tests pass, documentation doesn't specify but `CGRectIsEmpty` handles a null rectangle or a valid rectangle with zero height or width.

ocrickard:
Code looks good to me. Thanks for clarifying.

ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
ocrickard: Yeah, that makes sense. NSTextContainer is in charge of actually handling the rendering of the maximum number of lines. If there's a bug in rendering, let's handle that in a separate issue.

ocrickard:
@nscoding feel free to merge whenever you like.

ComponentTextKitApplicationTests/CKTextKitTruncationTests.mm
nscoding: 👍

ocrickard:
lol, perhaps I have to merge then.

PR #567, opened 2016-06-04
Fix stateful view reuse pool bug with superview changes

adamjernst:
Credit to @RomanoffAlex for finding this and proposing a fix. His description of the problem follows:

If we have preferredSuperviewMap like the following:

```
superview: [view].
```

and allViews:

```
[view].
```

Imagine that we call `viewWithPreferredSuperview(UIView *preferredSuperview)` with some `preferredSuperview` that is not in the map and a corresponding stateful view (which was picked to be recycled) doesn't have a superview.

In this case we will remove the `preferredSuperview` from `allViews` but keep it in `preferredSuperviewMap` which is really bad because if we later call `viewWithPreferredSuperview` with a superview that is a key in the map, we will reuse the view that already is used by some component that means we are stealing a stateful view from another superview that is on screen.

To fix this, get rid of the complexity of `preferredSuperviewMap` entirely. The map does not grow so large that its algorithmic efficiency is needed. Replace it with a simple linear search.

Introduce a new test for this behavior.

### DIFF
--- a/ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
@@ -16,46 +16,34 @@
 class FBStatefulReusePoolItem {
 public:
   FBStatefulReusePoolItem()
-  : preferredSuperviewMap([NSMapTable weakToStrongObjectsMapTable]), allViews([NSMutableArray array]) {};
+  : views([NSMutableArray array]) {};
 
   UIView *viewWithPreferredSuperview(UIView *preferredSuperview)
   {
-    NSMutableArray *matchingViews = [preferredSuperviewMap objectForKey:preferredSuperview];
-    UIView *view = [matchingViews lastObject] ?: [allViews lastObject];
-    if (view) {
-      if ([matchingViews count]) {
-        [matchingViews removeObject:view];
-      } else {
-        [[preferredSuperviewMap objectForKey:[view superview]] removeObject:view];
-      }
-      [allViews removeObject:view];
+    if ([views count] == 0) {
+      return nil;
     }
+    const NSUInteger matchingIndex = [views indexOfObjectPassingTest:^BOOL(UIView *view, NSUInteger idx, BOOL *stop) {
+      return [view superview] == preferredSuperview;
+    }];
+    const NSUInteger viewIndex = (matchingIndex == NSNotFound) ? 0 : matchingIndex;
+    UIView *view = views[viewIndex];
+    [views removeObjectAtIndex:viewIndex];
     return view;
   };
   
   NSUInteger viewCount()
   {
-    return [allViews count];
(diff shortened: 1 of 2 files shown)


ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -16,46 +16,34 @@
...
-        [[preferredSuperviewMap objectForKey:[view superview]] removeObject:view];
-      }
-      [allViews removeObject:view];
+    if ([views count] == 0) {
+      return nil;
     }
+    const NSUInteger matchingIndex = [views indexOfObjectPassingTest:^BOOL(UIView *view, NSUInteger idx, BOOL *stop) {
+      return [view superview] == preferredSuperview;
+    }];
+    const NSUInteger viewIndex = (matchingIndex == NSNotFound) ? 0 : matchingIndex;
```
eczarny: In the previous implementation it looks like view selection would fall back onto the last view in `allViews`. Do we want to fall back onto the first view in `views` here?

adamjernst: @eczarny  Yes, I think last vs first shouldn't matter.

RomanoffAlex: + we can have a local variable do something like that

```
const NSUInteger viewWithoutSuperviewIndex = NSNotFound;
const NSUInteger matchingIndex = [views indexOfObjectPassingTest:^BOOL(UIView *view, NSUInteger idx, BOOL *stop) {
if ([view superview] == nil ) {
viewWithoutSuperviewIndex =  idx;
}
     return [view superview] == preferredSuperview;
+    }];
```

And use matchingIndex if found below

eczarny: If it doesn't matter I'm fine with the change. I just wasn't sure if it was intentional.

ComponentKitTests/StatefulViews/CKStatefulViewReusePoolTests.mm
```diff
@@ -131,6 +131,34 @@ - (void)testEnqueueingViewThenDequeueingWithDifferentContextReturnsNewView
...
+  CKStatefulViewReusePool *pool = [[CKStatefulViewReusePool alloc] init];
+
+  UIView *container1 = [[UIView alloc] init];
+  CKTestStatefulView *view = [[CKTestStatefulView alloc] init];
+  [container1 addSubview:view];
+  [pool enqueueStatefulView:view
+         forControllerClass:[CKTestStatefulViewComponentController class]
+                    context:nil];
+
+  // remove the statefull view from the container
```
eczarny: Nit: Stateful?

eczarny:
Looks good to me! The logic is much easier to follow now.

PR #572, opened 2016-06-08
Adds CK::intersperse helper method

MDSilber:


adamjernst:
Make this take an "itemFactory" instead of a concrete object. it's not valid to have the same component in the hierarchy twice.

ComponentKit/Core/ComponentUtilities.h
```diff
@@ -75,6 +75,19 @@ namespace CK {
...
   }
+
+  template <typename T, typename Func>
+  auto intersperse(std::vector<T> a, Func &&func) -> std::vector<T>
+  {
+    std::vector<T> newVector;
+
+    for (T i : a) {
+      newVector.push_back(i);
+      newVector.push_back(func());
```
adamjernst: shouldn't intersperse *after* the last element

ComponentKit/Core/ComponentUtilities.h
```diff
@@ -75,6 +75,19 @@ namespace CK {
 
     return newVector;
   }
+
+  template <typename T, typename Func>
+  auto intersperse(std::vector<T> a, Func &&func) -> std::vector<T>
```
adamjernst: Make this take a "predicate function" that determines if elements from a should actually be added to the result.

this would be used for instance to exclude nil CKStackLayoutItems from the result vector.

MDSilber:
ping @adamjernst

ComponentKit/Core/ComponentUtilities.h
```diff
@@ -75,6 +75,31 @@ namespace CK {
...
+  template <typename T, typename Func, typename Predicate>
+  auto intersperse(std::vector<T> a, Func &&factory, Predicate predicate) -> std::vector<T>
+  {
+    std::vector<T> newVector;
+
+    if (a.size() < 2) {
+      return (a.size() == 1 && predicate(a.at(0))) ? a : newVector;
+    }
+
+    for (int i = 0; i < a.size() - 1; i++) {
```
eczarny: I imagine there's a cleaner way to iterate over the items in the vector?

ComponentKit/Core/ComponentUtilities.h
```diff
@@ -75,6 +75,31 @@ namespace CK {
 
     return newVector;
   }
+
+  template <typename T, typename Func, typename Predicate>
+  auto intersperse(std::vector<T> a, Func &&factory, Predicate predicate) -> std::vector<T>
```
eczarny: Also, it isn't entirely obvious what this is supposed to do. A brief comment perhaps?

ComponentKit/Core/ComponentUtilities.h
MDSilber: @eczarny I could do some fast enumeration but I'm going to a.size() - 1 so i figured indices was just easiest

PR #588, opened 2016-07-13
Move component scope collision detection to component layout

eczarny:
Component scope collision detection originally took place at component scope creation, during component preparation. This behavior wasn't entirely correct because components created during component preparation may not actually make their way to component layout.

Instead ComponentKit should attempt to detect component scope collisions during component layout. Not only is this more correct behavior but it also allows us to provide (not yet implemented) a more actionable error message.

To support this new behavior a couple of other notable changes were required:

1. Propagate the global identifiers of component scope handles with the same component class and identifier pair
2. Add a new `CKComputeTopLevelComponentLayout` function responsible for computing component layouts for top-level components; this avoids a problem where component scope collision detection would only be applied to portions of the overall component layout

I plan to improve the error message reported during component scope collisions in a follow up. For now the behavior is largely the same except for the omission of the scope identifier (it isn't available during component layout) in the component scope collision error message.

### DIFF
--- a/ComponentKit/Core/CKComponentInternal.h
+++ b/ComponentKit/Core/CKComponentInternal.h
@@ -65,7 +65,7 @@
 @property (nonatomic, weak) UIView *rootComponentMountedView;
 
 /** For internal use only; don't touch this. */
-@property (nonatomic, strong, readonly) NSNumber *scopeFrameToken;
+@property (nonatomic, strong, readonly) id<NSObject> scopeFrameToken;
 
 @end
 
--- a/ComponentKit/Core/CKComponentInternal.h
+++ b/ComponentKit/Core/CKComponentInternal.h
@@ -65,7 +65,7 @@
 @property (nonatomic, weak) UIView *rootComponentMountedView;
 
 /** For internal use only; don't touch this. */
-@property (nonatomic, strong, readonly) id scopeFrameToken;
+@property (nonatomic, strong, readonly) NSNumber *scopeFrameToken;
 
 @end
 
--- a/ComponentKit/Core/CKComponentLayout.h
+++ b/ComponentKit/Core/CKComponentLayout.h
@@ -62,10 +62,17 @@ NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
                               CKComponent *supercomponent);
 
 /**
- This exists as a safe alternative to layoutThatFits:parentSize: by guarding against nil components
- @param component The component being laid out
- @param sizeRange The size range for laying out the component
- @param parentSize The size of the parent of the component being laid out
+ Safely computes the layout of the given top-level component by guarding against nil components.
+ @param component The top-level component to compute the layout for.
+ @param sizeRange The size range to compute the component layout within.
+ */
+CKComponentLayout CKComputeTopLevelComponentLayout(CKComponent *component, const CKSizeRange &sizeRange);
+
+/**
+ Safely computes the layout of the given component by guarding against nil components.
+ @param component The component to compute the layout for.
+ @param sizeRange The size range to compute the component layout within.
+ @param parentSize The parent size of the component to compute the layout for.
  */
 CKComponentLayout CKComputeComponentLayout(CKComponent *component,
                                            const CKSizeRange &sizeRange,
--- a/ComponentKit/Core/CKComponentLayout.h
+++ b/ComponentKit/Core/CKComponentLayout.h
@@ -67,9 +67,9 @@ NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
  @param sizeRange The size range for laying out the component
  @param parentSize The size of the parent of the component being laid out
  */
-CKComponentLayout CKComponentComputeLayout(CKComponent *component,
+CKComponentLayout CKComputeComponentLayout(CKComponent *component,
                                            const CKSizeRange &sizeRange,
-                                           CGSize parentSize);
+                                           const CGSize parentSize);
 
 /** Unmounts all components returned by a previous call to CKMountComponentLayout. */
 void CKUnmountComponents(NSSet *componentsToUnmount);
--- a/ComponentKit/Core/CKComponentLayout.h
+++ b/ComponentKit/Core/CKComponentLayout.h
@@ -62,11 +62,11 @@ NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
                               CKComponent *supercomponent);
 
 /**
- Safely computes the layout of the given top-level component by guarding against nil components.
- @param component The top-level component to compute the layout for.
+ Safely computes the layout of the given root component by guarding against nil components.
+ @param rootComponent The root component to compute the layout for.
  @param sizeRange The size range to compute the component layout within.
  */
-CKComponentLayout CKComputeTopLevelComponentLayout(CKComponent *component, const CKSizeRange &sizeRange);
+CKComponentLayout CKComputeRootComponentLayout(CKComponent *rootComponent, const CKSizeRange &sizeRange);
 
 /**
  Safely computes the layout of the given component by guarding against nil components.
--- a/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
@@ -57,27 +57,32 @@ namespace std {
   CKCAssert([componentClass isSubclassOfClass:[CKComponent class]], @"%@ is not a component", NSStringFromClass(componentClass));
   CKAssertNotNil(pair.frame, @"Must have frame");
 
-  // Find the existing child, if any, in the old frame
-  CKComponentScopeFrame *existingChild;
+  CKComponentScopeFrame *existingChildFrameOfEquivalentPreviousFrame;
   if (pair.equivalentPreviousFrame) {
     const auto &equivalentPreviousFrameChildren = pair.equivalentPreviousFrame->_children;
     const auto it = equivalentPreviousFrameChildren.find({componentClass, identifier});
-    existingChild = (it == equivalentPreviousFrameChildren.end()) ? nil : it->second;
+    existingChildFrameOfEquivalentPreviousFrame = (it == equivalentPreviousFrameChildren.end()) ? nil : it->second;
   }
 
+  const auto existingChild = pair.frame->_children.find({componentClass, identifier});
+  CKComponentScopeHandleIdentifier existingGlobalIdentifier =
+  existingChild != pair.frame->_children.end() ? existingChild->second.handle.globalIdentifier : 0;
+
   CKComponentScopeHandle *newHandle =
-  existingChild ? [existingChild.handle newHandleWithStateUpdates:stateUpdates] :
-  [[CKComponentScopeHandle alloc] initWithListener:newRoot.listener
-                                    rootIdentifier:newRoot.globalIdentifier
-                                    componentClass:componentClass
-                               initialStateCreator:initialStateCreator];
+  existingChildFrameOfEquivalentPreviousFrame
+  ? [existingChildFrameOfEquivalentPreviousFrame.handle newHandleWithStateUpdates:stateUpdates]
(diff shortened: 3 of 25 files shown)


ComponentKit/Core/CKComponentInternal.h
```diff
@@ -65,7 +65,7 @@
 @property (nonatomic, weak) UIView *rootComponentMountedView;
 
 /** For internal use only; don't touch this. */
-@property (nonatomic, strong, readonly) id scopeFrameToken;
+@property (nonatomic, strong, readonly) NSNumber *scopeFrameToken;
```
adamjernst: Hm I'm not a fan of exposing this implementation detail. Is there really a need to do this? Couldn't we just use isEqual:/hash on id<NSObject>?

ComponentKit/Core/CKComponentLayout.h
```diff
@@ -62,14 +62,21 @@ NSSet *CKMountComponentLayout(const CKComponentLayout &layout,
...
- This exists as a safe alternative to layoutThatFits:parentSize: by guarding against nil components
- @param component The component being laid out
- @param sizeRange The size range for laying out the component
- @param parentSize The size of the parent of the component being laid out
+ Safely computes the layout of the given top-level component by guarding against nil components.
+ @param component The top-level component to compute the layout for.
+ @param sizeRange The size range to compute the component layout within.
  */
-CKComponentLayout CKComponentComputeLayout(CKComponent *component,
+CKComponentLayout CKComputeTopLevelComponentLayout(CKComponent *component, const CKSizeRange &sizeRange);
```
adamjernst: I mildly prefer Root to TopLevel since the former is already a componentkit concept.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -30,6 +30,7 @@
 
 /** Creates a conceptually brand new scope handle */
 - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
+                globalIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier
```
adamjernst: wait if it's conceptually "brand new", why are you passing a global identifier *in* instead of letting it be allocated by the callee?

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
```diff
@@ -0,0 +1,41 @@
...
+#import "CKDetectComponentScopeCollisions.h"
+
+#import <queue>
+
+#import "CKAssert.h"
+#import "CKComponentInternal.h"
+
+void CKDetectComponentScopeCollisions(const CKComponentLayout &layout)
+{
+#if DEBUG
```
adamjernst: Should we introduce a CK_ASSERTIONS_ENABLED flag?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -46,13 +46,14 @@ + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
...
 - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
+                globalIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier
                   rootIdentifier:(CKComponentScopeRootIdentifier)rootIdentifier
                   componentClass:(Class)componentClass
              initialStateCreator:(id (^)(void))initialStateCreator
 {
   static int32_t nextGlobalIdentifier = 0;
   return [self initWithListener:listener
-               globalIdentifier:OSAtomicIncrement32(&nextGlobalIdentifier)
+               globalIdentifier:(globalIdentifier > 0) ? globalIdentifier : OSAtomicIncrement32(&nextGlobalIdentifier)
```
eczarny: This feels a bit wrong to be so I'd appreciate any suggestions.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
```diff
@@ -0,0 +1,41 @@
...
+  std::queue<const CKComponentLayout> queue;
+  NSMutableSet<NSNumber *> *previouslySeenScopeFrameTokens = [NSMutableSet new];
+  queue.push(layout);
+  while (!queue.empty()) {
+    auto currentLayout = queue.front();
+    queue.pop();
+    NSNumber *scopeFrameToken = [currentLayout.component scopeFrameToken];
+    if (scopeFrameToken && [previouslySeenScopeFrameTokens containsObject:scopeFrameToken]) {
+      CKCFailAssert(@"Scope collision. Attempting to create duplicate scope for component %@", [currentLayout.component class]);
```
eczarny: An improvement here would be to include a "backtrace" of components leading to the collision.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
eczarny: Is this sufficient for keeping this added work out of release builds?

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.h
```diff
@@ -0,0 +1,19 @@
...
+
+#import <Foundation/Foundation.h>
+
+struct CKComponentLayout;
+
+/**
+ Detects, and reports, component scope collisions found in the given component layout.
+ @param layout The top-level component layout of the component hierarchy.
+ */
+void CKDetectComponentScopeCollisions(const CKComponentLayout &layout);
```
eczarny: I'll add unit tests in a follow up once I add more complexity to improve the error message.

ComponentKit/Core/CKComponentInternal.h
eczarny: That's what I had originally, but it felt a bit off since component scope collision detection uses this token in a set. Keeping this `id` works, just feels slightly less safe.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
eczarny: `CK_ASSERTIONS_ENABLED` is a good idea. I definitely like that better to using `DEBUG`.

ComponentKit/Core/CKComponentInternal.h
adamjernst: Use `id<NSObject>`? That indicates that you can indeed call `hash`/`isEqual:` on it.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
adamjernst: yes let's do that please!

ComponentKit/Core/CKComponentInternal.h
eczarny: I'm comfortable with using `id`, so I'll change this.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -57,27 +57,31 @@ + (CKComponentScopeFramePair)childPairForPair:(const CKComponentScopeFramePair &
...
   if (pair.equivalentPreviousFrame) {
     const auto &equivalentPreviousFrameChildren = pair.equivalentPreviousFrame->_children;
     const auto it = equivalentPreviousFrameChildren.find({componentClass, identifier});
-    existingChild = (it == equivalentPreviousFrameChildren.end()) ? nil : it->second;
+    existingChildFrameOfEquivalentPreviousFrame = (it == equivalentPreviousFrameChildren.end()) ? nil : it->second;
   }
 
+  const auto existingChild = pair.frame->_children.find({componentClass, identifier});
+  CKComponentScopeHandleIdentifier existingGlobalIdentifier =
+  existingChild != pair.frame->_children.end() ? existingChild->second.handle.globalIdentifier : 0;
```
adamjernst: Would anything break if you just did the following? 

It results in a pretty confusing situation where the map has two entries with the same `CKComponentScopeHandle`, but that actually makes sense I think. For example, we definitely want the state to be the same for each — not separate (created with `initialStateCreator` for the colliding component).

```
if (existingChild != pair.frame->children.end()) {
  return {.frame = existingChild->second, .equivalentPreviousFrame = existingChildFrameOfEquivalentPreviousFrame};
}
```

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -92,6 +92,16 @@ - (instancetype)newHandleWithStateUpdates:(const CKComponentStateUpdateMap &)sta
                                                controller:_controller];
 }
 
+- (instancetype)newHandleWillBeReacquiredDueToScopeCollision
```
eczarny: Instead of actually reusing the same component scope handle I decided to create an identical handle that hasn't already been acquired.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
eczarny: @adamjernst Any suggestions on the best way to go about adding this flag?

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -57,27 +57,47 @@ + (CKComponentScopeFramePair)childPairForPair:(const CKComponentScopeFramePair &
...
+
+       1. Any component state associated with the scope handle will be shared between components with colliding scopes
+       2. Any component controller associated with the scope handle will be responsible for each component with
+          colliding scopes; resulting in strange behavior while components are mounted, unmounted, etc.
+
+     Reusing the existing scope handle allows ComponentKit to detect component scope collisions during layout. Moving
+     component scope collision detection to component layout makes it possible to create multiple components that may
+     normally result in a scope collision even if only one component actually makes it to layout.
+    */
+    CKComponentScopeHandle *newHandle = [existingChild->second.handle newHandleWillBeReacquiredDueToScopeCollision];
```
adamjernst: `newHandleToBeReacquiredDueToScopeCollision` maybe?

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -57,27 +57,47 @@ + (CKComponentScopeFramePair)childPairForPair:(const CKComponentScopeFramePair &
...
+     The component was involved in a scope collision and the scope handle needs to be reacquired.
+     In the event of a component scope collision the component scope frames reuses the existing scope handle; any
+     existing state will be made available to the component that introduced the scope collision. This leads to some
+     interesting side effects:
+
+       1. Any component state associated with the scope handle will be shared between components with colliding scopes
+       2. Any component controller associated with the scope handle will be responsible for each component with
+          colliding scopes; resulting in strange behavior while components are mounted, unmounted, etc.
+
+     Reusing the existing scope handle allows ComponentKit to detect component scope collisions during layout. Moving
```
adamjernst: well, you're not reusing the same handle; you're reusing the same frame, different handle.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
adamjernst: @eczarny well in open source it's just NS_ASSERTIONS_ENABLED, and we have a private internal version at FB that you'd tie to the appropriate macro

adamjernst:
Merge at will!

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
eczarny: Hm. Technically we're reusing the same handle, we just create a new instance with the exact same state so it can be reacquired. An entirely new frame is created to hold that handle.

ComponentKit/Core/Scope/CKDetectComponentScopeCollisions.mm
eczarny: @adamjernst #592.

PR #589, opened 2016-07-14
Best Practice: Keep Controller in Component

ryanrhee:


ryanrhee:
screenshot 2016-07-14 04 36 17

Attached is a rendering of the proposed page.

_docs/42-keep-controller-in-component.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Keep Controller in Component
```
eczarny: Maybe _Keep Component and Component Controllers Together_ would be better?

eczarny:
Looks good! Not sure if you can merge so if you can't let me know.

ryanrhee:
> Keep Component and Component Controllers

Will update!

ryanrhee:
@eczarny it's actually too long, jekyll gets confused :| got an alternative name suggestion?

ryanrhee:
Gonna merge as-is, we can rename it later if need be.

_docs/42-keep-controller-in-component.md
adamjernst: ++

ryanrhee: @adamjernst that's too long, it makes jekyll mess up the rendering :/

Issue #590, opened 2016-07-14
Accessibility Identifier should not be part of Accessibility Context


taiber:
Accessibility Identifier is, unlike its name, not an accessibility feature, but actually an automation feature used to identify elements on the screen. More context here: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIAccessibilityIdentification_Protocol/

Passing accessibilityIdentifier in as part of accessibilityContext is limiting as it only gets set if accessibility is turned on. This prevents UI tests (e.g. WebDriver) from finding elements on the screen when accessibility is off.
The correct way of passing an identifier to a component is just passing `{@selector(setAccessibilityIdentifier:), @"accessibilityId"}` in its attributes.
We have some workarounds in our library like CKButtonComponent manually setting it in its constructor but that's only more confusing as it only works sometimes and when remembering to set it (https://github.com/facebook/componentkit/blob/master/ComponentKit/Components/CKButtonComponent.mm#L149) 

We should remove accessibilityIdentifier from accessibilityContext to prevent people from making that mistake.

ocrickard:
Sounds like this is resolved now?

PR #591, opened 2016-07-14
Removed accessibilityIdentifier from context (fixes issue #590)

taiber:
We shouldn't pass in accessibilityIdentifier as part of accessibility context as it's an automation feature, not an accessibility one. See issue #590 for more details.

This diff removes the use of the identifier, forcing people to pass it in correctly and consistently if they want to use it (e.g. `{@selector(setAccessibilityIdentifier:), @"accessibilityId"}` in attributes).

Test plan: ran all the tests, made sure that passing accessibilityIdentifier correctly works as expected

### DIFF
--- a/ComponentKit/Accessibility/CKComponentAccessibility.h
+++ b/ComponentKit/Accessibility/CKComponentAccessibility.h
@@ -43,17 +43,16 @@ private:
 
 /**
  Separate structure to handle accessibility as we want the components infrastructure to decide wether to use it or not depending if accessibility is enabled or not.
+ Not to be confused with accessibilityIdentifier which is used for automation to identify elements on the screen. To set the identifier pass in {@selector(setAccessibilityIdentifier:), @"accessibilityId"} with the viewConfiguration's attributes
  */
 struct CKComponentAccessibilityContext {
   NSNumber *isAccessibilityElement;
-  NSString *accessibilityIdentifier;
   CKComponentAccessibilityTextAttribute accessibilityLabel;
   CKComponentAction accessibilityComponentAction;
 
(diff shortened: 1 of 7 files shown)


facebook-github-bot-3:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

ComponentKit/Accessibility/CKComponentAccessibility.h
```diff
@@ -43,17 +43,16 @@ struct CKComponentAccessibilityTextAttribute {
 
 /**
  Separate structure to handle accessibility as we want the components infrastructure to decide wether to use it or not depending if accessibility is enabled or not.
+ Not to be confused with accessibilityIdentifier which is used for automation to identify elements on the screen.
```
eczarny: Good idea adding this comment! It may be a good idea to mention that `accessibilityIdentifier` should be set on the view instead of the accessibility context.

eczarny:
Looks great! Thanks @taiber.

benlodotcom:
The identifier is actually used for automatic aggregation of accessibility messages (which definitely can be considered an accessibility feature as well). Happy to split those semantics from the accessibility identifier but we should make sure we have a replacement API then.

taiber:
@benlodotcom, are you referring to the way we aggregate messages in news feed @ fb? That could still be done by accessing the view attributes (just not directly), I'm working with @adamjernst on a diff for that

Issue #592, opened 2016-07-14
Introduce CK_ASSERTIONS_ENABLED


eczarny:
As mentioned in #588 we should add a macro called `CK_ASSERTIONS_ENABLED` that evaluates to true if, you guessed it, assertions are enabled.

PR #594, opened 2016-07-15
Introduce CK_ASSERTIONS_ENABLED

eczarny:
To test I added `CKCFailAssert` immediate after `#if CK_ASSERTIONS_ENABLED` in `CKDetectComponentScopeCollisions`. I then ran the example app with both `Debug` and `Release` build configurations and verified that the former triggered the assertion, the latter did not.


eczarny:
Resolves #592.

PR #595, opened 2016-07-15
Add ability to temporarily override CKThreadLocalComponentScope

javache:
For some threading experiments we're doing, we need to able to temporarily override the current thread local component scope. I encapsulated this in a C++ class so cleanup is guaranteed to happen.

### DIFF
--- a/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
+++ b/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
@@ -28,3 +28,12 @@ public:
   const CKComponentStateUpdateMap stateUpdates;
   std::stack<CKComponentScopeFramePair> stack;
 };
+
+class CKThreadLocalComponentScopeOverride {
+public:
+  CKThreadLocalComponentScopeOverride(CKThreadLocalComponentScope *scope);
+  ~CKThreadLocalComponentScopeOverride();
+
+private:
+  CKThreadLocalComponentScope *const previousScope;
+};
--- a/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
+++ b/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
@@ -29,6 +29,10 @@ public:
   std::stack<CKComponentScopeFramePair> stack;
 };
 
+/**
+ Temporarily overrides the current thread's component scope.
+ Use for testing and advanced integration purposes only.
+ */
 class CKThreadLocalComponentScopeOverride {
 public:
   CKThreadLocalComponentScopeOverride(CKThreadLocalComponentScope *scope);
--- a/ComponentKitTests/Scope/CKComponentScopeTests.mm
+++ b/ComponentKitTests/Scope/CKComponentScopeTests.mm
@@ -219,4 +219,28 @@
   }
 }
 
+#pragma mark - Component scope override
+
+- (void)testComponentScopeCanBeOverridden
+{
+  CKComponentScopeRoot *root = [CKComponentScopeRoot rootWithListener:nil];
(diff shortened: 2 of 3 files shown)


javache:
cc @adamjernst

ComponentKitTests/Scope/CKComponentScopeTests.mm
```diff
@@ -219,4 +219,28 @@ - (void)testComponentScopeHandleGlobalIdentifierIsTheSameBetweenSiblingsWithComp
...
+#pragma mark - Component scope override
+
+- (void)testComponentScopeCanBeOverridden
+{
+  CKComponentScopeRoot *root = [CKComponentScopeRoot rootWithListener:nil];
+  CKThreadLocalComponentScope threadScope(root, {});
+  CKThreadLocalComponentScope *threadScopePtr = &threadScope;
+
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
eczarny: Does the test need to dispatch this work to another thread?

javache: Yeah, I'm not aware of another way of creating a component scope and _not_ making it also the thread scope. I could maybe stack them a afer b, and then overriding with b, but that might interplay in a strange way (and is not actually what we're planning to use it for).

eczarny: I'm a bit concerned with this causing flakiness. I'd try following what the tests above are doing, just nesting scopes manually.

javache: Sure, I can do that. It just won't really test how this will potentially be used.

javache: Actually, I can't test on a single thread, since the CKThreadLocalComponentScope initializer doesn't allow there to be multiple ones on the same thread (https://github.com/facebook/componentkit/blob/master/ComponentKit/Core/Scope/CKThreadLocalComponentScope.mm#L39)

adamjernst: @javache Hm, we could probably let you "fake it" by exposing the `pthread_key_t` for testing; then you could simply call `pthread_set_specific` yourself to fake different threads.

@eczarny @javache what do you think? Better or worse?

ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
```diff
@@ -28,3 +28,12 @@ class CKThreadLocalComponentScope {
   const CKComponentStateUpdateMap stateUpdates;
   std::stack<CKComponentScopeFramePair> stack;
 };
+
+class CKThreadLocalComponentScopeOverride {
```
adamjernst: docblock indicating it's only for testing?

adamjernst:
love it!

ComponentKitTests/Scope/CKComponentScopeTests.mm
javache: That could be an option too, but I feel like this is probably the simpler approach in terms of API surface.

The thread jumping done during this test seems very limited and I don't expect it to cause any issues but if you're particularly worried about this style, I can take another approach.

eczarny: @adamjernst That would work but @javache brings up a good point that it would be a bit more complex.

Let's try the test out as-is and go with the more complicated option if it proves to be flaky.

eczarny:
@javache Looks good! Let's just add the comment @adamjernst mentions and we can get this merged.

PR #599, opened 2016-08-01
Added function to verify validity of changeset

MDSilber:
This will be used in CKSectionedArrayController

### DIFF
--- a/ComponentKit/Utilities/CKArrayControllerChangesetVerification.h
+++ b/ComponentKit/Utilities/CKArrayControllerChangesetVerification.h
@@ -1,0 +1,36 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKArrayControllerChangeset.h>
+
+typedef NS_ENUM(NSUInteger, CKBadChangesetOperationType) {
+  CKBadChangesetOperationTypeNone,
+  CKBadChangesetOperationTypeUpdate,
+  CKBadChangesetOperationTypeInsertSection,
+  CKBadChangesetOperationTypeInsertRow,
+  CKBadChangesetOperationTypeRemoveSection,
+  CKBadChangesetOperationTypeRemoveRow,
+  CKBadChangesetOperationTypeMoveSection,
+  CKBadChangesetOperationTypeMoveRow
+};
+
+/**
+ This function determines whether a given changeset is valid for a particular data source state.
+ In particular, if this function returns true, the caller is guaranteed that for the given
+ data source state (sections parameter), applying the changeset will not cause an index out of
+ bounds crash.
+
+ @param changeset the changeset in question
+ @param sections the current sections (or state) of the data source
+ @return which section is causing an issue, if any. If there's no issue, this function will return CKBadChangesetOperationTypeNone
+ */
+extern CKBadChangesetOperationType CKIsValidChangesetForSections(CKArrayControllerInputChangeset changeset, NSArray<NSArray *> *sections);
--- a/ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
+++ b/ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
@@ -1,0 +1,152 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKArrayControllerChangesetVerification.h"
+
+static NSMutableArray<NSNumber *> *sectionCountsForSections(NSArray<NSArray *> *sections)
+{
+  NSMutableArray *sectionCounts = [NSMutableArray new];
+  for (NSArray *section in sections) {
+    [sectionCounts addObject:@(section.count)];
+  }
+  return sectionCounts;
+}
+
+CKBadChangesetOperationType CKIsValidChangesetForSections(CKArrayControllerInputChangeset changeset, NSArray<NSArray *> *sections)
+{
+  NSMutableArray<NSNumber *> *currentSectionCounts = sectionCountsForSections(sections);
+  __block BOOL invalidChangeFound = NO;
+
+  // Updates
+  changeset.items.enumerateItems(^(NSInteger section, NSInteger index, id<NSObject> object, BOOL *stop) {
+    if (section >= [currentSectionCounts count] || index >= [currentSectionCounts[section] integerValue] || section < 0 || index < 0) {
+      invalidChangeFound = YES;
+      *stop = YES;
+    }
+  }, nil, nil, nil);
+
+  if (invalidChangeFound) {
+    return CKBadChangesetOperationTypeUpdate;
+  }
+
+  /*
+   Removed rows
+   Note we need to use the dictionary of index sets so that we have an index set for each section.
+   For discussion about why we need index sets at all, see the 'removed sections' comment below
+   */
+  NSMutableDictionary<NSNumber *, NSMutableIndexSet *> *rowsToRemove = [NSMutableDictionary new];
+  changeset.items.enumerateItems(nil, ^(NSInteger section, NSInteger index, BOOL *stop) {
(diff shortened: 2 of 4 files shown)


ComponentKit/Utilities/CKArrayControllerChangesetVerification.h
```diff
@@ -0,0 +1,22 @@
...
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKArrayControllerChangeset.h>
+
+/**
+ This function determines whether a given changeset is valid for a particular data source state
```
eczarny: It may be nice to mention what exactly this function considers to be a valid changeset.

MDSilber:
@eczarny i can do that

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+
+#import "CKArrayControllerChangesetVerification.h"
+
+static NSArray<NSNumber *> *sectionCountsForSections(NSArray<NSArray *> *sections)
+{
+  NSMutableArray *sectionCounts = [NSMutableArray new];
+  for (NSArray *section in sections) {
+    [sectionCounts addObject:@(section.count)];
+  }
+  return [sectionCounts copy];
```
eczarny: Meh, for a static function I wouldn't bother with a copy.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+}
+
+BOOL CKIsValidChangesetForSections(CKArrayControllerInputChangeset changeset, NSArray<NSArray *> *sections)
+{
+  NSMutableArray<NSNumber *> *currentSectionCounts = [NSMutableArray arrayWithArray:sectionCountsForSections(sections)];
```
eczarny: Again, no need for another copy. Just have the function return a mutable array.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+  __block BOOL invalidChangeFound = NO;
+
+  // Updates
+  changeset.items.enumerateItems(^(NSInteger section, NSInteger index, id<NSObject> object, BOOL *stop) {
+    if (section >= [currentSectionCounts count] || index >= [currentSectionCounts[section] integerValue]) {
```
eczarny: I imagine section or index won't typically be less than zero. It may be prudent to check regardless.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /** 
+   Removed rows
+   Note we need to use the dictionary of index sets so that we have an index set for each section.
+   For discussion about why we need index sets at all, see the 'removed sections' comment below
+   */
+  NSMutableDictionary<NSNumber *, NSMutableIndexSet *> *rowsToRemove = [NSMutableDictionary new];
+  changeset.items.enumerateItems(nil, ^(NSInteger section, NSInteger index, BOOL *stop) {
+    if (section >= [currentSectionCounts count] || index >= [currentSectionCounts[section] integerValue]) {
```
eczarny: Do we need to do this check again if it was already handled for updates?

eczarny: Ah, never mind. I didn't realize the block was for removals specifically.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+    // Strictly greater than, since you can always insert at the end
+    if (insertion > [currentSectionCounts count]) {
+      return NO;
+    } else {
+      [currentSectionCounts insertObject:@0 atIndex:insertion];
+    }
+  }
+
+  // Moved sections
+  for (std::pair<NSInteger, NSInteger> move : changeset.sections.moves()) {
```
eczarny: Reasonable place for `auto` if you like.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+  changeset.items.enumerateItems(nil, nil, nil, ^(const CKArrayControllerIndexPath &fromIndexPath, const CKArrayControllerIndexPath &toIndexPath, BOOL *stop) {
+    BOOL fromIndexPathSectionInvalid = fromIndexPath.section >= [currentSectionCounts count];
+    BOOL toIndexPathSectionInvalid = toIndexPath.section >= [currentSectionCounts count];
+
+    /**
+     First only check section validity.
+     If we don't do this first, we risk an index out of bounds crash in the item validity check
+     */
+    if (fromIndexPathSectionInvalid ||
+        toIndexPathSectionInvalid) {
```
eczarny: I'd keep this on a single line, easier to read.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,158 @@
...
+      invalidChangeFound = YES;
+      *stop = YES;
+      // Need this return so that it doesn't continue to evaluate
+      return;
```
eczarny: This is a bit fragile. Maybe break this block up into a couple helper functions and add an else condition below (line 140)? That way we don't need to explicitly return here.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
MDSilber: I actively HATE `auto` lol

MDSilber:
@eczarny just updated the PR with the new enum type we discused

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,154 @@
...
+      invalidChangeFound = YES;
+      *stop = YES;
+    }
+  }, nil, nil, nil);
+
+  if (invalidChangeFound) {
+    return CKBadChangesetOperationTypeUpdate;
+  }
+
+  /**
```
eczarny: Nit: `/*` for a regular multi-line comment. No need for `/**` since this isn't a docblock.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
eczarny: Are we doing the `section < 0` or `index < 0` checks here?

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,154 @@
...
+
+  // Moved sections
+  for (std::pair<NSInteger, NSInteger> move : changeset.sections.moves()) {
+    NSInteger origin = move.first;
+    NSInteger destination = move.second;
+
+    if (origin >= [currentSectionCounts count] || destination >= [currentSectionCounts count]) {
+      return CKBadChangesetOperationTypeMoveSection;
+    } else {
+      NSNumber *originValue = currentSectionCounts[origin];
```
eczarny: `originValue` is a bit unclear. If these are section counts then how about `originSectionCount`. Same goes for `destinationValue`.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
```diff
@@ -0,0 +1,154 @@
...
+      NSNumber *destinationValue = currentSectionCounts[destination];
+
+      [currentSectionCounts replaceObjectAtIndex:origin withObject:destinationValue];
```
eczarny: I'm a bit confused by this. Why are we essentially swapping sections? Is that the expected behavior here?

I think this logic could benefit from a bit more explanation in the comments.

eczarny:
@MDSilber back to you.

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
MDSilber: Oh you're right, this might actually not be correct. Looking at it now.

eczarny: Is there a test case that covers this scenario?

MDSilber: @eczarny yeah there are a few. Check out tests with "moveSection" in them (there should be 4 valid cases and 1 invalid case)

MDSilber:
Just rebasing on master

ComponentKit/Utilities/CKArrayControllerChangesetVerification.mm
MDSilber: We can! Adding that now + new tests to cover that

PR #600, opened 2016-08-02
Don't dispatch the deleter off the main thread if the layout vector is empty

jarman:
When you set an CKComponentLayout ivar for the first time, it’s already
initialized with the default initializer. Because of this, the deleter
is triggered with an empty array. The innocuous setting of an ivar
therefore triggers a semi-expensive dispatch. It’s a lot cheaper to
simply delete this object on the main thread.

### DIFF
--- a/ComponentKit/Core/CKComponentLayout.mm
+++ b/ComponentKit/Core/CKComponentLayout.mm
@@ -22,12 +22,16 @@
 
 using namespace CK::Component;
 
+static void _deleteComponentLayoutChild(void *target)
+{
+  delete (std::vector<CKComponentLayoutChild> *)target;
+}
+
 void CKOffMainThreadDeleter::operator()(std::vector<CKComponentLayoutChild> *target)
 {
   if ([NSThread isMainThread]) {
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-      delete target;
-    });
+    // use dispatch_async_f to avoid block allocations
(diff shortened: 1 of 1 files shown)


eczarny:
Oh! Interesting. I'll try and get to this today.

ComponentKit/Core/CKComponentLayout.mm
```diff
@@ -30,7 +30,7 @@ static void _deleteComponentLayoutChild(void *target)
 
 void CKOffMainThreadDeleter::operator()(std::vector<CKComponentLayoutChild> *target)
 {
-  if ([NSThread isMainThread]) {
+  if ([NSThread isMainThread] && target && !(*target).empty()) {
```
eczarny: Can this be `!target->empty()` instead?

ComponentKit/Core/CKComponentLayout.mm
```diff
@@ -30,7 +30,7 @@ static void _deleteComponentLayoutChild(void *target)
 
 void CKOffMainThreadDeleter::operator()(std::vector<CKComponentLayoutChild> *target)
 {
-  if ([NSThread isMainThread]) {
+  if ([NSThread isMainThread] && !target->empty()) {
```
eczarny: We need to still check that `target` exists or else this will crash.

ComponentKit/Core/CKComponentLayout.mm
```diff
@@ -30,7 +30,7 @@ static void _deleteComponentLayoutChild(void *target)
 
 void CKOffMainThreadDeleter::operator()(std::vector<CKComponentLayoutChild> *target)
 {
-  if ([NSThread isMainThread]) {
+  if ([NSThread isMainThread] && target && !target->empty()) {
```
eczarny: 👍 

@jarman mind adding a comment as well?

PR #603, opened 2016-08-05
Include instructions in the README that explain how to open the Xcode projects

eczarny:


README.md
```diff
@@ -26,6 +26,24 @@ If you have CocoaPods and want to quickly try WildeGuess, the ComponentKit demo
 pod try ComponentKit
```
adamjernst: I might drop or demote these instructions

eczarny: Good catch. I'll update before merging.

Issue #504, opened 2016-03-15
Improve Component Controllers


adamjernst:
#### Summary

`CKComponentController.component` should never be nil and should always point to the latest component. There are two cases:

**Before First Mount**
Users frequently want to access the component from the controller before it is mounted for the first time. Currently the `component` property is modified as part of mounting, so it is nil until the first mount.

**While Unmounted**
Users expect the component to be the latest "version" of the component when it is updated while offscreen and unmounted. For example, imagine a state update that occurs while the component is far offscreen. It would be ideal to allow the controller's `willUpdateComponent`/`didUpdateComponent` methods to execute and to update the `component` property even though the component is unmounted.
#### Areas to Investigate

**Creating Controllers**: Controllers are currently created as components are created. It would be better to create them at some later point, after a complete `CKComponentLayout` has been created.

**Updating Controllers**: To fully fix this problem, we'd need to redefine how the controller's current component is determined to divorce it from the mounting process — but hopefully without breaking the invariant that the controller's component is always the currently mounted component.

adamjernst:
One interesting challenge is that `CKComponentController.component` is currently `weak` to prevent a retain cycle. It should probably be `strong` instead (to strengthen the never-nil guarantee) and then we manually break the retain cycle at some later point (when removing the item from the data source, or destroying the component hosting view).

This would also neatly solve #429 by giving us an obvious point to implement a component controller "death" hook (when the `component` property is set to nil).

adamjernst:
… which in turn allows fixing #406, since right now there's no obvious place to fix the leak of the context object.

ocrickard:
So there are 3 clear parts here:

1. controller.component is nil before mount.
2. controller.component is nil after dealloc of component.
3. controller.component may not be the most recent component.

1 is resolved by my PR, but 2 & 3 remains. I'm targeting 0.15 for 1, 2 for 0.16, and 3 for 0.20

fabiomassimo:
Since the last time this got updated the core team has been focused on different features that require this topic to be revisited once again. Closing this and following up on our roadmap.

PR #605, opened 2016-08-17
[ck] Fix controller.component = nil before mount

ocrickard:
As discussed in https://github.com/facebook/componentkit/issues/504
CKComponentController component can be nil until first mount. This diff
changes the initializer of CKComponentController to instead take in a
component on init, and changes the initialization order for controllers
inside scope handles.

### DIFF
--- a/ComponentKit/Core/CKComponentController.h
+++ b/ComponentKit/Core/CKComponentController.h
@@ -79,4 +79,13 @@
  */
 - (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
 
+/** 
+ Initializes a controller with the first generation of component. You should not directly initialize a controller,
+ they are initialized for you by the infrastructure.
+ */
+- (instancetype)initWithComponent:(CKComponent *)component NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)init NS_UNAVAILABLE;
++ (instancetype)new NS_UNAVAILABLE;
+
 @end
--- a/ComponentKit/Core/CKComponentController.mm
+++ b/ComponentKit/Core/CKComponentController.mm
@@ -79,6 +79,14 @@ static void eraseAnimation(CKAppliedComponentAnimationMap &map, CKComponentAnima
   CKAppliedComponentAnimationMap _appliedAnimations;
 }
 
+- (instancetype)initWithComponent:(CKComponent *)component
+{
+  if (self = [super init]) {
+    _component = component;
+  }
+  return self;
+}
+
 - (void)willMount {}
 - (void)didMount {}
 - (void)willRemount {}
--- a/ComponentKit/Core/Scope/CKComponentScope.mm
+++ b/ComponentKit/Core/Scope/CKComponentScope.mm
@@ -17,6 +17,7 @@
 CKComponentScope::~CKComponentScope()
 {
   if (_threadLocalScope != nullptr) {
+    [_scopeHandle resolve];
     _threadLocalScope->stack.pop();
   }
 }
--- a/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
@@ -163,8 +163,6 @@ namespace std {
                                       componentClass:componentClass
                                  initialStateCreator:initialStateCreator];
 
-  [newRoot registerAnnounceableEventsForController:newHandle.controller];
-
   CKComponentScopeFrame *newChild = [[CKComponentScopeFrame alloc] initWithHandle:newHandle];
   pair.frame->_children.insert({{componentClass, identifier}, newChild});
   return {.frame = newChild, .equivalentPreviousFrame = existingChildFrameOfEquivalentPreviousFrame};
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -11,6 +11,7 @@
 #import "CKComponentScopeHandle.h"
 
 #import "CKComponentController.h"
+#import "CKComponentControllerInternal.h"
 #import "CKComponentScopeRootInternal.h"
 #import "CKComponentSubclass.h"
 #import "CKInternalHelpers.h"
@@ -21,8 +22,14 @@
 {
   id<CKComponentStateListener> __weak _listener;
   Class _componentClass;
+  CKComponentController *_controller;
   CKComponentScopeRootIdentifier _rootIdentifier;
   BOOL _acquired;
+  // Temporarily stored reference to the specific component that acquired this handle. This forms a reference cycle
+  // that is broken in `resolve`. This reference will always be partially initialized, and should not be used outside
+  // the `resolve` call.
+  CKComponent *_acquiredComponent;
+  BOOL _resolved;
 }
 
 + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
@@ -56,7 +63,7 @@
                  rootIdentifier:rootIdentifier
                  componentClass:componentClass
                           state:initialStateCreator ? initialStateCreator() : [componentClass initialState]
-                     controller:newController(componentClass)];
+                     controller:nil]; // controllers are built on resolution of the handle
 }
 
 - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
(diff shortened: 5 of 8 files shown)


ComponentKit/Core/CKComponentController.h
```diff
@@ -79,4 +79,14 @@
  */
 - (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
 
+
+/**
+ Initializes a controller with the first generation of component. You should not directly initialize a controller,
+ they are initialized for you by the infrastructure.
+ */
+- (instancetype)initWithComponent:(CKComponent *)component NS_DESIGNATED_INITIALIZER;
```
ocrickard: Although I believe this is cleaner, I want to note for readers that **this is a breaking change with the public API**, which will force people updating ComponentKit to have to fix up their initializers in the subclasses.

The other alternative I considered is pushing onto a thread-local stack and keeping the component as-is to maintain backwards compatibility, but thought this was clearer. Open to other ideas here.

eczarny: Honestly I think this is fine. There are other breaking changes in the roadmap (removing `CKComponentDataSource` in favor of `CKTransactionalComponentDataSource` being the most notable) so I'm not too worried about this one.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -56,7 +58,7 @@ - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
                  rootIdentifier:rootIdentifier
                  componentClass:componentClass
                           state:initialStateCreator ? initialStateCreator() : [componentClass initialState]
-                     controller:newController(componentClass)];
+                     controller:nil];
```
eczarny: I think it may be worth adding a quick comment here explaining that the controller will be created once the component acquires the scope handle.

ocrickard: Good call

eczarny:
I think this looks great! @adamjernst mind taking a quick pass before I accept?

eczarny:
@ocrickard feel free to add the comment and merge at your leisure.

ocrickard:
@eczarny thanks, I'll merge once I verify that this passes all internal tests as well. I still have a few failing.

ocrickard:
After thinking about this more, we're going to be passing in the incompletely-initialized CKComponent on init of the controller, which likely isn't a good idea. I'm going to rework this to instead generate the controller on destruction of the scope, however that has a significant tradeoff: component.controller will *be nil* during CKComponent's initializer.

I actually think that's fine, since components' initializers should not access controllers in init.

ComponentKit/Core/CKComponentController.mm
```diff
@@ -79,6 +79,56 @@ @implementation CKComponentController
...
++ (void)load {
+  NSMutableArray *badClasses = [NSMutableArray array];
+  for (Class klass in ClassGetSubclasses(self)) {
+    if ([klass instanceMethodForSelector:@selector(init)] !=
+        [[CKComponentController class] instanceMethodForSelector:@selector(init)]) {
+      [badClasses addObject:klass];
+    }
+  }
+  FBFatal(badClasses.count == 0, @"found bad component controllers:%@", badClasses);
+}
```
ocrickard: lol, gotta remove this. Hopefully other apps have designated initializer warnings turned on.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -21,8 +22,13 @@ @implementation CKComponentScopeHandle
 {
   id<CKComponentStateListener> __weak _listener;
   Class _componentClass;
+  CKComponentController *_controller;
   CKComponentScopeRootIdentifier _rootIdentifier;
   BOOL _acquired;
+  // Temporarily stored reference to the specific component that acquired this handle. This forms a reference cycle
+  // that is broken in `resolve`.
+  CKComponent *_acquiredComponent;
```
ocrickard: Please review this.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -125,12 +142,28 @@ - (BOOL)acquireFromComponent:(CKComponent *)component
...
   if (!_acquired && [component isMemberOfClass:_componentClass]) {
     _acquired = YES;
+    _acquiredComponent = component;
     return YES;
   } else {
     return NO;
   }
 }
 
+- (void)resolve
```
ocrickard: Let me know what you think of this design. Acquisition is followed by resolution in the scope's destructor.

ocrickard:
@eczarny mind checking this again? I moved controller init from scope handle acquisition to a new phase I call "resolution" of the scope handle. Up until that point, the component.controller will return nil, and will assert, which I believe is better behavior anyway? This revealed several components internally that were doing some pretty bad things. Better that component.controller within new returns nil/asserts.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -125,12 +142,28 @@ - (BOOL)acquireFromComponent:(CKComponent *)component
...
+  if (!_controller) {
+    CKThreadLocalComponentScope *currentScope = CKThreadLocalComponentScope::currentScope();
+    CKAssert(currentScope != nullptr, @"Current scope should never be null here. Thread-local stack is corrupted.");
+
+    // A controller can be non-nil at this callsite during component re-generation because a new scope handle is
+    // generated in a new tree, that is acquired by a new component. We pass in the original component controller
+    // in that case, and we should avoid re-generating a new controller in that case.
+    _controller = newController(_acquiredComponent, currentScope->newScopeRoot);
+  }
+  _acquiredComponent = nil;
```
eczarny: I'd just leave a comment here explaining this is where `CKComponentScopeHandle` explicitly breaks the purposeful retain cycle.

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -163,7 +163,7 @@ The next time the component hierarchy is created (e.g. after a component state u
                                       componentClass:componentClass
                                  initialStateCreator:initialStateCreator];
 
-  [newRoot registerAnnounceableEventsForController:newHandle.controller];
+
```
eczarny: Nit: So much whitespace.

ComponentKit/Core/Scope/CKComponentScope.mm
```diff
@@ -17,6 +17,7 @@
 CKComponentScope::~CKComponentScope()
 {
   if (_threadLocalScope != nullptr) {
+    [_scopeHandle resolve];
```
eczarny: This is clever! Cool approach.

So we completely avoid the problem of a partially initialized component by deferring "resolution" until the component scope is destroyed (and the component has finished initializing). Neat.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
eczarny: I think it's worthwhile calling out that the component `_acquiredComponent` references is likely not completely initialized. I'd hate to see a change later on that tries to pass this component along elsewhere before the handle has been fully resolved.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
eczarny: I think it's kind of nifty. Nice way to work around the problem of controllers receiving a reference to a partially initialized component.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
ocrickard: good idea

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
ocrickard: lol, sorry

eczarny:
> I actually think that's fine, since components' initializers should not access controllers in init.

I do wonder how common (if at all) this is in practice. I tend to agree, but I want to be sure there aren't cases where this assumption no longer holds true.

ocrickard:
Running FB, and trying to navigate all over I found only 2 components that accessed controllers in +new, and every case was unsafe and needed to be redone anyway.

eczarny:
Looks good to me! Let's just add a few more comments and get this merged.

ocrickard:
I think this is ready to merge now. I will wait for CI to kick over tonight, and I have async builds on the FB side that I'm monitoring. Everything **appears** to be working. We'll see what I missed though...

ocrickard:
All CI passed, merging.

PR #610, opened 2016-08-23
Add main thread assertions for mount info

ocrickard:
_mountInfo is a mutable main-thread affined portion of the component. We don't lock on access, yet internally have some crashes due to what appears may be multithreaded access to the mount info. If any of this is ever accessed off-main, then you're likely to corrupt the heap, since you can't control when you are unmounted, and that pointer is unset on the main thread. There is no atomic barrier around the unique_ptr reset().

### DIFF
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -101,6 +101,7 @@ struct CKComponentMountInfo {
 
 - (CKComponentViewContext)viewContext
 {
+  CKAssertMainThread();
   return _mountInfo ? _mountInfo->viewContext : CKComponentViewContext();
 }
 
@@ -111,6 +112,7 @@ struct CKComponentMountInfo {
                                     children:(std::shared_ptr<const std::vector<CKComponentLayoutChild>>)children
                               supercomponent:(CKComponent *)supercomponent
 {
+  CKAssertMainThread();
   // Taking a const ref to a temporary extends the lifetime of the temporary to the lifetime of the const ref
   const CKComponentViewConfiguration &viewConfiguration = CK::Component::Accessibility::IsAccessibilityEnabled() ? CK::Component::Accessibility::AccessibleViewConfiguration(_viewConfiguration) : _viewConfiguration;
 
@@ -154,6 +156,7 @@ struct CKComponentMountInfo {
 
 - (void)unmount
 {
+  CKAssertMainThread();
   if (_mountInfo != nullptr) {
     [_scopeHandle.controller componentWillUnmount:self];
     [self _relinquishMountedView];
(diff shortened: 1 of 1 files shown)


ComponentKit/Core/CKComponent.mm
```diff
@@ -164,12 +167,15 @@ - (void)unmount
...
 - (void)_relinquishMountedView
 {
-  UIView *view = _mountInfo->view;
-  if (view) {
-    CKAssert(view.ck_component == self, @"");
-    [_scopeHandle.controller component:self willRelinquishView:view];
-    view.ck_component = nil;
-    _mountInfo->view = nil;
+  CKAssertMainThread();
+  if (_mountInfo != nullptr) {
```
ocrickard: Somehow internally in hosting views we end up with what appears to be a re-entrant relinquish call. It won't hurt to add this call, since its equivalent actually used to exist in this code path.

eczarny: Is it worth asserting that if `_mountInfo` is `nullptr` or is the main thread assertion adequate? I'd hate to get into a situation where we avoid the crash but somehow remain in a strange state without realizing it.

I guess the question here is what could be the unintended side effects of `_mountInfo` being `nullptr` when we assume it isn't? Prior to this change we had a crash. What would happen after this change?

ocrickard: That's a fantastic point. My understanding is that nothing should change. Every other callsite to _mountInfo has a safety check around it.

ocrickard: Happy to add an assertion here though that _mountInfo is not nullptr.

eczarny: If the behavior is consistent elsewhere I'm less concerned about it. I think the main thread assertions are likely enough.

eczarny:
@ocrickard feel free to merge once CI passes.

ocrickard:
@MDSilber this is failing to build due to this error:

```
❌  Undefined symbols for architecture i386
> Symbol: CKIsValidChangesetForSections(CK::ArrayController::Input::Changeset, NSArray<NSArray*>*)
> Referenced from: -[CKArrayControllerChangesetVerificationTests test_emptyChangesetEmptySections] in CKArrayControllerChangesetVerificationTests.o
```

What's going on here?

ocrickard:
lol @ my ineptitude with git... got too used to mercurial. Rebased, let's see if this works.

ocrickard:
Argh, why is CI taking so long!

ocrickard:
Internal CI is passing, and this merged to our internal repos. I'm merging. @MDSilber is looking into the CI issue.

ocrickard:
https://travis-ci.org/facebook/componentkit/builds/154521889 succeeded, demonstrating that this diff is not the cause of the build failure.

PR #373, opened 2015-09-11
Remount transactional data source

itsthejb:
Resolves https://github.com/facebook/componentkit/issues/342

As noted in the issue, I encountered the same remounting issue with the new data source as encountered in the original (https://github.com/facebook/componentkit/pull/256). The first part of the problem was to use the same technique of grabbing any existing cell for the index path and reusing it. Then an additional problem *seemed* to be that changes were applied to the collection view before updating the local state in `CKCollectionViewTransactionalDataSource`. Simply moving the `applyChangesToCollectionView()` call until after that fixed this.

So... this appears at quick inspection to fix the issue although I'm certainly not 100% confident about the ramifications. I also thought about migrating the example app to the new data source to illustrate the issue, but as far as I can remember it only uses `-updateState`, and not changeset modifications.

Anyway, looking forward to feedback :+1:

PR #612, opened 2016-08-23
Install and update Carthage in build script

itsthejb:
As noted in https://github.com/facebook/componentkit/pull/373, `build.sh` doesn't appear to be fully set up to build using Carthage. Hot-wiring here :). Looking good now.


ocrickard:
yay! Not sure how we landed the other version of this...

PR #615, opened 2016-08-24
Note that CKComponentActionSend must be called on main thread

ryanrhee:
screenshot 2016-08-24 14 17 45


ocrickard:
You can directly edit the docs in the future without doing a PR. Merge at your leisure.

Issue #604, opened 2016-08-16
.flexGrow/.flexShrink should be integer values that represent flex grow/flex shrink factors, not boolean values


eczarny:
The Flexbox specification states that `flex-grow` and `flex-shrink` should be integer value (non-negative integers only) that set the flex grow factor and flex shrink factor, respectively. Currently `CKStackLayoutComponent` only support boolean values for these properties.

PR #618, opened 2016-08-26
Flex factor support

eczarny:
The [Flexbox specification][0] actually relies on the `flex-grow` and `flex-shrink` properties to set the [flex grow factor][1] and [flex shrink factor][2], respectively. These "flex factors" actually describe how any positive or negative free space should be distributed across flex items.

For example, consider you have three flex items within a flex container. The first flex item specifies a `flex-grow` of 1, the second flex item specifies a `flex-grow` of 2, and the third flex item specifies a `flex-grow` of 1 as well. Now assume the flex container itself has a size of, say, 350 _somethings_ (the unit of measurement doesn't really matter for this example). If each flex item has a size of 50 then that leaves 200 remaining to "flex" by.

Based on my changes here the value of 200 is captured by the `violation` and it will need to be distributed over a `flexFactorSum` of 4 (the sum of each flex item's `flex-grow` value). This leaves a `violationPerFlexibleChild` of 50 that needs to be distributed _proportionally_ across each flex item. The amount to distribute to each flex item is once again determined by the `flex-grow` value.

Assuming a `violationPerFlexibleChild` of 50 the remaining space would be distributed in the following way:

- The first `flex-item` would grow by _50_ (`violationPerFlexibleChild` * a `flex-grow` of 1) to a size of **100**
- The second `flex-item` would grow by _100_ (`violationPerFlexibleChild` * a `flex-grow` of 2) to a size of **150**
- The third `flex-item` would grow by _50_ (`violationPerFlexibleChild` * a `flex-grow` of 1) to a size of **100**

`flex-shrink` behaves slightly differently. Instead of distributing the remaining space evenly the `flex-basis` for each `flex-item` is taken into consideration. Per the specification:

> For every unfrozen item on the line, multiply its flex shrink factor by its inner flex base size, and note this as its **scaled flex shrink factor**. Find the ratio of the item’s scaled flex shrink factor to the sum of the scaled flex shrink factors of all unfrozen items on the line. Set the item’s target main size to its flex base size minus a fraction of the absolute value of the remaining free space proportional to the ratio.

This means that each `flex-item` that specifies a non-zero `flex-shrink` factor will flex proportionally to its `flex-basis`. Even if the `flex-shrink` factor is 1 each `flex-item` will shrink by some portion of the remaining violation based on its scaled flex shrink factor ratio.

[0]: https://www.w3.org/TR/css-flexbox-1/
[1]: https://www.w3.org/TR/css-flexbox-1/#flex-flex-grow-factor
[2]: https://www.w3.org/TR/css-flexbox-1/#flex-flex-shrink-factor


eczarny:
This pull request addresses #604.

ComponentKit/LayoutComponents/CKStackLayoutComponent.h
```diff
@@ -78,9 +78,9 @@ struct CKStackLayoutComponentChild {
   /** Additional space to place after the component in the stacking direction. */
   CGFloat spacingAfter;
   /** If the sum of childrens' stack dimensions is less than the minimum size, should this component grow? */
-  BOOL flexGrow;
+  NSInteger flexGrow;
```
eczarny: Migrating to the new type is actually quite simple. `CKStackLayoutComponent` will continue to specify a default of _do not flex_ (a `.flexGrow` of 0 and a `.flexShrink` of 0) for the numeric type. `NO` will presumably continue to evaluate to 0 and `YES` will follow suit and evaluate to 1.

Existing callsites can continue to use `YES` or `NO` while folks more interested in more complex flex behaviors can use a proper flex factor instead. That being said it probably makes sense to do a codemod and ensure everybody is using proper integer values for these two members.

ocrickard:
So cool! cc @appleguy since these layout algorithms are used in both ASDK and CK.

ComponentKit/LayoutComponents/CKStackLayoutComponent.h
```diff
@@ -77,10 +77,18 @@ struct CKStackLayoutComponentChild {
...
+   This value represents the "flex grow factor" and determines how much this component should be grow in relation to any
+   other flexible children.
+   */
+  NSInteger flexGrow;
+  /**
+   If the sum of childrens' stack dimensions is greater than the maximum size, how much should this component shrink?
+   This value represents the "flex shrink factor" and determines how much this component should be shrink in relation to
+   any other flexible children.
+   */
+  NSInteger flexShrink;
```
ocrickard: So does the spec say these are integers, and not floats? Looks like mozilla has support for floating point flex values: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink

eczarny: Good point. These can likely be floats given the specification's definition of a `number`:

https://www.w3.org/TR/css3-values/#number-value

eczarny:
I still plan on adding a few more test cases.

eczarny:
Additional snapshot tests are in.

/cc @ocrickard

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
```diff
@@ -603,44 +583,321 @@ - (void)testCrossAxisStretchingOccursAfterStackAxisFlexing
...
+       .flexShrink = 2,
+     },
+     {
+       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor greenColor]}}} size:{200,50}],
+       .flexShrink = 1,
+     },
+   }];
+  // In this scenario a width of 400 results in a negative violation of 200.
+  // The first and third child components specify a flex shrink factor of 1 and will flex by 50.
+  // The second child component specifies a flex shrink factor of 2 and will flex by 100. It will have a width of 0.
```
ocrickard: But wait, doesn't the spec require that the scaled flex shrink factor would not size the middle element to 0?

https://www.w3.org/TR/css-flexbox-1/#scaled-flex-shrink-factor

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
```diff
@@ -256,30 +255,34 @@ static void flexChildrenAlongStackDimension(std::vector<CKStackUnpositionedItem>
...
   for (CKStackUnpositionedItem &item : items) {
-    if (isFlex(item)) {
+    const NSInteger currentFlexFactor = flexFactor(item);
+    // Children are considered inflexible if they their flex factor is 0 (i.e. both flex-grow and flex-shrink are 0).
+    if (currentFlexFactor != 0) {
       const CGFloat originalStackSize = stackDimension(style.direction, item.layout.size);
-      // The first flexible child is given the additional violation remainder
-      const CGFloat flexedStackSize = originalStackSize + violationPerFlexChild + (isFirstFlex ? violationRemainder : 0);
+      // Now adjust each flexible child by applying the remaining violation based on the flex factor.
+      const CGFloat flexAdjustment = violationPerFlexibleChild * currentFlexFactor;
```
ocrickard: I'm not sure it's this simple. I think the scaled flex factor shoudl be used here instead?

ocrickard:
Back to you for questions!

appleguy:
Very awesome, thanks @ocrickard!  cc @maicki, @Adlai-Holler, @nuygenhuy, @garrettmoon, @levi

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
eczarny: Totally right! I'll fix.

eczarny:
`flex-shrink` now better matches the behavior described in the specification. Unfortunately this a breaking change for `CKStackLayoutComponent` children that specify a `.flexShrink` value of `YES` as they will shrink proportionally (vs. uniformly, based on the violation) to their flex basis.

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
```diff
@@ -97,6 +97,118 @@ static void stretchChildrenAlongCrossDimension(std::vector<CKStackUnpositionedIt
...
+                                                                                          const CGFloat violation)
+{
+  const CGFloat scaledFlexShrinkFactorSum = std::accumulate(items.begin(), items.end(), 0, [&](CGFloat x, const CKStackUnpositionedItem &item) {
+    return x + scaledFlexShrinkFactor(item, style);
+  });
+  return [style, scaledFlexShrinkFactorSum, violation](const CKStackUnpositionedItem &item, BOOL isFirstFlex) {
+    const CGFloat scaledFlexShrinkFactorRatio = scaledFlexShrinkFactor(item, style) / scaledFlexShrinkFactorSum;
+    // The item should shrink proportionally to the scaled flex shrink factor ratio computed above.
+    // Unlike the flex grow adjustment the flex shrink adjustment needs to take the size of each item into account.
+    return -fabs(scaledFlexShrinkFactorRatio * violation);
```
eczarny: Unlike the flex grow adjustment below the flex shrink adjustment doesn't allocate the remaining violation to the first flexible item. This does mean that the resulting layout could result in pixel misalignment. You can see this behavior in the `testOverflowBehaviors_flex@2x.png` reference image below.

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
```diff
@@ -97,6 +97,118 @@ static void stretchChildrenAlongCrossDimension(std::vector<CKStackUnpositionedIt
   }
 }
 
+/** The threshold that determines if a violation has actually occurred. */
+static const CGFloat kViolationEpsilon = 0.01;
```
ocrickard: Nice.

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
```diff
@@ -97,6 +97,118 @@ static void stretchChildrenAlongCrossDimension(std::vector<CKStackUnpositionedIt
...
+static std::function<CGFloat(const CKStackUnpositionedItem &)> flexFactorInViolationDirection(const CGFloat violation)
+{
+  if (fabs(violation) < kViolationEpsilon) {
+    return [](const CKStackUnpositionedItem &item) { return 0; };
+  } else if (violation > 0) {
+    return [](const CKStackUnpositionedItem &item) { return item.child.flexGrow; };
+  } else {
+    return [](const CKStackUnpositionedItem &item) { return item.child.flexShrink; };
+  }
+}
```
ocrickard: Interesting!

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
ocrickard: I think that's OK for now.

ocrickard:
This is beautiful. Thanks for doing this @eczarny. Merge when you're ready.

eczarny:
I plan on merging this Monday.

eczarny:
Oops. Completely deleted this branch by accident. Didn't realize the pull request would be closed.

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
eczarny: Actually. I need to update the comment and the test name. I'll update before merging.

eczarny:
The reference images are a bit of a mess. I'm planning to clean them up substantially next week.

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3886489) to review internal test results.

PR #627, opened 2016-08-31
Remove xcpretty from the CI build script

eczarny:
`xcpretty` hides potentially helpful output when there are build failures. Due to recent CI flakiness I think it makes sense to drop `xcpretty` support to get more verbose output.


eczarny:
The CI build failed due to #630 so I'm going to go ahead and merge this.

PR #631, opened 2016-09-01
Add tvOS as a deployment target to the Podspec

eczarny:
I think this is all that needs to be done to support tvOS in the Podspec. I'll need to do a bit of testing to confirm though.

### DIFF
--- a/ComponentKit.podspec
+++ b/ComponentKit.podspec
@@ -1,16 +1,14 @@
 Pod::Spec.new do |s|
-  s.name = "ComponentKit"
-  s.version = "0.15"
-  s.summary = "A React-inspired view framework for iOS"
-  s.homepage = "https://componentkit.org"
-  s.authors = 'adamjernst@fb.com'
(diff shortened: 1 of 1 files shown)


ocrickard:
@eczarny is this still needed?

eczarny:
Likely. I'm going to create a dummy tvOS app to see if this works. Figured if we support tvOS folks using CocoaPods should be able to use ComponentKit in a tvOS app.

eczarny:
`pod lib lint` works so I'm going to merge so I can fully test in a dummy app.

eczarny:
Confirmed that this works with a dummy app and the following Podfile:

```
platform :tvos, '9.2'

target 'DummyApp' do
  pod 'ComponentKit', :git => 'https://github.com/facebook/componentkit.git'
end
```

eczarny:
Doh. For some reason I thought this was accepted. Sorry about that.

PR #632, opened 2016-09-01
Fix Travis CI for good

eczarny:


eczarny:
First run:

https://travis-ci.org/facebook/componentkit/builds/156704445

Looks like the logs are too long. WTF. I may need to add xcpretty back just to deal with that.

eczarny:
The one nice thing about using a build matrix is we get parallel builds. Nice.

eczarny:
I think this is getting close. I still need to fix the WildeGuess build but otherwise the ComponentKit and the ComponentKitAppleTV tests succeed.

eczarny:
Okay, quick summary of where this is so far:

1. I replaced `carthage update` with `carthage bootstrap` so we don't actually update the tests dependencies; now Carthage just pulls down the versions we wants and builds them
2. I reintroduced xcpretty to help reduce the size of the build log, which is something that Travis CI complained about (and failed a build for)
3. I created a build matrix for the ComponentKit, ComponentKitAppleTV, and WildeGuess projects/targets; this gives us a bit more speed and a more obvious idea what failed

eczarny:
Success: https://travis-ci.org/facebook/componentkit/builds/156836795

ocrickard:
Looks reasonable to me.

PR #635, opened 2016-09-02
Improvements to CKComponentContext

adamjernst:
See changesets for details.


ComponentKit/Utilities/CKComponentContext.h
```diff
@@ -13,42 +13,42 @@
...
+ You may nest contexts with the same class, in which case the innermost context defines the value when fetched:
+
+ {
+   CKComponentContext<CKFoo> c1(foo1);
+   {
+     CKComponentContext<CKFoo> c2(foo2);
+     // CKComponentContext<CKFoo>::get() will return foo2 here
+   }
+   // CKComponentContext<CKFoo>::get() will return foo1 here
+ }
```
eczarny: Great comments.

ComponentKit/Utilities/CKComponentContextHelper.mm
```diff
@@ -25,19 +25,19 @@
...
+  return {.key = key, .originalValue = originalValue, .newValue = object};
 }
 
-void CKComponentContextHelper::clear(id key)
+void CKComponentContextHelper::clear(const CKComponentContextStoreResult &storeResult)
 {
-  NSMutableDictionary *c = contextDictionary(NO);
-  CKCAssertNotNil(c[key], @"Who removed %@ behind our back?", key);
-  [c removeObjectForKey:key];
+  NSMutableDictionary *c = contextDictionary(YES);
```
eczarny: Nit: `const` would be good here as well, especially since we do the same thing in the method above.

ComponentKit/Utilities/CKComponentContextHelper.mm
```diff
@@ -25,19 +25,19 @@
...
+  CKCAssert(c[storeResult.key] == storeResult.newValue, @"Context value for %@ unexpectedly mutated", storeResult.key);
```
eczarny: Much clearer assertion message, nice.

ComponentKit/Utilities/CKComponentContextHelper.mm
eczarny: Hm. Do we want to create the context dictionary here? Is that change on purpose?

eczarny:
Looks fine to me, I added a couple inline comments to the outdated diffs (GitHub doesn't really do a good job surfacing them) as I reviewed each commit.

ComponentKit/Utilities/CKComponentContextHelper.mm
adamjernst: @eczarny yes in fact we have to, it's very subtle. See the last unit test, which tests a very subtle case. It's hard to explain but if you change YES to NO here, you'll see that the test breaks.

eczarny: Gotcha. Probably worth calling that out in the code with a brief comment pointing to the unit test. Especially since it's subtle.

PR #637, opened 2016-09-05
Re-attach existing cells using new state to support animations in reloads

itsthejb:
Hi all :),

So had a chance to revisit this (https://github.com/facebook/componentkit/pull/373) and refresh my memory. New version should be cleaner and more sensical.

In particular, found a better way to approach the test: by force-casting `CKTransactionalComponentDataSourceConfiguration` to accept the `XCTestCase` instance itself as the component provider, it's possible to retain the generated components and access the controller directly. Then, can just log the callbacks received and everything becomes straightforward and conventional enough. Not to mention far shorter.

Other comments on their respective lines.

Thanks!

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -65,9 +65,13 @@ CKTransactionalComponentDataSourceListener
                               userInfo:userInfo];
 }
 
-static void applyChangesToCollectionView(CKTransactionalComponentDataSourceAppliedChanges *changes, UICollectionView *collectionView)
+static void applyChangesToCollectionView(UICollectionView *collectionView, CKComponentDataSourceAttachController *attachController, CKTransactionalComponentDataSourceState *currentState, CKTransactionalComponentDataSourceAppliedChanges *changes)
 {
-  [collectionView reloadItemsAtIndexPaths:[changes.updatedIndexPaths allObjects]];
+  [changes.updatedIndexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, BOOL *stop) {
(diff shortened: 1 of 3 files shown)


ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -85,7 +89,7 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
                        byApplyingChanges:(CKTransactionalComponentDataSourceAppliedChanges *)changes
 {
   [_collectionView performBatchUpdates:^{
-    applyChangesToCollectionView(changes, _collectionView);
+    applyChangesToCollectionView(_collectionView, _attachController, [_componentDataSource state], changes);
```
itsthejb: It was not necessary to change the order of operations here after all: just pass the current state as one of the new the arguments.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -65,9 +65,13 @@ - (void)applyChangeset:(CKTransactionalComponentDataSourceChangeset *)changeset
                               userInfo:userInfo];
 }
 
-static void applyChangesToCollectionView(CKTransactionalComponentDataSourceAppliedChanges *changes, UICollectionView *collectionView)
+static void applyChangesToCollectionView(UICollectionView *collectionView, CKComponentDataSourceAttachController *attachController, CKTransactionalComponentDataSourceState *currentState, CKTransactionalComponentDataSourceAppliedChanges *changes)
```
itsthejb: I changed the order of the arguments to more closely match `-transactionalComponentDataSource:didModifyPreviousState:byApplyingChanges:`

itsthejb:
Looks like some Travis issues on `master`, had to pull all the way back to https://github.com/facebook/componentkit/releases/tag/0.15

PR #638, opened 2016-09-06
Update docs on component to controller data flow

ryanrhee:
Since we added the new `-initWithComponent:` we should document it

Attached is a screenshot of the changed section:

screenshot 2016-09-06 19 44 39

/cc @ocrickard

### DIFF
--- a/_docs/12-component-controllers.md
+++ b/_docs/12-component-controllers.md
@@ -33,4 +33,52 @@ But sometimes, you do need an object with a longer lifecycle. *Component control
 
 There is a only a one-way communication channel between the component and its component controller - you can only pass data off of a component to a component controller. A component has no reference its corresponding component controller. This is by design. 
 
-To pass data from a component to its controller, expose a `@property` on the component in a class extension. The controller can read those properties in `didUpdateComponent`.
+To pass data from a component to its controller, expose a `@property` on the component in a class extension. The controller can initialize itself with the properties in `initWithComponent:`. If these properties will be changing in subsequent state changes (i.e. a new component is being created with different values for these properties), keep them up to date in `didUpdateComponent`.
+
+{% highlight objc %}
+@interface MySongComponent()
+@property (nonatomic, strong, readonly) MySong *song;     // All components for a controller share the same value
+@property (nonatomic, assign, readonly) BOOL isPlaying;   // Different components may have different values (part of component state)
+@end
+@implementation MySongComponent : CKCompositeComponent
++ (instancetype)newWithSong:(MySong *)song
+{
+  CKComponentScope scope(self, song.unique_id);
+  const BOOL isPlaying = [scope.state() boolValue];
+  MySongComponent *const c =
+  [MySongComponent
+   newWithComponent:[SongUIComponent
+                     newWithIsPlaying:isPlaying]];
+  if (c) {
+    c->_song = song;
+    c->_isPlaying = isPlaying;
+  }
+  return c;
+}
+@end
+@interface MySongComponentController : CKComponentController
+@end
+@implementation MySongComponentController
+{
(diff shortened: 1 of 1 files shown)


_docs/12-component-controllers.md
```diff
@@ -33,4 +33,48 @@ But sometimes, you do need an object with a longer lifecycle. *Component control
 
 There is a only a one-way communication channel between the component and its component controller - you can only pass data off of a component to a component controller. A component has no reference its corresponding component controller. This is by design. 
 
-To pass data from a component to its controller, expose a `@property` on the component in a class extension. The controller can read those properties in `didUpdateComponent`.
+To pass data from a component to its controller, expose a `@property` on the component in a class extension. The controller can initialize itself with the properties here in `initWithComponent:`. If these properties will be changing in subsequent state changes (i.e. a new component is being created with different values for these properties), keep them up to date in `didUpdateComponent`.
```
eczarny: > The controller can initialize itself with the properties here in `initWithComponent:`.

Should that sentence be written a bit differently? Maybe "the controller can initialize itself with the properties in `initWithComponent:`" instead?

eczarny:
Nice example!

_docs/12-component-controllers.md
```diff
@@ -33,4 +33,48 @@ But sometimes, you do need an object with a longer lifecycle. *Component control
 
 There is a only a one-way communication channel between the component and its component controller - you can only pass data off of a component to a component controller. A component has no reference its corresponding component controller. This is by design.
```
ocrickard: > A component has no reference its corresponding component controller.

Actually not true. component.controller is the component's associated controller. We *discourage* accessing it, and *forbid* access to it inside the component's initializer via an assertion, and return nil if access is attempted.

_docs/12-component-controllers.md
```diff
@@ -33,4 +33,48 @@ But sometimes, you do need an object with a longer lifecycle. *Component control
...
+  }
+  return self;
+}
+- (void)songStateDidChange:(BOOL)isPlaying
+{
+  [self.component updateState:^{
+    return @(isPlaying);
+  } mode:CKUpdateModeAsynchronous];
+}
+@end
```
ocrickard: This is awesome. Considering that you updateState here, but only implement initWithComponent:, what do you think of also implementing didUpdateComponent in your example to clarify that state updates will not flow through the initializer path?

_docs/12-component-controllers.md
ryanrhee: @ocrickard : Yup, but I didn't update this section since it was already there, and since we discourage it anyway. I think this PR should focus on the new `initWithComponent:` method you added, and we can figure out separately whether to update this bit about components accessing their controllers.

ocrickard: Sounds good to me.

_docs/12-component-controllers.md
eczarny: /cc @ryanrhee

ryanrhee: @ocrickard @eczarny I see what you guys are saying. I could add a statement that logs something in `didUpdateComponent:` and mention in a comment that this will only log when a state *changes*, but not through the initializer path. Thoughts?

eczarny: @ryanrhee sounds like a great idea!

ryanrhee:
oops, i totally dropped the ball on this. will address comments now :x

ryanrhee:
@orickard are you happy w/ the way i've added `didUpdateComponent:` ?

ryanrhee:
@eczarny addressed your rewording concern here as well, lmk if that looks better

ocrickard:
Looks great to me!

PR #641, opened 2016-09-08
[CKImageComponent] Allow to pass in additional attributes

jaydee3:
In the same way CKNetworkImageComponent allows it.


ocrickard:
So cool! Can we get some snapshot tests for this as well?

jaydee3:
It's nothing new? It's an existing component.

ocrickard:
Yeah, realize it isn't new, but we do like snapshot tests for all new view behavior in the core components. I'm cool with fixing forward if you'd prefer to do that.

jaydee3:
Sounds great. I can add them in another PR.

ComponentKit/Components/CKImageComponent.h
```diff
@@ -18,14 +18,16 @@ struct CKComponentSize;
...
 
 /**
- Uses a static layout with the given image size.
+ Uses a static layout with the given image size and applies additional attributes.
  */
 + (instancetype)newWithImage:(UIImage *)image
+                  attributes:(const CKViewComponentAttributeValueMap &)attributes
                         size:(const CKComponentSize &)size;
 
 @end
```
ocrickard: This is so sad that we have multiple initializers with one of our core components.

jaydee3: I'm happy to kill the first one, but it definitely adds convenience. I can split it out as a CKAutoSizeImageComponent?

ocrickard: That would be awesome. Doesn't need to be in this diff.

ocrickard:
Looks reasonable to me. Merge into master whenever you like. Be sure to submit the internal merge soon after though.

jaydee3:
Alright I split it out and added snapshot tests.

ComponentKit/Components/CKImageComponentAutoSized.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#import <ComponentKit/CKCompositeComponent.h>
+
+struct CKComponentSize;
+
+/**
+ A component that displays an image using UIImageView.
+ */
+@interface CKImageComponentAutoSized : CKCompositeComponent
```
ocrickard: This violates the naming convention we have. CKAutoSizedImageComponent would be appropriate.

ocrickard:
CI failed due to snapshot tests. These need to be recorded for many different simulators sadly.

jaydee3:
Okay it should all work now.

ocrickard:
Bold, I like it. I'm in favor of this change, though it is a breaking API change. We've been pretty comfortable with these types of changes in the past, so long as you're cool with merging internally, I'm on board.

jaydee3:
Let's do it!

Issue #640, opened 2016-09-07
CKComponentScopeHandle can create CKComponentController if nil _acquiredComponent


ocrickard:
If you declare a CKComponentScope(self), but then early-return nil from inside your +new method, then we can create a CKComponentController in the scope handle. This shouldn't really cause any harm, but instead it should just have a nil component controller.

PR #642, opened 2016-09-08
Add assertion & avoid useless controller creation

ocrickard:
Fixes https://github.com/facebook/componentkit/issues/640 and is more
defensive with the resolution assertion.

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -152,7 +152,10 @@
 
 - (void)resolve
 {
-  if (!_controller) {
+  CKAssertFalse(_resolved);
+  // _acquiredComponent may be nil if a component scope was declared before an early return. In that case, the scope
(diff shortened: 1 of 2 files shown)


ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -152,7 +152,10 @@ - (BOOL)acquireFromComponent:(CKComponent *)component
 
 - (void)resolve
 {
-  if (!_controller) {
+  CKAssertFalse(_resolved);
```
eczarny: A bit of a bummer that this assertion doesn't take a description.

ocrickard:
So! Turns out this also fixes the problem internally with early-returning causing no controller creation later. I'm adding a unit test to confirm before landing.

ocrickard:
Confirmed, the added test fails without this patch applied, and passes when it succeeds.

eczarny:
Awesome! Feel free to merge once CI gives the go ahead.

PR #643, opened 2016-09-09
Move CKTestRunLoopRunning to ComponentKitTestLib and make it public

indragiek:
Making this public so that other libraries that build on top of ComponentKit can use this functionality in their own unit tests.


ocrickard:
Looks good to me. Merge once CI passes.

PR #652, opened 2016-09-21
Register for announceable events after state updates

eczarny:
To help address nil components in component controllers we changed the lifecycle of component controllers. Unfortunately this introduced a regression in "announceable" event registration. More concretely folks are not receiving the following appearance events after a state update has been applied:

1. `-[CKComponentController componentTreeWillAppear]`
2. `-[CKComponentController componentTreeDidDisappear]`

To fix this problem `CKComponentScopeHandle` must make sure to register component controllers for announceable events with the new `CKComponentScopeRoot`. This should only happen when an existing `CKComponentScopeHandle` is updated with the relevant state updates.

Prior to this change the only time a component controller would register for announceable events is at the point in time the component controller was created.

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -85,12 +85,14 @@
 }
 
 - (instancetype)newHandleWithStateUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+                       componentScopeRoot:(CKComponentScopeRoot *)componentScopeRoot
 {
   id updatedState = _state;
   const auto range = stateUpdates.equal_range(_globalIdentifier);
   for (auto it = range.first; it != range.second; ++it) {
     updatedState = it->second(updatedState);
   }
+  [componentScopeRoot registerAnnounceableEventsForController:_controller];
   return [[CKComponentScopeHandle alloc] initWithListener:_listener
                                          globalIdentifier:_globalIdentifier
                                            rootIdentifier:_rootIdentifier
@@ -205,7 +207,6 @@ static CKComponentController *newController(CKComponent *component, CKComponentS
     CKCAssert([controllerClass isSubclassOfClass:[CKComponentController class]],
               @"%@ must inherit from CKComponentController", controllerClass);
     CKComponentController *controller = [[controllerClass alloc] initWithComponent:component];
-
     [root registerAnnounceableEventsForController:controller];
     return controller;
   }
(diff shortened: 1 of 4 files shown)


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @dbworku to be a potential reviewer.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -85,12 +85,14 @@ - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
...
 
 - (instancetype)newHandleWithStateUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+                       componentScopeRoot:(CKComponentScopeRoot *)componentScopeRoot
 {
   id updatedState = _state;
   const auto range = stateUpdates.equal_range(_globalIdentifier);
   for (auto it = range.first; it != range.second; ++it) {
     updatedState = it->second(updatedState);
   }
+  [componentScopeRoot registerAnnounceableEventsForController:_controller];
```
eczarny: I explicitly decided to pass in the component scope root instead of grabbing it from the current scope. It felt a bit safer.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -205,7 +207,6 @@ static Class controllerClassForComponentClass(Class componentClass)
     CKCAssert([controllerClass isSubclassOfClass:[CKComponentController class]],
               @"%@ must inherit from CKComponentController", controllerClass);
     CKComponentController *controller = [[controllerClass alloc] initWithComponent:component];
-
     [root registerAnnounceableEventsForController:controller];
```
eczarny: This is the bug, more or less. Since `CKComponentScopeHandle` only registers a component controller for receiving announceable events here the association will be lost when a new component scope root is created. Typically when a state update takes place.

ocrickard: Uggh. Nice find.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
ocrickard: Oh man, you're right!

ocrickard:
Very nice find. Looks reasonable to me.

eczarny:
@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing. If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3903745) to review internal test results.

eczarny:
I'll merge once CI passes. Unfortunately that may be a while:

https://www.traviscistatus.com/incidents/v0bxrn0hff5v

PR #658, opened 2016-10-01
Treat flex factors as "true" factors

eczarny:
Had a bit of a derp moment with this bug. Turns out `CKUnpositionedLayout` was not actually treating flex factors as "true" factors. This meant that `flex-grow` and `flex-shrink` could only be expressed in one way: as whole numbers used to determine how much of the violation to assign to each flexible child.

Technically speaking there is no such restriction on these values. You can use any (positive) values that you want. The layout algorithm should simply take each flexible item's flex factor over the sum of flex factors to determine how much of the total violation each flexible child should receive.

This change results in the following two layouts being equivalent:

```
[CKStackLayoutComponent
   newWithView:kWhiteBackgroundView
   size:{}
   style:{
     .direction = CKStackLayoutDirectionHorizontal
   }
   children:{
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor redColor]}}} size:{50,50}],
       .flexGrow = 1,
     },
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor blueColor]}}} size:{50,50}],
       .flexGrow = 0,
     },
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor greenColor]}}} size:{50,50}],
       .flexGrow = 1,
     },
   }]
```

In the above layout each flexible child will receive 1/2 of the available violation. That's each flex factor over the sum of flex factors.

```
[CKStackLayoutComponent
   newWithView:kWhiteBackgroundView
   size:{}
   style:{
     .direction = CKStackLayoutDirectionHorizontal
   }
   children:{
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor redColor]}}} size:{50,50}],
       .flexGrow = 0.5,
     },
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor blueColor]}}} size:{50,50}],
       .flexGrow = 0,
     },
     {
       [CKComponent newWithView:{[UIView class], {{@selector(setBackgroundColor:), [UIColor greenColor]}}} size:{50,50}],
       .flexGrow = 0.5,
     },
   }]
```

The above is expressed slightly differently but the behavior is the same. Each flex factor of 0.5 over 1.0 results in the same violation distribution.

**Note:** _The changes here do introduce a subtle shift in behavior for distributing any violation remainder. Any notable changes in layout are very minor however. The first flexible item may receive a slightly larger remainder than previously implemented._


eczarny:
@facebook-github-bot import

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
```diff
@@ -351,17 +351,23 @@ static void flexChildrenAlongStackDimension(std::vector<CKStackUnpositionedItem>
...
-                                                                                                                    style,
-                                                                                                                    violation,
-                                                                                                                    flexFactorSum);
+  std::function<CGFloat(const CKStackUnpositionedItem &)> flexAdjustment = flexAdjustmentInViolationDirection(items,
+                                                                                                              style,
+                                                                                                              violation,
+                                                                                                              flexFactorSum);
+
+  // Compute any remaining violation. Any remaining violation it will be allocated to the first flexible child.
+  const CGFloat remainingViolation = std::accumulate(items.begin(), items.end(), violation, [&](CGFloat x, const CKStackUnpositionedItem &item) {
```
eczarny: It's a bit unfortunate to do a completely separate pass just to compute the remaining violation. As an alternative I can compute the remaining violation in the loop below and the apply it to the first flexible child.

This felt a bit cleaner though and I don't really anticipate this being too expensive.

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
```diff
@@ -617,6 +617,35 @@ - (void)testPositiveViolationIsDistributedEquallyAmongFlexibleChildComponents
   CKSnapshotVerifyComponent(c, kSize, nil);
 }
 
+- (void)testPositiveViolationIsDistributedEquallyAmongFlexibleChildComponentsExpressedAsPercentages
```
eczarny: I went with `*ExpressedAsPercentages` due to a lack of a better alternative.

ComponentKit/LayoutComponents/CKStackUnpositionedLayout.mm
adamjernst: I think it's fine.

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
adamjernst: Sounds good to me.

eczarny:
I'll merge Monday.

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
```diff
@@ -681,6 +740,40 @@ - (void)testPositiveViolationIsDistributedEquallyAmongGrowingAndShrinkingFlexibl
   CKSnapshotVerifyComponent(c, kSize, nil);
 }
 
+- (void)testPositiveViolationIsDistributedEquallyAmongGrowingAndShrinkingFlexibleChildComponentsExpressedAsPercentages
```
GotEmB: nit: How about `*ArbitraryFloats`? Since the flex factors in this test isn't really adding up to 1.

ComponentKitApplicationTests/CKStackLayoutComponentTests.mm
```diff
@@ -716,6 +809,41 @@ - (void)testPositiveViolationIsDistributedProportionallyAmongGrowingAndShrinking
   CKSnapshotVerifyComponent(c, kSize, nil);
 }
 
+- (void)testPositiveViolationIsDistributedProportionallyAmongGrowingAndShrinkingFlexibleChildComponentsExpressedAsPercentages
```
GotEmB: Same nit here.

eczarny:
@facebook-github-bot import

eczarny:
@facebook-github-bot import

PR #663, opened 2016-10-05
Allow CK::map to work with any iterable

keykoo:
These should be using the std::begin/std::end iterators explicitly.

Previous this worked with std::collections due to argument dependent lookup, but by being explicit we can allow any collection class that is compatible with std::iterators to be passed to the iterable map.


eczarny:
Looks good to me. I'll ship once CI passes.

eczarny:
@facebook-github-bot shipit

PR #665, opened 2016-10-10
Do not attempt to apply bounds animations if the collection view is not visible

eczarny:
It's possible that the collection view driving bounds animations may not be visible. In these situations the mapping of index paths to original layout attributes will be empty. If ComponentKit cannot determine the largest animating visible element it will crash when trying to lookup the layout attributes.

To avoid the crash bounds animations should not be applied if the visible rect is empty.


ocrickard:
Perfect!

ocrickard:
@facebook-github-bot shipit

PR #666, opened 2016-10-18
Add support for CKTransactionalComponentDataSourceChangeset verification

eczarny:
A `CKTransactionalComponentDataSourceChangeset` port of #599.

### DIFF
--- a/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
+++ b/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
@@ -1,0 +1,161 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <UIKit/UIKit.h>
+
+#import "CKTransactionalComponentDataSourceChangesetVerification.h"
+
+#import <ComponentKit/CKTransactionalComponentDataSourceChangesetInternal.h>
+#import <ComponentKit/CKTransactionalComponentDataSourceStateInternal.h>
+
+static NSMutableArray<NSNumber *> *sectionCountsForState(CKTransactionalComponentDataSourceState *state);
+
+static NSArray<NSIndexPath *> *sortedIndexPaths(NSArray<NSIndexPath *> *indexPaths);
+
+CKBadChangesetOperationType CKIsValidChangesetForState(CKTransactionalComponentDataSourceChangeset *changeset,
+                                                       CKTransactionalComponentDataSourceState *state)
+{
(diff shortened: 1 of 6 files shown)


ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
```diff
@@ -0,0 +1,161 @@
...
+
+CKBadChangesetOperationType CKIsValidChangesetForState(CKTransactionalComponentDataSourceChangeset *changeset,
+                                                       CKTransactionalComponentDataSourceState *state)
+{
+  NSMutableArray<NSNumber *> *sectionCounts = sectionCountsForState(state);
+  __block BOOL invalidChangeFound = NO;
+  // Updated items
+  [changeset.updatedItems enumerateKeysAndObjectsUsingBlock:^(NSIndexPath * _Nonnull indexPath, id _Nonnull model, BOOL * _Nonnull stop) {
+    const NSInteger section = indexPath.section;
+    const NSInteger item = indexPath.item;
```
eczarny: I chose to remain consistence and use the `UICollectionView` variant. Not really sure this has implications for a component data source that may be backed by a `UITableView`?

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
```diff
@@ -0,0 +1,161 @@
...
+  }];
+  if (invalidChangeFound) {
+    return CKBadChangesetOperationTypeInsertSection;
+  }
+  /*
+   Inserted items
+   Section counts may immediately reflect insertions as they are guaranteed to be contiguous by virtue of sorting the index paths.
+   As long as each item is located within the bounds of its section the changeset is valid.
+   */
+  [sortedIndexPaths(changeset.insertedItems.allKeys) enumerateObjectsUsingBlock:^(NSIndexPath * _Nonnull indexPath, NSUInteger index, BOOL * _Nonnull stop) {
```
eczarny: This case is a bit weird since there is no guaranteed order of item insertions. I chose to sort the index paths so that they may be inserted contiguously. If for some reason there's a "gap", or an item is inserted outside the bounds of the current section, the insertion will be invalid.

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
ocrickard: For Index path? There should be no problem here. section/item work properly as you would expect.

ocrickard:
Looks reasonable to me.

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
eczarny: There's also section/row for `UITablevView`. I'm assuming they're interchangeable.

ocrickard:
@facebook-github-bot shipit

PR #667, opened 2016-10-19
Verify transactional component data source changesets

eczarny:
Adds an assertion to `CKTransactionalComponentDataSource` as changesets are applied.

Test Plan:

Trigger bad changesets in WildeGuess and verify assertions are raised.

### DIFF
--- a/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
+++ b/ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
@@ -10,24 +10,19 @@
 
 #import <UIKit/UIKit.h>
 
-#import <map>
-#import <unordered_map>
-
 #import "CKTransactionalComponentDataSourceChangesetVerification.h"
 
 #import <ComponentKit/CKTransactionalComponentDataSourceChangesetInternal.h>
 #import <ComponentKit/CKTransactionalComponentDataSourceChangesetModification.h>
 #import <ComponentKit/CKTransactionalComponentDataSourceStateInternal.h>
 
-static CKTransactionalComponentDataSourceState *foldModificationsIntoState(CKTransactionalComponentDataSourceState *state,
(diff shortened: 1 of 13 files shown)


Examples/WildeGuess/WildeGuess/WildeGuessCollectionViewController.mm
```diff
@@ -68,7 +68,7 @@ - (void)viewDidLoad
   [[[CKTransactionalComponentDataSourceChangesetBuilder transactionalComponentDataSourceChangeset]
     withInsertedSections:[NSIndexSet indexSetWithIndex:0]]
    build];
-  [_dataSource applyChangeset:initialChangeset mode:CKUpdateModeAsynchronous userInfo:nil];
+  [_dataSource applyChangeset:initialChangeset mode:CKUpdateModeSynchronous userInfo:nil];
```
eczarny: Thinking about this a bit more I don't think we can reliably verify changesets like this. For correctness it may be necessary to "fold" all pending changesets into the current state and verify the resulting state.

eczarny:
More test cases incoming.

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceChangesetVerification.mm
```diff
@@ -146,6 +165,85 @@ CKBadChangesetOperationType CKIsValidChangesetForState(CKTransactionalComponentD
...
+
+static NSArray *emptyMutableArrays(NSUInteger count)
+{
+  NSMutableArray *arrays = [NSMutableArray array];
+  for (NSUInteger i = 0; i < count; i++) {
+    [arrays addObject:[NSMutableArray array]];
+  }
+  return arrays;
+}
+
```
eczarny: Copy/Pasta :(

Likely okay since this is just debugging code but it still feels gross. I'll see if I can simplify this a bit more once I get some more test cases written.

eczarny:
@facebook-github-bot import

PR #671, opened 2016-10-26
Include missing tvOS tests

eczarny:
This tests were missing from the tvOS test target.


ocrickard:
@facebook-github-bot shipit

PR #477, opened 2016-01-28
Implement memoization for the CKTransactionalComponentDataSource

conradev:
This change was pretty straightforward to implement, let me know if you have any feedback!

PR #675, opened 2016-11-03
Implement memoization for the CKTransactionalComponentDataSource

conradev:
This is an updated version of #477 with tests, per https://github.com/facebook/componentkit/pull/477#issuecomment-247075486


conradev:
Would you be able to give this a look @ocrickard?

ocrickard:
Alright, sorry for the delay, this isn't something that I think we should add. I'd like to reduce the interface of memoization in our core. Let me take a bit to do some more investigation, but my guess is we won't merge this in favor of moving up the calls into user-space.

ocrickard:
@conradev can you take a look at #678 and let me know if that would work for you instead? You just wrap your hierarchy at the root with a memoizing component, and it will work in whatever rendering context you use, legacy datasources, transactional datasources, hosting views, etc. so long as the scope tree is properly configured.

PR #676, opened 2016-11-08
Make actions typesafe through templated type

ocrickard:
This one is a doozey, and will break some current code written with ComponentKit, though I've tried my best to make the conversion path simple. Please review thoroughly.

This PR is attempting to make the CKComponentAction declaration a little bit clearer. We're providing a few new facilities:

1. CKComponentActions can now take arbitrary target/selector pairs instead of necessarily using the responder chain.
2. CKComponentAction now has a typed variant which allows specification of the arguments to the action method.
3. An execution block can be created from an action which can simply be executed with the arguments defined in the variadic types. This allows simple interop with systems that expect raw blocks as inputs for their actions (like JSC). (I haven't implemented this yet, tomorrow... should be simple).
4. Component actions may now take an arbitrary number of arguments instead of just 2.
5. Component actions may now take primitives or other non-NSObject types. You can pass things like C++ structs using this system.

The use of actions is currently ad-hoc and only documented in headers. It is a little tough to understand looking at the header for a component what arguments are available for your action method.

On top of that, the status quo uses the responder chain, which was a natural way to build it for the obj-c roots we come from, but ultimately has turned out to be a dangerous and annoying pattern in terms of debuggability. We'd prefer for targets/actions to be captured explicitly, which will allow us to do runtime and/or compile-time checking that the method is implemented at the level that the action was created. I have some changes coming tomorrow to support full integration with component scopes in actions.

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.h
+++ b/ComponentKit/Components/CKButtonComponent.h
@@ -33,7 +33,7 @@ struct CKButtonComponentAccessibilityConfiguration {
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(const CKTypedComponentAction<UIEvent *> &)action
+                       action:(CKTypedComponentAction<UIEvent *>)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)attributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration;
--- a/ComponentKit/Components/CKButtonComponent.h
+++ b/ComponentKit/Components/CKButtonComponent.h
@@ -33,7 +33,7 @@ struct CKButtonComponentAccessibilityConfiguration {
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(const CKComponentAction &)action
+                       action:(const CKTypedComponentAction<UIEvent *> &)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)attributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration;
--- a/ComponentKit/Components/CKButtonComponent.h
+++ b/ComponentKit/Components/CKButtonComponent.h
@@ -33,7 +33,7 @@ struct CKButtonComponentAccessibilityConfiguration {
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(CKComponentAction)action
+                       action:(const CKComponentAction &)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)attributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration;
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -90,7 +90,7 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(const CKComponentAction &)action
+                       action:(const CKTypedComponentAction<UIEvent *> &)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
@@ -161,7 +161,7 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                             std::move(attributes),
                             {
                               .accessibilityLabel = accessibilityConfiguration.accessibilityLabel,
-                              .accessibilityComponentAction = enabled ? action : NULL
+                              .accessibilityComponentAction = enabled ? CKComponentAction(action) : NULL
                             }
                           }
                           size:size];
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -90,7 +90,7 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(CKComponentAction)action
+                       action:(const CKComponentAction &)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -90,7 +90,7 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(const CKTypedComponentAction<UIEvent *> &)action
+                       action:(CKTypedComponentAction<UIEvent *>)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
--- a/ComponentKit/Core/Scope/CKComponentScope.h
+++ b/ComponentKit/Core/Scope/CKComponentScope.h
@@ -65,6 +65,13 @@ public:
   */
   CKComponentStateUpdater stateUpdater(void) const;
 
+  /**
+   @return The scope handle associated with this scope.
+   @discussion This is exposed for use by the framework. You should almost certainly never call this for any reason
+               in your components.
+   */
+  CKComponentScopeHandle *scopeHandle(void) const;
+
 private:
   CKComponentScope(const CKComponentScope&) = delete;
   CKComponentScope &operator=(const CKComponentScope&) = delete;
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.h
@@ -51,12 +51,15 @@
  return nil until `resolve` is called.
  */
 @property (nonatomic, strong, readonly) CKComponentController *controller;
-@property (nonatomic, weak, readonly) CKComponent *acquiredComponent;
-@property (nonatomic, assign, readonly) BOOL resolved;
 
 @property (nonatomic, assign, readonly) Class componentClass;
 
 @property (nonatomic, strong, readonly) id state;
 @property (nonatomic, readonly) CKComponentScopeHandleIdentifier globalIdentifier;
+
+/**
+ Provides a responder corresponding with this scope handle. The controller will assert if called before resolution.
+ */
+- (id)responder;
 
 @end
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.h
@@ -51,6 +51,11 @@
  return nil until `resolve` is called.
  */
 @property (nonatomic, strong, readonly) CKComponentController *controller;
+@property (nonatomic, weak, readonly) CKComponent *acquiredComponent;
+@property (nonatomic, assign, readonly) BOOL resolved;
+
+@property (nonatomic, assign, readonly) Class componentClass;
+
 @property (nonatomic, strong, readonly) id state;
 @property (nonatomic, readonly) CKComponentScopeHandleIdentifier globalIdentifier;
 
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -14,6 +14,7 @@
 #import "CKComponentControllerInternal.h"
 #import "CKComponentScopeRootInternal.h"
 #import "CKComponentSubclass.h"
+#import "CKComponentInternal.h"
 #import "CKInternalHelpers.h"
 #import "CKMutex.h"
 #import "CKThreadLocalComponentScope.h"
@@ -21,15 +22,9 @@
 @implementation CKComponentScopeHandle
 {
   id<CKComponentStateListener> __weak _listener;
-  Class _componentClass;
   CKComponentController *_controller;
   CKComponentScopeRootIdentifier _rootIdentifier;
   BOOL _acquired;
-  // Temporarily stored reference to the specific component that acquired this handle. This forms a reference cycle
-  // that is broken in `resolve`. This reference will always be partially initialized, and should not be used outside
-  // the `resolve` call.
-  CKComponent *_acquiredComponent;
-  BOOL _resolved;
 }
 
 + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
@@ -46,7 +41,7 @@
     }
     return handle;
   }
-  CKCAssertNil(controllerClassForComponentClass([component class]), @"%@ has a controller but no scope! "
+  CKCAssertNil(CKComponentControllerClassFromComponentClass([component class]), @"%@ has a controller but no scope! "
                "Use CKComponentScope scope(self) before constructing the component or CKComponentTestRootScope "
                "at the start of the test.", [component class]);
   return nil;
@@ -166,43 +161,14 @@
     // in that case, and we should avoid re-generating a new controller in that case.
     _controller = newController(_acquiredComponent, currentScope->newScopeRoot);
   }
-  // We break the retain cycle with the acquired component here.
-  _acquiredComponent = nil;
   _resolved = YES;
 }
 
 #pragma mark Controllers
 
-static Class controllerClassForComponentClass(Class componentClass)
-{
-  if (componentClass == [CKComponent class]) {
-    return Nil; // Don't create root CKComponentControllers as it does nothing interesting.
-  }
-
-  static CK::StaticMutex mutex = CK_MUTEX_INITIALIZER; // protects cache
-  CK::StaticMutexLocker l(mutex);
-
-  static std::unordered_map<Class, Class> *cache = new std::unordered_map<Class, Class>();
-  const auto &it = cache->find(componentClass);
-  if (it == cache->end()) {
-    Class c = NSClassFromString([NSStringFromClass(componentClass) stringByAppendingString:@"Controller"]);
-
-    // If you override animationsFromPreviousComponent: or animationsOnInitialMount then we need a controller.
-    if (c == nil &&
-        (CKSubclassOverridesSelector([CKComponent class], componentClass, @selector(animationsFromPreviousComponent:)) ||
-         CKSubclassOverridesSelector([CKComponent class], componentClass, @selector(animationsOnInitialMount)))) {
-      c = [CKComponentController class];
-    }
-
-    cache->insert({componentClass, c});
-    return c;
-  }
-  return it->second;
-}
-
 static CKComponentController *newController(CKComponent *component, CKComponentScopeRoot *root)
 {
-  Class controllerClass = controllerClassForComponentClass([component class]);
+  Class controllerClass = CKComponentControllerClassFromComponentClass([component class]);
   if (controllerClass) {
     CKCAssert([controllerClass isSubclassOfClass:[CKComponentController class]],
               @"%@ must inherit from CKComponentController", controllerClass);
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -25,6 +25,8 @@
   CKComponentController *_controller;
   CKComponentScopeRootIdentifier _rootIdentifier;
   BOOL _acquired;
+  BOOL _resolved;
+  CKComponent *__weak _acquiredComponent;
 }
 
 + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
@@ -162,6 +164,12 @@
     _controller = newController(_acquiredComponent, currentScope->newScopeRoot);
   }
   _resolved = YES;
+}
+
+- (id)responder
+{
+  CKAssert(_resolved, @"Asking for responder from scope handle before resolution:%@", NSStringFromClass(_componentClass));
+  return _acquiredComponent;
 }
 
 #pragma mark Controllers
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -15,13 +15,6 @@
 
 @class CKComponent;
 
-typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
-  /** Starts searching at the sender's next responder. Usually this is what you want to prevent infinite loops. */
-  CKComponentActionSendBehaviorStartAtSenderNextResponder,
-  /** If the sender itself responds to the action, invoke the action on the sender. */
-  CKComponentActionSendBehaviorStartAtSender,
-};
-
 /**
  CKTypedComponentAction is a struct that represents a method invocation that can be passed to a child component to
  trigger a method invocation on a target.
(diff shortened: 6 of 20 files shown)


ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -161,7 +161,7 @@ + (instancetype)newWithTitles:(const std::unordered_map<UIControlState, NSString
                             std::move(attributes),
                             {
                               .accessibilityLabel = accessibilityConfiguration.accessibilityLabel,
-                              .accessibilityComponentAction = enabled ? action : NULL
+                              .accessibilityComponentAction = enabled ? CKComponentAction(action) : NULL
```
ocrickard: In this callsite, we're basically demoting the action, which shows the power of this approach. Before what was happening is we were passing a component action to a system that didn't provide *any* context object. This meant that at runtime, actions sent from this API would silently fail to provide any arguments, leaving you with nil events. This is a little unexpected.

I've provided an explicit constructor to allow conversion of these types to allow for demotion, but it does force you to be aware that it's happening.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,102 @@
...
+template<typename... T>
+struct CKTypedComponentAction {
+  CKTypedComponentAction() : _variant(CKTypedComponentActionVariantRawSelector), _target(nil), _selector(NULL) {};
+  CKTypedComponentAction(id target, SEL selector) : _variant(CKTypedComponentActionVariantTargetSelector), _target(target), _selector(selector)
+  {
+#if DEBUG
+    std::vector<const char *> typeEncodings;
+    _CKTypedComponentActionTypeVectorBuild(typeEncodings, _CKTypedComponentActionTypelist<T...>{});
+    _CKTypedComponentDebugCheckTargetSelector(target, selector, typeEncodings);
+#endif
```
ocrickard: This little function does as much type-checking as we can do at the component action constructor site, which should be easier to understand when it fails on you than later, when actions are sent. I've seen this a few times where actions get out of sync, and we don't find out until the action fires. People sometimes don't test them.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,102 @@
...
+
+  /** 
+   We allow demotion from actions with types to untyped actions, but only when explicit. This means arguments to the
+   method specified here will have nil values at runtime. Used for interoperation with older API's.
+   */
+  template<typename... Ts>
+  explicit CKTypedComponentAction<>(const CKTypedComponentAction<Ts...> &action)
+  { _variant = action._variant; _target = action._target; _selector = action._selector; }
+  
+  /** Allows conversion from NULL actions. */
```
ocrickard: This is just for conversion. Ideally I'd be able to strip these out and you can use {} instead. Only problem with that is that {} can't be used on the right side of a ternary... so it'll inflate the size of the call a bit if we did this.

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -125,7 +131,38 @@ - (instancetype)initWithAction:(CKComponentAction)action
...
+  CKCAssert(typeEncodings.size() + 3 >= signature.numberOfArguments, @"Expected action method %@ to take less than %lu arguments, but it suppoorts %lu", NSStringFromSelector(selector), typeEncodings.size(), (unsigned long)signature.numberOfArguments - 3);
+
+  CKCAssert(signature.methodReturnLength == 0, @"Component action methods should not have any return value. Any objects returned from this method will be leaked.");
+
+  for (int i = 0; i + 3 < signature.numberOfArguments; i++) {
+    const char *methodEncoding = [signature getArgumentTypeAtIndex:i + 3];
+    const char *typeEncoding = typeEncodings[i];
+
+    CKCAssert(strcmp(methodEncoding, typeEncoding) == 0, @"Implementation of %@ does not match expected types.\nExpected type %s, got %s", NSStringFromSelector(selector), typeEncoding, methodEncoding);
+  }
```
ocrickard: This method provides really nice warnings to you if you're going to do something stupid, like pass a method with multiple arguments to an action that only takes 1. We decided to make requirements a bit weak on objc methods. For instance, if your action is defined:

```
CKTypedComponentAction<id, id, id>
```

You could pass in a method that took only one:

```
[SomeComponent
newWithAction:{target, @selector(sender:context1:)}]
```

This will invoke the selector with the sender, and the first param present. This is built to mirror the behavior of performSelector:withObject: which we used before this diff.

ComponentKit/Utilities/CKComponentGestureActions.h
```diff
@@ -17,44 +17,44 @@ typedef void (*CKComponentGestureRecognizerSetupFunction)(UIGestureRecognizer *)
...
 /**
  Returns a view attribute that creates and configures a long press gesture recognizer to send the given CKComponentAction.
-
+ 
  @param action Sent up the responder chain when a long press occurs. Sender is the component that created the view.
-        Context is the gesture recognizer. May be NULL, in which case no action will be sent.
+ Context is the gesture recognizer. May be NULL, in which case no action will be sent.
  */
-CKComponentViewAttributeValue CKComponentLongPressGestureAttribute(CKComponentAction action);
+CKComponentViewAttributeValue CKComponentLongPressGestureAttribute(const CKTypedComponentAction<UIGestureRecognizer *> &action);
```
ocrickard: Sadly the C++ templating system is unaware of obj-c's loose typing system, so it requires the exact type in any specializations of the template.

ocrickard:
The gesture action tests are failing, I touched that map last, probably unrelated to the rest of the diff since everything else passed before.

ComponentKit/Utilities/CKComponentGestureActions.mm
```diff
@@ -113,26 +113,29 @@ CKComponentViewAttributeValue CKComponentGestureAttribute(Class gestureRecognize
...
   static auto *reusePoolMap = new std::unordered_map<CKGestureRecognizerReusePoolMapKey, CKGestureRecognizerReusePool *>();
   static CK::StaticMutex reusePoolMapMutex = CK_MUTEX_INITIALIZER;
   CK::StaticMutexLocker l(reusePoolMapMutex);
   auto &reusePool = (*reusePoolMap)[{gestureRecognizerClass, setupFunction}];
   if (reusePool == nullptr) {
     reusePool = new CKGestureRecognizerReusePool(gestureRecognizerClass, setupFunction);
   }
+  // We're capturing the C++ struct in a block. We use the __block storage class which will call the copy constructor
+  // on our struct when the block is copied to the heap.
+  __block CKTypedComponentAction<UIGestureRecognizer *> blockAction = action;
```
ocrickard: This is a bummer with the conversion to C++ struct. We don't get implicit block capture anymore, and I don't know if there's a way to ban the implicit capture that will result in reading garbage memory. Finding these bugs will be difficult in the main app, but we probably have to do it...

ocrickard:
@facebook-github-bot import

ComponentKit/Components/CKButtonComponent.h
```diff
@@ -33,7 +33,7 @@ struct CKButtonComponentAccessibilityConfiguration {
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
-                       action:(CKComponentAction)action
+                       action:(const CKTypedComponentAction<UIEvent *> &)action
```
adamjernst: Technically, passing this around by reference might be overkill. For very small structs like `CKTypedComponentAction`, it is probably almost as cheap to just copy everywhere instead of passing references.

It's your call — I would decide on a standard (always pass by reference, or always pass by value) and then let's enforce it, both for ComponentKit and internally at FB.

ComponentKit/Components/CKButtonComponent.mm
adamjernst: Cool.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -51,6 +51,11 @@
  return nil until `resolve` is called.
  */
 @property (nonatomic, strong, readonly) CKComponentController *controller;
+@property (nonatomic, weak, readonly) CKComponent *acquiredComponent;
+@property (nonatomic, assign, readonly) BOOL resolved;
+
+@property (nonatomic, assign, readonly) Class componentClass;
```
adamjernst: exposing all these individual pieces of data publicly makes me feel pretty bad — but I don't have a good enough grasp to suggest an alternative. let's land it, and i can always try to poke around and see what i could change later…

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -14,22 +14,17 @@
...
 @implementation CKComponentScopeHandle
 {
   id<CKComponentStateListener> __weak _listener;
-  Class _componentClass;
   CKComponentController *_controller;
   CKComponentScopeRootIdentifier _rootIdentifier;
   BOOL _acquired;
-  // Temporarily stored reference to the specific component that acquired this handle. This forms a reference cycle
-  // that is broken in `resolve`. This reference will always be partially initialized, and should not be used outside
-  // the `resolve` call.
```
adamjernst: where'd this comment go?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -46,7 +41,7 @@ + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
     }
     return handle;
   }
-  CKCAssertNil(controllerClassForComponentClass([component class]), @"%@ has a controller but no scope! "
+  CKCAssertNil(CKComponentControllerClassFromComponentClass([component class]), @"%@ has a controller but no scope! "
```
adamjernst: why must this function now be public?

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -30,25 +23,184 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+            respond to the selector. Targets are captured weakly by the action. Promotion, as in option 2 above is also
+            supported for target/selector actions. This constructor is useful for triggering actions on objects outside
+            of the component hierarchy like view controllers.
+ 
+ [MyComponent newWithAction:{[SomeObject sharedInstance], @selector(methodWithNoArguments)}];
+ ...
+ on SomeObject: - (void)methodWithNoArguments {}
+ 
+
+ Option 4 - Scope action. Similar to target/selector action in that it skips the responder chain from the sender, and
```
adamjernst: so this should be option 1

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -30,25 +23,184 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+            accept fewer arguments than defined in the declaration of the action above. However, types of received
+            parameters should be the same as the declaration, if they're present.
+
+ [MyComponent newWithAction:{@selector(methodWithNoArguments)}];
+ ...
+ - (void)methodWithNoArguments {}
+
+ 
+ Option 2 - Exact raw-selector component action. Uses a raw selector which also traverses upwards using the responder
+            chain, but accepts the arguments defined in the interface.
```
adamjernst: aren't option 1 and option 2 the same? the rule is "you can supply up to exactly as many parameters as the action takes, or any small number of arguments." or is it truly all-or-nothing—you must take all arguments or no arguments?

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -30,25 +23,184 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+    __weak id weakSender = sender;
+    CKTypedComponentAction<T...> copy {*this};
+    return ^(T... args) {
+      id strongSender = weakSender;
+      CKCAssert(strongSender, @"Curried sender should be retained by caller.");
+      copy.send(strongSender, behavior, args...);
+    };
+  }
+
+  /** Private details, do not use. */
```
adamjernst: well then stick `private:` right above it, right?

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,102 @@
...
+ 
  @param action The action to send up the responder chain.
  @param sender The component sending the action. Traversal starts from the component itself, then its next responder.
- @param context An optional context-dependent second parameter to the component action. Defaults to nil.
+ @param context An optional context-dependent second parameter to the component action.
  @param behavior @see CKComponentActionSendBehavior
  */
-void CKComponentActionSend(CKComponentAction action, CKComponent *sender, id context = nil,
-                           CKComponentActionSendBehavior behavior = CKComponentActionSendBehaviorStartAtSenderNextResponder);
+void CKComponentActionSend(const CKComponentAction &action, CKComponent *sender);
```
adamjernst: (y)

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -0,0 +1,87 @@
...
+  invocation.selector = selector;
+  invocation.target = responder;
+  if (signature.numberOfArguments >= 3) {
+    [invocation setArgument:&sender atIndex:2];
+  }
+  // We use a recursive argument unpack to unwrap the variadic arguments in-order on the invocation in a type-safe
+  // manner.
+  _CKConfigureInvocationWithArguments(invocation, 3, args...);
+  [invocation invoke];
+}
```
adamjernst: yup, templates work that way. @swolchok  might be able to help, as this could have a pretty nasty impact on code size otherwise…

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -30,25 +23,184 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+ 
+ Usage in your component header:
+ 
+ @interface MyComponent : CKComponent
+ + (instancetype)newWithAction:(const CKTypedComponentAction<NSString *, int> &)action;
+ @end
+ 
+ When creating the action:
+ 
+ Option 1 - Promoted raw-selector component action. Uses a raw selector which traverses upwards looking for a parent
```
ocrickard: lol, k good point.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
ocrickard: Because we call it now in more than one context. We use it to verify the action is either implemented on the component or its controller.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
ocrickard: We now just take a weak reference, and we don't remove that reference, so the comment no longer made sense. Removed.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
ocrickard: Yep, we can keep it internal and call methods on it instead.

ComponentKit/Components/CKButtonComponent.h
ocrickard: Interesting. Passing by value allows correct capture in blocks. Perhaps we should just use that as the standard.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -30,25 +23,184 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+#endif
+  }
+  
+  /** Legacy constructor for raw selector actions. Traverse up the mount responder chain. */
+  CKTypedComponentAction(SEL selector) : _variant(_CKTypedComponentActionVariantRawSelector), _target(nil), _selector(selector) { };
+
+  /** We support promotion from actions that take no arguments. */
+  template <typename T1, typename... Ts>
+  CKTypedComponentAction<T1, Ts...>(const CKTypedComponentAction<> &action)
+  { _variant = action._variant; _target = action._target; _selector = action._selector; }
```
ocrickard: Yeah, fair enough.

ComponentKit/Utilities/CKComponentAction.h
ocrickard: That means we can't use it in the copy constructors above.

swolchok:
I can't reply directly to the comment that summoned me because the commit has been force-pushed to and it no longer exists.

I recommend splitting `CKTypedComponentAction` into two pieces: the templated part, containing the code that actually needs the template arguments (AFAICT, this is only the DEBUG-gated type checking code?) and living in the header, and a plain old struct that contains all the boring bits that implement the actual functionality of the class and lives in an implementation file (except for pieces that are guaranteed to be really really cheap, if any; touching Obj-C references does not fall into that category). The templated bit can re-use the implementation bit by aggregation if possible and inheritance if necessary (or manifestly more convenient because of C++11 `using`).

ocrickard:
Sweet, these are great suggestions. Let me give that a try.

ocrickard:
The templated parts are core to the act of sending actions, so there wasn't a ton I could move out of the header easily. I'm a n00b with this though, so happy to learn if I'm missing something obvious.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,177 @@
...
+  {
+    __weak id weakSender = sender;
+    CKTypedComponentAction<T...> copy {*this};
+    return ^(T... args) {
+      id strongSender = weakSender;
+      copy.send(strongSender, behavior, args...);
+    };
+  }
+
+  CKTypedComponentActionValue _internal;
```
ocrickard: I couldn't make the values private without explicit specializations/friends, so I opted for this guy, which prevents raw access to the members.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -21,50 +21,74 @@
...
+
+typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
+  /** Starts searching at the sender's next responder. Usually this is what you want to prevent infinite loops. */
+  CKComponentActionSendBehaviorStartAtSenderNextResponder,
+  /** If the sender itself responds to the action, invoke the action on the sender. */
+  CKComponentActionSendBehaviorStartAtSender,
+};
+
+struct CKTypedComponentActionValue {
+  CKTypedComponentActionValue() : _variant(CKTypedComponentActionVariantRawSelector), _target(nil), _scopeHandle(nil), _selector(NULL) {}
```
swolchok: I might consider outlining this if it's going to show up in lots of end-user code; it's at least 4 instructions, whereas the function call should take 2. Can always wait on that though.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -21,50 +21,74 @@
...
+  CKTypedComponentActionValue(const CKTypedComponentActionValue &value) : _variant(value._variant), _target(value._target), _scopeHandle(value._scopeHandle), _selector(value._selector) {};
```
swolchok: more inclined to outline this one than the default ctor. In addition to the 4 stores, it's got `objc_storeWeak` calls.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -21,50 +21,74 @@
...
+  CKTypedComponentActionValue(CKTypedComponentActionVariant variant, __unsafe_unretained id target, __unsafe_unretained CKComponentScopeHandle *scopeHandle, SEL selector) : _variant(variant), _target(target), _scopeHandle(scopeHandle), _selector(selector) {};
```
swolchok: ditto

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -21,50 +21,74 @@
...
 
 /** 
  Recursion through variadic argument type unpacking. This allows us to build a vector of encoded const char * before
  any actual arguments have been provided. All of this is done at compile-time.
  */
 template<typename T, typename... Ts>
-void _CKTypedComponentActionTypeVectorBuild(std::vector<const char *> &typeVector, const _CKTypedComponentActionTypelist<T, Ts...> &list)
+void CKTypedComponentActionTypeVectorBuild(std::vector<const char *> &typeVector, const CKTypedComponentActionTypelist<T, Ts...> &list)
 {
   typeVector.push_back(@encode(T));
```
swolchok: it might be nice to DEBUG-gate this implementation, just in case.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -21,50 +21,74 @@
...
  require the intermediary struct.
  */
 template <typename T, typename... Ts>
-void _CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger index, T t, Ts... args)
+void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger index, T t, Ts... args)
 {
   // We have to be able to handle methods that take less than the provided number of arguments, since that will cause
   // an exception to be thrown.
   if (index < invocation.methodSignature.numberOfArguments) {
     [invocation setArgument:&t atIndex:index];
```
swolchok: i'm a little confused as to how this is safe, but I think you mention that it's not safe in comments?

ComponentKit/Utilities/CKComponentActionInternal.h
ocrickard: sgtm

ComponentKit/Utilities/CKComponentActionInternal.h
ocrickard: Yeah, so this works fine for plain data structs, primitives, and pointers, but more complex data types just silently fail here. If I get an IMP and cast it to the required type then I'm pretty sure it would work with complex types, but then I have other problems to deal with such as zeroing parameters that aren't provided in the send call (which we actually do, it's terrible). Suggestions? I considered just restricting types to NSObjects which wouldn't be that hard, but would still force us to use all these value wrappers everywhere we have right now.

ocrickard: Thinking about this more, I'm going to try and forbid complex types and just provide some specializations for normal types like passing NSUInteger.

ocrickard: After a bit of exploration, I added a static_assert that you can't provide values that are not trivially constructible, or pointer types. I believe these should be safe to pass via an invocation.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ocrickard:
ping @adamjernst I see you accepted this internally, are we good to go here? Can I merge?

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,180 @@
...
 @class CKComponent;
 
 /**
- A component action is simply a selector. The selector may optionally take one argument: the sending component.
+ CKTypedComponentAction is a struct that represents a method invocation that can be passed to a child component to
+ trigger a method invocation on a target.
+ 
+ We allow a typed specification of the parameters that will be provided as arguments to the component action at runtime
+ through the variadic templated arguments. You may specify an arbitrary number of arguments to your component action,
+ and you may use either object, or primitive arguments. Only trivially-constructible arguments or pointers can be used
```
natansh: I wonder whether people understand "trivially-constructible". It's a C++ oriented (or maybe even specific) term. I didn't know about this, but it made sense once I read the docs.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,180 @@
...
+ Usage in your component header:
+ 
+ @interface MyComponent : CKComponent
+ + (instancetype)newWithAction:(CKTypedComponentAction<NSString *, int>)action;
+ @end
+ 
+ When creating the action:
+ 
+ Option 1 - Scope action. Similar to target/selector action in that it skips the responder chain from the sender, and
+            directly invokes the selector on the component or controller corresponding with the scope. Promotion is also
```
natansh: What do you mean by promotion, exactly?

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,180 @@
...
+            supported for scope-based actions. Scope actions weakly capture the component or controller. Does not
```
natansh: nit: `s/Does not/They do not/` or `s/Scope actions weakly capture/A scope action weakly captures/`

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -11,44 +11,180 @@
...
+             {
+               CKComponentScope scope(self);
+               return [super
+                       newWithComponent:
+                        [MyComponent
+                         newWithAction:{scope, @selector(methodWithSender:)}]];
+             }
+             - (void)methodWithSender:(CKComponent *)sender {}
+ 
+ Option 2 - Target/selector action. Ensures that the target responds to the given selector. Target must directly
```
natansh: This might be needed, but I think the first option seems the best and the pattern that should be encouraged the most - it tends to ensure that actions are handled in the component hierarchy whenever possible as well. Would we want to discourage Option 2, or at least suggest the usage of Option 1 over this one?

ComponentKit/Core/Scope/CKComponentScope.h
```diff
@@ -65,6 +65,13 @@ class CKComponentScope {
   */
   CKComponentStateUpdater stateUpdater(void) const;
 
+  /**
+   @return The scope handle associated with this scope.
+   @discussion This is exposed for use by the framework. You should almost certainly never call this for any reason
+               in your components.
+   */
+  CKComponentScopeHandle *scopeHandle(void) const;
+
```
ocrickard: How? C++ doesn't support this I think without a friend type, and I don't like the indirection.

ComponentKit/Utilities/CKComponentAction.h
ocrickard: Yeah, fair point.

ComponentKit/Utilities/CKComponentAction.h
ocrickard: Agreed, prefer (1). This is for interop with things like owning view controllers. People do use the responder chain for this right now, and I don't like it.

natansh: This might be a bad idea/not thought out fully but - If you think this option is only required interop with things like owning view controllers, how about we ensure that it directly goes up to the root component scope (I guess around the root component view too) and then follows the responder chain upwards to get to the relevant view controller? This will force the usage of option 1 for actions that can be handled in components themselves.

Reiterating again that it might be a bad idea. :-)

ocrickard: Hmm, yeah, I mean I'm trying to kill the responder chain in general. Actions shouldn't be dependent on any type of mounting/context I think?

Anyway, great idea, let's see how usage of this goes, and if it's problematic, reconsider.

ComponentKit/Utilities/CKComponentAction.h
ocrickard: Yeah, I tried to make this clear, but what I mean by promotion is providing an action that receives less arguments than the declaration. So for instance, if my action is declared as:

```
- (instancetype)newWithAction:(CKTypedComponentAction<id, NSString *, int>)action
```

Then a *promotion* would be passing a typed action that receives less than the number of arguments in the declaration of the type.

So, these would all be promotions:

`[MyComponent newWithAction:CKTypedComponentAction<>(scope, @selector(mySelector))]`
`[MyComponent newWithAction:CKTypedComponentAction<id>(scope, @selector(mySelectorWithSender:firstArgument:))]`
`[MyComponent newWithAction:CKTypedComponentAction<id, NSString *>(scope, @selector(mySelectorWithSender:firstArgument:stringArgument:))]`

Any of the above would be implicitly converted to the type declared in the component header that receives the action.

**demotions** on the other hand, go from an action type that receives N arguments to a type that provides less than N arguments.  So for instance given this interface:

```
- (instancetype)newWithAction:(CKTypedComponentAction<id>)action
```

All of the following would be demotions:

`[MyComponent newWithAction:CKTypedComponentAction<id, NSString *>(scope, @selector(mySelectorWithSender:firstArgument:stringArgument))]`
`[MyComponent newWithAction:CKTypedComponentAction<id, int>(scope, @selector(mySelectorWithSender:firstArgument:intArgument))]`

These will fail to compile, since we're slicing off one of the arguments that the action expects. You can force a demotion through an explicit conversion constructor:

`[MyComponent newWithAction:CKTypedComponentAction<id>(CKTypedComponentAction<id, int>(scope, @selector(mySelectorWithSender:firstArgument:intArgument)))]`

But this should rarely be used. You can see above where I've had to do this for a button action that is being passed to the accessibility system. The accessibility API takes a CKTypedComponent<>, but in the button API, we've provided an action of type CKTypedComponent<UIEvent *>.

Anyway, in practice, the action is called, as it is currently with performSelector:withObject, except we use NSInvocation which will provide nil (or zeroed) arguments for any unspecified parameters, up to a certain size.

`[MyComponent newWithAction:CKTypedComponentAction<id>(scope, @selector(mySelectorWithSender:firstArgument:))]`
`[MyComponent newWithAction:CKTypedComponentAction<id, NSString *>(scope, @selector(mySelectorWithSender:firstArgument:stringArgument:))]`

ocrickard:
This was accepted internally. Trying to land.

ocrickard:
Merged internally.

PR #678, opened 2016-11-10
Lift memoization out of core

ocrickard:
My goal here is to reduce the need for customizations to the build and layout calls in every rendering context that wants memoization. Here's an example of what I'm trying to avoid:

https://github.com/facebook/componentkit/pull/675

If your component hierarchy needs memoization, it should work regardless of how you configure your components, and ideally we wouldn't have to do this type of change to every place that builds components.

In this diff I introduce CKMemoizingComponent which prepares the thread-local state for component and layout memoization for its children. This component stores its memoization state into its scope's state.

Due to the nature of the thread-local stack that memoizers are built with, it should be fine to have multiple memoization components in the hierarchy with little cost, and everything "just works". This is important, since if you embed a sub-hierarchy that adds one of these so it can memoize in a different context, you don't want it to conflict with a higher-level memoizing component wrapper.

I'd like to land this first, then look at fixing up the scopes/states as in https://github.com/facebook/componentkit/pull/348 after this is in master. Looking at that diff, we can apply basically the same process here.

### DIFF
--- a/ComponentKit/Core/CKComponentMemoizer.h
+++ b/ComponentKit/Core/CKComponentMemoizer.h
@@ -22,6 +22,8 @@ struct CKMemoizationKey;
 
 id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void));
 
+CKComponentLayout CKMemoizeLayout(CKComponent *component, CKSizeRange constrainedSize, const CKComponentSize& size, CGSize parentSize, CKComponentLayout (^block)());
+
 /**
  
  How to use the component memoization:
@@ -34,10 +36,13 @@ id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void));
             [CKStackLayoutComponent
               newWith...
    });
- 
  }
  
  MyKey must consist only of objects that define -hash and -isEqual:.
+ 
+ You should have a CKMemoizingComponent at the root of your component hierarchy above the component you wish to memoize,
+ or the memoization call will fail, and new versions of your components and layouts will be created every time they are
+ requested.
 
  If you're building something that calls CKMountComponentLayout() directly,
  just add a CKComponentMemoizer in the right scope:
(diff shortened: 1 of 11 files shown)


ocrickard:
@facebook-github-bot import

ocrickard:
@darknoon feel free to take a look!

conradev:
This looks great! Good to see it being moved out of the core.

ComponentKit/Core/CKComponentMemoizer.h
```diff
@@ -56,15 +61,27 @@ id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void));
...
+ - (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize 
+                           restrictedToSize:(const CKComponentSize &)size
+                       relativeToParentSize:(CGSize)parentSize
+ {
+   return CKMemoizeLayout(self, constrainedSize, size, parentSize, ^CKComponentLayout{
+     return [super computeLayoutThatFits:constrainedSize
+                        restrictedToSize:size
+                    relativeToParentSize:parentSize];
+   });
+ }
```
eczarny: It may be a good idea to add these comments to the `CKMemoizeLayout` function itself.

How much of this header is actually part of the public-facing API? I ask because there is a fair amount of stuff here and we should take this opportunity to clean this up a bit if we want to.

ComponentKit/Core/CKComponentMemoizer.mm
```diff
@@ -164,14 +171,17 @@ id CKMemoize(CKMemoizationKey memoizationKey, id (^block)(void))
   return impl ? impl->_next : nil;
 }
 
-CKComponentLayout CKMemoizeOrComputeLayout(CKComponent *component, CKSizeRange constrainedSize, const CKComponentSize& size, CGSize parentSize)
+CKComponentLayout CKMemoizeLayout(CKComponent *component, CKSizeRange constrainedSize, const CKComponentSize& size, CGSize parentSize, CKComponentLayout (^block)())
 {
-  if (component && [component shouldMemoizeLayout]) {
```
eczarny: This checks if the component is non-nil. Is that something we still need to guarantee?

eczarny:
@ocrickard back to you for my feedback and resolving merge conflicts.

ComponentKit/Core/CKComponentMemoizer.h
ocrickard: Hmm, I'd like to handle hiding this stuff later. There are a few functions we can move out-of-line, but most of it has to be public.

ocrickard:
@facebook-github-bot import

ocrickard:
Alright, this is landing internally. Travis seems super backed up, but I'll wait for a green build here before merging.

PR #679, opened 2016-11-13
Try to reduce some of the code in the headers of component actions

ocrickard:
Pulling in the templated struct action, we saw an inflation of the binary. Let's try moving some stuff our of the headers to try and reduce some of the bloat.

The curried block function can be implemented fully by users, doesn't need to be present on every template instantiation. Also, move invocation creation and initial configuration to an extern function.


ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ocrickard:
Accepted internally and landed.

PR #680, opened 2016-11-13
Fix build for certain device and dylib builds

ocrickard:
This already landed internally to fix dylib builds for certain device configurations. I'm unclear on why certain architectures require this explicit destructor.


ocrickard:
This was accepted and landed internally. Merging here.

PR #681, opened 2016-11-14
Fix forwarded selectors not respected

ocrickard:
According to https://www.mikeash.com/pyblog/friday-qa-2009-03-27-objective-c-message-forwarding.html we missed a step in doing our invocation here. The performSelector:withObject: we were using before looks for a fast-forwarding path if the object doesn't respond to the selector.

We have to make sure that once we get to the responder in the chain that wants the action, we give it the opportunity to forward that selector.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -70,9 +70,7 @@ NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id t
   while (!signature) {
     // From https://www.mikeash.com/pyblog/friday-qa-2009-03-27-objective-c-message-forwarding.html
     // 1. Lazy method resolution
-    BOOL resolved = [[responder class] resolveInstanceMethod:selector];
-    
-    if (resolved) {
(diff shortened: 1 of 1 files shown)


ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -67,6 +67,27 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
                   NSStringFromSelector(selector), _CKComponentResponderChainDebugResponderChain(target));
   // This is not performance-sensitive, so we can just use an invocation here.
   NSMethodSignature *signature = [responder methodSignatureForSelector:selector];
+  while (!signature) {
+    // From https://www.mikeash.com/pyblog/friday-qa-2009-03-27-objective-c-message-forwarding.html
+    // 1. Lazy method resolution
+    [[responder class] resolveInstanceMethod:selector];
```
eczarny: Do we care about the return value here or are we just relying on getting a signature immediately below? Would being explicit be better?

ocrickard:
@facebook-github-bot import

ocrickard:
Landing.

PR #682, opened 2016-11-14
Use a base-class for CKTypedComponentAction to reduce size

ocrickard:
I met with @craffert0 this morning to go over strategies to reduce the size of the component action struct because we caused a lot of template bloat due to the tens of thousands of callsites we inflated with the initial templated struct approach. We move the bulk of the implementations to a non-templated base class, and moved the internal struct to be protected, so no more direct access to the value struct! Yay!

We also moved for an explicit external template instantiation for <> and <id> since these are the most common uses, and we can force the compiler to use the out-of-line definition instead of inlining the method bodies. This should make all the callsites just call the same functions in the binary instead of spreading our implementations all over the place.

This shouldn't change any client code. Interface remains identical.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -83,17 +83,16 @@
  {
    _action.send(self, @"hello", 4);
  }
-
  */
 template<typename... T>
-struct CKTypedComponentAction {
+class CKTypedComponentAction : public CKTypedComponentActionBase {
   static_assert(std::is_same<
                 CKTypedComponentActionBoolPack<(std::is_trivially_constructible<T>::value || std::is_pointer<T>::value)...>,
                 CKTypedComponentActionBoolPack<(CKTypedComponentActionDenyType<T>::value)...>
                 >::value, "You must either use a pointer (like an NSObject) or a trivially constructible type. Complex types are not allowed as arguments of component actions.");
-
-  CKTypedComponentAction<T...>() : _internal({}) {};
-  CKTypedComponentAction<T...>(id target, SEL selector) : _internal({CKTypedComponentActionVariantTargetSelector, target, nil, selector})
+public:
+  CKTypedComponentAction<T...>() : CKTypedComponentActionBase() {};
+  CKTypedComponentAction<T...>(id target, SEL selector) : CKTypedComponentActionBase(target, selector)
   {
 #if DEBUG
     std::vector<const char *> typeEncodings;
@@ -101,8 +100,8 @@ struct CKTypedComponentAction {
     _CKTypedComponentDebugCheckTargetSelector(target, selector, typeEncodings);
 #endif
   }
-
-  CKTypedComponentAction<T...>(const CKComponentScope &scope, SEL selector) : _internal({CKTypedComponentActionVariantComponentScope, nil, scope.scopeHandle(), selector})
+  
+  CKTypedComponentAction<T...>(const CKComponentScope &scope, SEL selector) : CKTypedComponentActionBase(scope, selector)
   {
 #if DEBUG
     std::vector<const char *> typeEncodings;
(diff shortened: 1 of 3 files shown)


ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -145,12 +139,14 @@ struct CKTypedComponentAction {
...
   }
-
-  CKTypedComponentActionValue _internal;
 };
 
 typedef CKTypedComponentAction<> CKComponentAction;
 
+/** Explicit instantiation of our most commonly-used templates to avoid bloat in callsites. */
+extern template class CKTypedComponentAction<>;
+extern template class CKTypedComponentAction<id>;
```
ocrickard: Here's the bit that forces the extern instantiation of our common action templates. We may want to add a few more here like UIEvent * and NSString * or whatever else people use. For now, this should do for the immediate conversion.

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -249,3 +249,8 @@ void _CKTypedComponentDebugCheckTargetSelector(id target, SEL selector, const st
 }
 
 @end
+
+#pragma mark - Template instantiations
+
+template class CKTypedComponentAction<>;
+template class CKTypedComponentAction<id>;
```
ocrickard: we declared extern in the header, we do have to actually do the instantiation somewhere...

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -55,6 +55,34 @@ struct CKTypedComponentActionValue {
...
+  CKTypedComponentActionBase(SEL selector) : _internal(CKTypedComponentActionVariantRawSelector, nil, nil, selector) {};
+  
+  /** Allows conversion from NULL actions. */
+  CKTypedComponentActionBase(int s) : _internal({}) {};
+  CKTypedComponentActionBase(long s) : _internal({}) {};
+  CKTypedComponentActionBase(std::nullptr_t n) : _internal({}) {};
+  
+  ~CKTypedComponentActionBase() {};
+  
+  CKTypedComponentActionValue _internal;
```
ocrickard: Yay, only the base class can access _internal now.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -83,26 +83,25 @@
...
 #if DEBUG
     std::vector<const char *> typeEncodings;
     CKTypedComponentActionTypeVectorBuild(typeEncodings, CKTypedComponentActionTypelist<T...>{});
     _CKTypedComponentDebugCheckTargetSelector(target, selector, typeEncodings);
 #endif
   }
-
-  CKTypedComponentAction<T...>(const CKComponentScope &scope, SEL selector) : _internal({CKTypedComponentActionVariantComponentScope, nil, scope.scopeHandle(), selector})
+  
+  CKTypedComponentAction<T...>(const CKComponentScope &scope, SEL selector) : CKTypedComponentActionBase(scope, selector)
```
ocrickard: @craffert0 I tried to make the change to do the whole
```
using CKTypedComponentActionBase::CKTypedComponentActionBase;
```
when not in debug mode, but it fails to compile, saying that callsites are accessing protected constructors of the base class. Any idea how to fix?

ocrickard:
@facebook-github-bot import

ComponentKit/Utilities/CKComponentAction.h
craffert0: I think you just need to put that in the `public` section of `CKTypedComponentAction`.  But I could be wrong.  In which case this is reasonable, and the compiler should be smart about it.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -111,31 +110,26 @@ struct CKTypedComponentAction {
...
-  CKTypedComponentAction(std::nullptr_t n) : _internal({}) {};
-
-  ~CKTypedComponentAction() {};
-
-  explicit operator bool() const { return bool(_internal); };
-  bool operator==(const CKTypedComponentAction& rhs) const { return _internal == rhs._internal; }
+  explicit CKTypedComponentAction<>(const CKTypedComponentAction<Ts...> &action) : CKTypedComponentActionBase(action) { };
   
-  SEL selector() const { return _internal.selector(); };
+  ~CKTypedComponentAction() {};
```
craffert0: I don't believe you need the destructor, since it's just default.

ocrickard:
@eczarny  or @craffert0 mind shipping this with the facebook bot?

craffert0:
@facebook-github-bot shipit

eczarny:
@facebook-github-bot shipit

ocrickard:
Hmm, build failures appear unrelated.

ocrickard:
Ugh, we have some sort of flaky build. Try again.

ocrickard:
Alright, this flakiness is annoying. This merged internally, and we're green for the build.

PR #683, opened 2016-11-18
Reduce type bloat by removing sub-struct

ocrickard:
In inspecting the type encodings of these classes inside the compiled binaries I noted that we're seeing both the CKTypedComponentAction<...> signature AND its sub-struct the CKTypedComponentActionValue struct in the type encoded strings when stored as ivars or passed to obj-c methods. This indirection was made unnecessary by the move to a non-templated base-class, so let's move all this to directly use members of the base-class.

This should change no client code, but should reduce binary bloat a little bit.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -94,7 +94,7 @@ NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id t
       // The responder resolved its instance method, we now have a valid responder/signature
       break;
     }
-    
+
     // 2. Fast-forwarding path
     id forwardingTarget = [responder forwardingTargetForSelector:selector];
     if (!forwardingTarget || forwardingTarget == responder) {
@@ -102,7 +102,7 @@ NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id t
       CKCFailAssert(@"Forwarding target failed for action:%@ %@", target, NSStringFromSelector(selector));
       return nil;
     }
-    
+
     responder = forwardingTarget;
     signature = [responder methodSignatureForSelector:selector];
   }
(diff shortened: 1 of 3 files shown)


ocrickard:
@facebook-github-bot import

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -36,46 +36,36 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
-  __weak CKComponentScopeHandle *_scopeHandle;
-  SEL _selector;
-};
-
 #pragma mark - Action Base
 
 /** A base-class for typed components that doesn't use templates to avoid template bloat. */
 class CKTypedComponentActionBase {
 protected:
-  CKTypedComponentActionBase() = default;
```
eczarny: Is explicitly specifying the default constructor no longer necessary?

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -95,15 +94,15 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
       // The responder resolved its instance method, we now have a valid responder/signature
       break;
     }
-
+
```
eczarny: Do we need the additional whitespace?

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -36,46 +36,36 @@ typedef NS_ENUM(NSUInteger, CKComponentActionSendBehavior) {
...
+  
+  id initialTarget(CKComponent *sender) const;
+  CKComponentActionSendBehavior defaultBehavior() const;
+  
+  bool operator==(const CKTypedComponentActionBase& rhs) const;
+  
+  CKTypedComponentActionVariant _variant;
+  __weak id _target;
+  __weak CKComponentScopeHandle *_scopeHandle;
+  SEL _selector;
```
eczarny: I'm still not quite sure how this helps reduce the size of the binary but this feels a bit simpler than before.

ocrickard:
Alright, I was interested in where the size reductions I observed came from, and as I suspected, the type encoding system has to encode the name of any sub-structs present in the structs:

Before:

```
(const char *) encodedSig = 0x00007fbedef1c545 "{CKTypedComponentAction<UIEvent *>=Q@@:{CKTypedComponentActionValue=}}"
```

After:

```
(const char *) encodedSig = 0x00007fc971222b75 "{CKTypedComponentAction<UIEvent *>=Q@@:}"
```

PR #684, opened 2016-11-18
Add call operator

ocrickard:
Based on https://github.com/facebook/componentkit/pull/683

I think it makes sense to add a call operator for component actions instead of using the send() function. This makes them feel like functions...

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -132,12 +132,12 @@ public:
   ~CKTypedComponentAction() {};
 
   void send(CKComponent *sender, T... args) const
-  { this->send(sender, _internal.defaultBehavior(), args...); };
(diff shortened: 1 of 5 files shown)


ocrickard:
@facebook-github-bot import

adamjernst:
whoa! cool

eczarny:
Totally on board with this. Looks like a great way to clean up the API a bit.

conradev:
Thoughts on making `CKComponentAction(NULL)()` a no-op? Because actions are optional for all of our components, we have a lot of code that looks like this:

```objc++
if (_action)
    _action.call(self/*, stuff */);
```

Excited to see this land!

ocrickard:
OK, I think I'm going to drop this one. We can revive it later, should be a quick codemod, but I think there are some good arguments to not adding a call operator.

PR #685, opened 2016-11-18
Revert "Merge pull request #469 from DeskConnect/transactional-bounds…

ocrickard:
…-animation"

This reverts https://github.com/facebook/componentkit/pull/469 because it breaks animations in certain conditions. @eczarny is going to get to the bottom of why, and re-commit this, but for now we need to back it out.

This reverts commit e587533db470fef3884a371fb08b35ccb1a6c382, reversing
changes made to e61515a2ea3dea19c154fe86d7718123715e6d09.

PR #469, opened 2016-01-21
Add CKComponentBoundsAnimation to CKCollectionViewTransactionalDataSource

conradev:
In `CKCollectionViewDataSource`, synchronous cell updates do not trigger a collection view reload and instead mount the new component layout in place and animate any size changes. I made this the default behavior for both asynchronous and synchronous updates to `CKCollectionViewTransactionalDataSource`, where the content of the changeset is only cell updates.

Calling `-[UICollectionView reloadItemsAtIndexPaths:]` causes the component to be unmounted and remounted. This is fine in some cases, but it has adverse side effects. It causes the cell to flash to its new state without animation and it also disrupts the first responder.

PR #686, opened 2016-11-21
Reintroduce bounds animation changes from #469

eczarny:
#469 was originally reverted by #685 to resolve an internal bounds animation regression. Now that I've spent a bit of time narrowing down the issue we can look into reintroducing these bounds animation changes.

These changes can be merged as-is since they don't introduce the regression. A bad merge internally is to blame and we can handle that fix independently of the OSS contribution.

/cc @conradev


eczarny:
@facebook-github-bot import

PR #688, opened 2016-11-23
RFC: Typed Props Component

ocrickard:
This is an RFC that allows a formalized C++ prop component. This lets us move to an implementation of composite components to be implemented as +render instead of +new, which will force people to avoid storing into ivars, and removes the necessity to use component scope in client code to use state.

This brings us a little closer to React, and provides an opportunity for automatic memoization.

### DIFF
--- a/ComponentKit/Components/CKTestTypedPropsComponent.h
+++ b/ComponentKit/Components/CKTestTypedPropsComponent.h
@@ -15,4 +15,6 @@ struct CKTestTypedPropsComponentProps {
 
 @interface CKTestTypedPropsComponent : CKTypedPropsComponent
 
+CKTypedPropsComponentConstructor(CKTestTypedPropsComponentProps);
+
 @end
--- a/ComponentKit/Components/CKTestTypedPropsComponent.h
+++ b/ComponentKit/Components/CKTestTypedPropsComponent.h
@@ -1,0 +1,18 @@
+//
+//  CKTestTypedPropsComponent.h
+//  ComponentKit
+//
+//  Created by Oliver Rickard on 11/23/16.
+//
+//
+
+#import <ComponentKit/CKTypedPropsComponent.h>
+
+struct CKTestTypedPropsComponentProps {
+  NSString *string;
+  UIFont *font;
+};
+
+@interface CKTestTypedPropsComponent : CKTypedPropsComponent
+
+@end
--- a/ComponentKit/Components/CKTestTypedPropsComponent.h
+++ b/ComponentKit/Components/CKTestTypedPropsComponent.h
@@ -15,6 +15,8 @@ struct CKTestTypedPropsComponentProps {
 
 @interface CKTestTypedPropsComponent : CKTypedPropsComponent
 
-CKTypedPropsComponentConstructor(CKTestTypedPropsComponentProps);
++ (instancetype)newWithProps:(const CKTestTypedPropsComponentProps &)props
+                        view:(const CKComponentViewConfiguration &)view
+                        size:(const CKComponentSize &)size;
 
 @end
--- a/ComponentKit/Components/CKTestTypedPropsComponent.h
+++ b/ComponentKit/Components/CKTestTypedPropsComponent.h
@@ -1,10 +1,12 @@
-//
-//  CKTestTypedPropsComponent.h
-//  ComponentKit
-//
-//  Created by Oliver Rickard on 11/23/16.
-//
-//
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
 
 #import <ComponentKit/CKTypedPropsComponent.h>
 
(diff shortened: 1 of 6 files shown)


ComponentKit/Components/CKTestTypedPropsComponent.h
```diff
@@ -0,0 +1,24 @@
...
+  UIFont *font;
+};
+
+@interface CKTestTypedPropsComponent : CKTypedPropsComponent
+
++ (instancetype)newWithProps:(const CKTestTypedPropsComponentProps &)props
+                        view:(const CKComponentViewConfiguration &)view
+                        size:(const CKComponentSize &)size;
+
+@end
```
ocrickard: This is just here as a demonstration. In order to provide type-safety, you have to provide your own +new that does a pass-through to the parent's +new method.

ComponentKit/Components/CKTypedPropsComponent.h
```diff
@@ -0,0 +1,61 @@
...
+template <typename T>
+struct CKRequiredProp {
+  CKRequiredProp<T>() = delete;
+  CKRequiredProp<T>(const T &val) : _value(val) {};
+  operator T() const {
+    return _value;
+  }
+private:
+  T _value;
+};
```
ocrickard: I'm liking this pattern. I think we should formalize it as the recommended way to provide required props. I dislike the mixed named Obj-C param/options structs as our components have grown more complex.

ComponentKit/Components/CKTypedPropsComponent.h
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  template<typename T> operator T () const
+  {
+    return std::dynamic_pointer_cast<holder<T>>(_content)->content;
+  }
+
+private:
+  std::shared_ptr<placeholder> _content;
+};
```
ocrickard: This is basically boost::any. It copies any struct to the heap, and allows implicit conversion to the underlying type. Conversion to *any* type will successfully compile, but when done at runtime, may fail with a failed exception.

ComponentKit/Components/CKTestTypedPropsComponent.mm
```diff
@@ -0,0 +1,42 @@
...
+@implementation CKTestTypedPropsComponent
+
++ (instancetype)newWithProps:(const CKTestTypedPropsComponentProps &)props
+                        view:(const CKComponentViewConfiguration &)view
+                        size:(const CKComponentSize &)size
+{
+  return [self newWithPropsStruct:props
+                             view:view
+                             size:size];
+}
```
ocrickard: Sucks we have to do this passthrough. Anyone have ideas on how to remove it while maintaining type-safety?

ComponentKit/Components/CKTestTypedPropsComponent.mm
```diff
@@ -0,0 +1,42 @@
...
+
++ (CKComponent *)renderWithProps:(const CKTypedComponentStruct &)p
```
ocrickard: The subclass should implement this method, but is forced to unwrap the props. I don't think that's *too* big of a deal since it's internal to the component, and we currently do this for id type to state already, so it's not a dealbreaker for me.

ComponentKit/Components/CKTypedPropsComponentSubclass.h
```diff
@@ -0,0 +1,28 @@
...
++ (instancetype)newWithPropsStruct:(const CKTypedComponentStruct &)props
+                              view:(const CKComponentViewConfiguration &)view
+                              size:(const CKComponentSize &)size;
+
++ (CKComponent *)renderWithProps:(const CKTypedComponentStruct &)props
+                           state:(id)state
+                            view:(const CKComponentViewConfiguration &)view
+                            size:(const CKComponentSize &)size;
+
+@property (nonatomic, assign, readonly) CKTypedComponentStruct props;
```
GotEmB: @ocrickard, how about making this property an `std::shared_ptr< CKTypedComponentStruct>` where `CKTypedComponentStruct<T>` is a very simple templated structure that just stores a member of type `T` (no handling optionals).

The `shared_ptr` itself should be able to represent null as a default value, `*props == nullptr`.

ocrickard: Well, Obj-C code doesn't have support for templates. We have to only interact with *specializations of templates* within Obj-C interfaces. This is why all this complicated logic exists.

Is there a way to support CKTypedComponentStruct<T>?

ocrickard: Yep, that's almost exactly what we're doing here. Sadly, that allows you to pass any subclass of the component struct to any component...

ryanrhee:
How come you removed the macros?

ComponentKit/Components/CKTypedPropsComponentSubclass.h
GotEmB: :/ I don't see a good way to get the type-safety this without making the entire component a C++ object inheriting from say `CK::Component`.

ryanrhee: +1, I can't think of any way of making this type-safe as an objc class.

We could make them actual c++ classes, then have objc wrappers around them, but that seems like it would affect binary size. :cry:

GotEmB: @ryanrhee It would really increase the binary size right? I'm thinking we'd have exactly one ObjC wrapper class that wraps any subclass of `CK::Component`.

ocrickard:
Alright, I'm closing this for now. Let's revive it down the road if we want to take on the project.

PR #689, opened 2016-11-28
Fix crash on unmounted component delegate attributes

ocrickard:
Delegates, and gesture attributes are implemented with this forwarder object. They're set on the view through an applicator. Sadly, view applicators do not run on entry into the reuse pool for performance reasons. If no change is necessary, we don't want to unset it for no reason.

What commonly happens is this: a gesture action is added to a view. Some action triggers a change in the view hierarchy, forcing the component view to enter the reuse pool, now hidden. The gesture, however, is still attached to the view, and its delegate is this forwarder object. Now, due perhaps to a pan gesture recognizer or long-press gesture recognizer, we get a late gesture method invocation on the delegate after unmounting.

We were hard-crashing in this delegate forwarder in the time between unmount of the component, and un-applicator removing the gesture handlers from the views. The view's component property will return nil for forwardingTargetForSelector: and the method will be directly invoked on this delegate, crashing the app.

Short of having a "un-apply the delegate before entrance into the reuse pool", this is the best option to prevent us from crashing. We assert, but don't hard-crash.


ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -61,9 +61,23 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
+{
+  // The delegate forwarder is applied as a component view attribute, which is not un-applied to the view on entry
+  // into the reuse pool, yet the *component* property on the view will begin returning nil in this case. This would
+  // turn into a hard-crash because above forwardingTargetForSelector: will return nil, and the method will be directly
+  // invoked on this object. In this case, we have no option but to assert, and not crash, as we do with unhandled
+  // component actions.
+  const BOOL shouldRespond = std::find(_selectors.begin(), _selectors.end(), aSelector) != std::end(_selectors);
+  if (!shouldRespond) {
+    [super doesNotRecognizeSelector:aSelector];
+  }
```
adamjernst: should we otherwise log the dropped message somehow?

ocrickard: Well, it asserts it above. I didn't want two assertions. I'd be fine moving it here, but doesn't matter really.

ocrickard:
@facebook-github-bot import

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
eczarny: It may be a bit more obvious if the forwarder asserts closer to where it previously crashed. Assuming this is where the crash was before.

ocrickard: Sure, sounds good to me.

ocrickard:
@facebook-github-bot import

eczarny:
Looks reasonable to me.

ocrickard:
Ugh, sadly, the first approach was insufficient. The fast-forwarding path won't work here, since we're going to return nil, and ultimately the objc_msgSend call will crash if we don't provide a method signature. The problem is that we may not *have* a target at that point, and simply implementing doesNotRecognizeSelector: isn't sufficient. The runtime kills the app right after calling that, which I had apparently forgotten.

Now we've moved to the slow forwarding path when the fast fowarding path fails and forwardingTargetForSelector: returns nil.

We return a dummy method signature to force the forwarding system to call forwardInvocation so we can silently fail to call the method instead of crashing.

Added a unit test. Before this update, the unit test would assert, but then would just crash in __forwarding__ after calling doesNotRecognizeSelector:

ocrickard:
@facebook-github-bot import

ComponentKitTests/CKComponentDelegateAttributeTests.mm
```diff
@@ -110,6 +110,38 @@ - (void)testProxiedEventsProxy
...
+  UIView *container = [UIView new];
+  NSSet *mounted = CKMountComponentLayout(layout, container, nil, nil);
+  
+  XCTAssertFalse(hierarchy.receivedScroll, @"Should not have triggered yet");
+  
+  UIScrollView *scroll = findScrollView(container);
+  
+  CKUnmountComponents(mounted);
+  
+  XCTAssertThrows(scroll.contentOffset = CGPointMake(0, 100));
```
ocrickard: Before this update, this line would throw, but then hard-crash. Now it just throws.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -53,9 +57,51 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
+
+  if (selectorInList(aSelector, _selectors)) {
+    CKComponent *responder = _view.ck_component;
+    id target = [responder targetForAction:aSelector withSender:responder];
+    // We must return a non-nil method signature even if there is no real method signature to return, or we will just
+    // crash in the forwarding system. This ensures the forwardInvocation: call above is called.
+    return (target ? [target methodSignatureForSelector:aSelector] : [self methodSignatureForSelector:@selector(_doNothing)]);
+  }
+
+  return nil;
```
ocrickard: Before going down this route I tried a variety of approaches... The standard NSProxy won't work because we actually do need to return a method signature, even if we don't have an alternate target for the selector.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -53,9 +57,51 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
 }
 
 - (id)forwardingTargetForSelector:(SEL)aSelector
```
ocrickard: Note here, we maintain the fast-forwarding path. Only in the case that the target is nil do we fall back to the methodSignature/forwardInvocation path.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -53,9 +57,51 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
+  SEL selector = anInvocation.selector;
+  if (selectorInList(selector, _selectors)) {
+    CKComponent *responder = _view.ck_component;
+    id target = [responder targetForAction:selector withSender:responder];
+    if (!target) {
+      CKFailAssert(@"Delegate method is being called on an unmounted component's view: %@ selector:%@", _view, NSStringFromSelector(selector));
+      return;
+    }
+    [anInvocation invokeWithTarget:target];
+  } else if(selector == @selector(_doNothing)) {
```
ocrickard: I'll clean this up. It works as-is, but no need for this branch at all.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -53,9 +57,51 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
+  // turn into a hard-crash because above forwardingTargetForSelector: will return nil, and the method will be directly
+  // invoked on this object. In this case, we have no option but to assert, and not crash, as we do with unhandled
+  // component actions.
+  SEL selector = anInvocation.selector;
+  if (selectorInList(selector, _selectors)) {
+    CKComponent *responder = _view.ck_component;
+    id target = [responder targetForAction:selector withSender:responder];
+    if (!target) {
+      CKFailAssert(@"Delegate method is being called on an unmounted component's view: %@ selector:%@", _view, NSStringFromSelector(selector));
+      return;
```
eczarny: Hm. This is a bit confusing. `-[CKComponentDelegateForwarder methodSignatureForSelector:]` seems to be doing the same thing. If the target cannot be obtained `CKComponentDelegateForwarder` will simply message `-[CKComponentDelegateForwarder _doNothing]` instead. Do we need to implement `-[CKComponentDelegateForwarder forwardInvocation:]` at all? Could we just assert in `-[CKComponentDelegateForwarder _doNothing]`?

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -53,9 +57,51 @@ - (BOOL)respondsToSelector:(SEL)aSelector
...
+    }
+    [anInvocation invokeWithTarget:target];
+  } else if(selector == @selector(_doNothing)) {
+    
+  } else {
+    [super forwardInvocation:anInvocation];
+  }
+}
+
+- (void)_doNothing {}
```
eczarny: I'd rename this to make it a bit more obvious why it exists. Maybe `-[CKComponentDelegateForwarder _handleUnknownResponder]`?

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
ocrickard: Ahh, yeah, you can't just implement methodSignature, doesNotRespondToSelector: is called if only the method signature is implemented, and we'll crash. You have to implement both.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
ocrickard: This is never actually invoked... ever.

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
eczarny: Got it. Makes sense!

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
eczarny: Oh, you're right. Good point.

ocrickard:
@facebook-github-bot import

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -48,24 +48,70 @@ + (instancetype)newWithSelectors:(CKComponentForwardedSelectors)selectors
   return f;
 }
 
+/** 
+ This method is never invoked, and is used to provide a dummy method signature to the forwarding system if our
+ normal fast-path forwarding doesn't work because the component has been unmounted.
+ */
+- (void)_doNothing {}
```
adamjernst: (y)

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
adamjernst: sounds good!

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
```diff
@@ -48,24 +48,70 @@ + (instancetype)newWithSelectors:(CKComponentForwardedSelectors)selectors
...
 - (id)forwardingTargetForSelector:(SEL)aSelector
 {
   CKComponent *responder = _view.ck_component;
   return [responder targetForAction:aSelector withSender:responder];
 }
 
-@end
+static BOOL selectorInList(SEL selector, const CKComponentForwardedSelectors &selectors)
+{
+  return std::find(selectors.begin(), selectors.end(), selector) != std::end(selectors);
```
adamjernst: `selectors.end()` ?

adamjernst:
I like it!

ComponentKit/Utilities/CKComponentDelegateForwarder.mm
ocrickard: lol, sure, this was the existing code.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ocrickard:
this landed internally. I'm going to wait for travis, then merge here.

eczarny:
I kicked off another Travis CI build. Maybe time to try a different CI provider? This is getting a bit ridiculous.

ocrickard:
Alright, I'm merging this. We've tried enough times here, we have green builds internally, and what appears to be the same intermittent build failure for tvos. I agree, we need to consider another provider.

PR #691, opened 2016-12-08
Redo how debug "reflow" works

adamjernst:
This was never updated after we transitioned away from `CKComponentLifecycleManager`, so it was mostly broken. Now instead of attempting to use the view hierarchy to trigger reflow, we simply message all interested parties and tell them to update themselves.

Crucially this means that offscreen components are recomputed too, not just onscreen ones. Neat.

### DIFF
--- a/ComponentKit/Core/CKComponentLifecycleManager.mm
+++ b/ComponentKit/Core/CKComponentLifecycleManager.mm
@@ -16,6 +16,7 @@
 
 #import "CKComponent.h"
 #import "CKComponentInternal.h"
+#import "CKComponentDebugController.h"
 #import "CKComponentLayout.h"
 #import "CKComponentLifecycleManagerAsynchronousUpdateHandler.h"
 #import "CKComponentMemoizer.h"
@@ -36,6 +37,9 @@ const CKComponentLifecycleManagerState CKComponentLifecycleManagerStateEmpty = {
   .layout = {},
   .root = nil,
 };
+
+@interface CKComponentLifecycleManager () <CKComponentDebugReflowListener>
+@end
 
 @implementation CKComponentLifecycleManager
 {
@@ -62,6 +66,7 @@ const CKComponentLifecycleManagerState CKComponentLifecycleManagerStateEmpty = {
   if (self = [super init]) {
     _componentProvider = componentProvider;
     _sizeRangeProvider = sizeRangeProvider;
+    [CKComponentDebugController registerReflowListener:self];
   }
   return self;
 }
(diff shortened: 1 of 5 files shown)


ComponentKit/Debug/CKComponentDebugController.mm
```diff
@@ -109,9 +111,6 @@ + (BOOL)debugMode
     return context; // no need for a debug view if the component has a view.
   }
 
-  static CK::StaticMutex l = CK_MUTEX_INITIALIZER;
-  CK::StaticMutexLocker lock(l);
```
adamjernst: unnecessary; mount is only performed on the main thread.

ComponentKit/Debug/CKComponentDebugController.mm
```diff
@@ -133,46 +132,35 @@ + (BOOL)debugMode
...
-    }
+  NSArray<id<CKComponentDebugReflowListener>> *copiedListeners;
+  {
+    std::lock_guard<std::mutex> l(reflowMutex);
+    copiedListeners = [reflowListeners allObjects];
   }
-  if (view) {
-    CKRecursiveComponentReflow(view);
+  for (id<CKComponentDebugReflowListener> listener in copiedListeners) {
+    [listener didReceiveReflowComponentsRequest];
```
eczarny: This is so much better than traversing the view hierarchy to find the component lifecycle manager or explicitly telling component hosting views to layout.

ComponentKit/Core/CKComponentLifecycleManager.mm
```diff
@@ -62,6 +66,7 @@ - (instancetype)initWithComponentProvider:(Class<CKComponentProvider>)componentP
   if (self = [super init]) {
     _componentProvider = componentProvider;
     _sizeRangeProvider = sizeRangeProvider;
+    [CKComponentDebugController registerReflowListener:self];
```
eczarny: Though it's a bit of a shame to have `CKComponentLifecycleManager` aware of debug code I think it's totally worthwhile considering how much it simplifies `CKComponentDebugController`.

ComponentKit/Debug/CKComponentDebugController.mm
```diff
@@ -133,46 +132,35 @@ + (BOOL)debugMode
...
-  } else {
-    for (UIView *subview in view.subviews) {
-      CKRecursiveComponentReflow(subview);
-    }
+  if (listener == nil) {
+    return;
+  }
+  std::lock_guard<std::mutex> l(reflowMutex);
+  if (reflowListeners == nil) {
+    reflowListeners = [NSHashTable weakObjectsHashTable];
```
eczarny: Nice. So we weakly hold the reflow listeners to naturally keep the collection bounded.

ComponentKit/Debug/CKComponentDebugController.mm
eczarny: Do we want to assert this is called on the main thread? Or do you think that's overkill in this case?

eczarny:
Looks good to merge to me. I like how much this simplifies things.

ComponentKit/Debug/CKComponentDebugController.mm
adamjernst: I think it's overkill in this particular case.

ocrickard:
Looks good to me.

PR #692, opened 2016-12-08
Add a relinquish delay controller

ocrickard:
Stateful views have 3 major jobs:

1. When an insertion occurs above the viewport, they should hold onto their view and not allow it to enter the reuse pool to prevent flickering of videos while actively playing.
2. Stateful views should not be reused when the controller is manually managing their appearance (we've entered a fullscreen video or photo or text editor, and the stateful view cannot be reused in another context).
3. Stateful views generally are our heaviest, memory-intensive views, and unlike normal views, should participate in a global reuse pool. Normal component views are simply hidden, and their reuse pool is localized to a view hierarchy. Stateful views on the other hand share a global reuse pool.

In order to enable (1), we have a hack that does a dispatch_async before adding a stateful view to the global reuse pool (in order to achieve (3)). However, in normal use, this delayed relinquish can mean that we grow our reuse pool, and create many heavy views for no benefit.

This is my proposal to fix the situation. I think we should just do something simple here, and provide a global "delay relinquish" during datasource updates. By default, we remain with the status quo, but if you really want to manage this stuff yourself, you can disable the delayed insertion trick, and manually enable the delayed relinquish hack for a specific runloop turn.

So, before inserting into a collection view, you would call `[[CKStatefulViewRelinquishController sharedInstance] delayRelinquishForRunloopTurn]`. This would mean that any cell reuse/remounting that happens for that runloop turn as the collection view shuffles its cells would have the benefit of the delay hack, and the *rest* of the time, it is disabled. This allows us to have both worlds.

Is this perfect? Absolutely not. It's darn-right ugly. However, we've used similar patterns for async transactions for several years, and the simple flag inside infra has not been a problem.

Other ideas we considered:

1. Some people proposed we try to differentiate the "about to jump viewport" and "you're just being reused" points. I believe this is possible, but would result in a large amount of tricky code that will be very product-specific. I believe it punctures our abstractions even more than this proposal.
2. Provide a "pending" reuse pool that is a staging area for the normal reuse pool. For normal operation, the hope is that if we did this, then we could prioritize returning stateful views that are truly unused, but we still allow access to the ones that are delaying. My problem with this proposal is that it can lead to hard-to-debug flickers as we fall back to the pending reuse pool.

Overall, I believe this proposal is the best idea that we currently have. Either delayed relinquish is enabled, or it is disabled. Simple.

The main problem however is going to be this call to `[[CKStatefulViewRelinquishController sharedInstance] delayRelinquishForRunloopTurn]` in your datasource before touching cells. It also means if you call `-reloadData` on your collection view manually, you have to manually delay relinquish there as well. Since there should be no reason to ever call `-reloadData` on a Components-managed collection view, I don't view that as a huge problem.

## What I would like feedback on:

1. **Naming**. This came out ugly. Suggestions?
2. **Singleton**. Alternative solutions? I could manually thread the controller down to the stateful views, but I view that as yet more boilerplate.
3. **Defaults**. I have enabled the delayed relinquish hack by default in this diff since it's the status quo. Is that right?

### DIFF
--- a/ComponentKit.xcodeproj/project.pbxproj
+++ b/ComponentKit.xcodeproj/project.pbxproj
@@ -373,11 +373,15 @@
 		A29D85791D80D37F00913E60 /* CKComponentContextHelper.h in Headers */ = {isa = PBXBuildFile; fileRef = A243B52C1D79A40800E6C393 /* CKComponentContextHelper.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		A2CD66321AF2F0C70083A839 /* CKTransactionalComponentDataSourceStateTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = A2CD66311AF2F0C70083A839 /* CKTransactionalComponentDataSourceStateTests.mm */; };
 		A2D20CF41B431CCF002B2DC7 /* CKComponentHostingViewAsyncStateUpdateTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = A2D20CF31B431CCF002B2DC7 /* CKComponentHostingViewAsyncStateUpdateTests.mm */; };
-		B1EF8FE81DD1562B00E74F43 /* CKComponentActionInternal.h in Headers */ = {isa = PBXBuildFile; fileRef = B1EF8FE71DD1562B00E74F43 /* CKComponentActionInternal.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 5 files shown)


ocrickard:
Alright, the more I think about this, the more I would actually like to reconsider my second proposal here. It's possible it will be less problematic than I thought.

ocrickard:
Alright, I'm closing this in favor of my pending pool version. I think it makes more sense to start there, and do this only if it doesn't work.

PR #693, opened 2016-12-09
Pending reuse pool

ocrickard:
This is the counter-proposal to https://github.com/facebook/componentkit/pull/692.

The core idea here is that we introduce a pending reuse pool, and move the dispatch_async hack into the reuse pool. This will allow us to prefer reusing "fully" relinquished views, but will allow fallbacks in the case that no fully relinquished view is present.

This still needs a little bit of work, but here it is.

### DIFF
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -133,13 +133,13 @@
 {
   if ([self canRelinquishStatefulView]) {
     // There is no guarantee when this reuse pool will execute this block
-    __weak __typeof(self) weakSelf = self;
+    CKStatefulViewComponentController *__weak weakSelf = self;
     [[CKStatefulViewReusePool sharedPool]
      enqueueStatefulView:_statefulView
      forControllerClass:[self class]
      context:_statefulViewContext
      mayRelinquishBlock:^BOOL{
-       __typeof(self) strongSelf = weakSelf;
+       CKStatefulViewComponentController *strongSelf = weakSelf;
        if (!strongSelf) {
          return YES;
        }
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -144,7 +144,7 @@
          return YES;
        }
        if (!strongSelf->_mounted && [strongSelf canRelinquishStatefulView]) {
-         [strongSelf willRelinquishStatefulView:_statefulView];
+         [strongSelf willRelinquishStatefulView:strongSelf->_statefulView];
          strongSelf->_statefulView = nil;
          strongSelf->_statefulViewContext = nil;
          return YES;
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -136,7 +136,7 @@
      enqueueStatefulView:_statefulView
      forControllerClass:[self class]
      context:_statefulViewContext
-     enqueueComplete:^BOOL{
+     mayRelinquishBlock:^BOOL{
        if (!_mounted && [self canRelinquishStatefulView]) {
          [self willRelinquishStatefulView:_statefulView];
          _statefulView = nil;
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -131,19 +131,21 @@
 
 - (void)_relinquishStatefulViewIfPossible
 {
-  [[CKStatefulViewReusePool sharedPool]
-   enqueueStatefulView:_statefulView
-   forControllerClass:[self class]
-   context:_statefulViewContext
-   enqueueComplete:^BOOL{
-     if (!_mounted && [self canRelinquishStatefulView]) {
-       [self willRelinquishStatefulView:_statefulView];
-       _statefulView = nil;
-       _statefulViewContext = nil;
-       return YES;
-     }
-     return NO;
-   }];
+  if ([self canRelinquishStatefulView]) {
+    [[CKStatefulViewReusePool sharedPool]
+     enqueueStatefulView:_statefulView
+     forControllerClass:[self class]
+     context:_statefulViewContext
+     enqueueComplete:^BOOL{
+       if (!_mounted && [self canRelinquishStatefulView]) {
+         [self willRelinquishStatefulView:_statefulView];
+         _statefulView = nil;
+         _statefulViewContext = nil;
+         return YES;
+       }
+       return NO;
+     }];
+  }
 }
 
 @end
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -131,18 +131,19 @@
 
 - (void)_relinquishStatefulViewIfPossible
 {
-  // Wait for the run loop to turn over before trying to relinquish the view. That ensures that if we are remounted on
-  // a different root view, we reuse the same view (since didMount will be called immediately after didUnmount).
-  dispatch_async(dispatch_get_main_queue(), ^{
-    if (_statefulView && !_mounted && [self canRelinquishStatefulView]) {
-      [self willRelinquishStatefulView:_statefulView];
-      [[CKStatefulViewReusePool sharedPool] enqueueStatefulView:_statefulView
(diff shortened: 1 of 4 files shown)


ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -10,40 +10,76 @@
...
+#import "CKAssert.h"
 
 #import <unordered_map>
 
+struct FBStatefulReusePoolItemEntry {
+  UIView *view;
+  CKStatefulViewReusePoolPendingMayRelinquishBlock block;
+};
+
 class FBStatefulReusePoolItem {
```
ocrickard: I decided to just reuse this custom multimap for my new pending pool and just add on the block API. For the normal pool, the block will be nil.

ComponentKit/StatefulViews/CKStatefulViewReusePool.h
```diff
@@ -20,6 +22,7 @@
 
 - (void)enqueueStatefulView:(UIView *)view
          forControllerClass:(Class)controllerClass
-                    context:(id)context;
+                    context:(id)context
+            enqueueComplete:(CKStatefulViewReusePoolPendingMayRelinquishBlock)enqueueComplete;
```
adamjernst: neat.

adamjernst:
I like this quite a bit.

ComponentKit/StatefulViews/CKStatefulViewReusePool.h
eczarny: I'd name the parameter to better match how the block is being used. "Enqueue complete" sounds more informative than instructive. Maybe something that better captures the spirit of how the block is being used?

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -10,40 +10,76 @@
...
+                           });
+    if (preferIt != _entries.end()) {
+      FBStatefulReusePoolItemEntry entry = *preferIt;
+      _entries.erase(preferIt);
+      if (entry.block == NULL || entry.block()) {
+        return entry.view;
+      }
+    }
+
+    if (_entries.empty()) {
```
eczarny: I'd include a comment explaining why we're checking if `_entries` is empty again. Even though there's a comment explaining what's going on just a bit further down.

At first I thought this was a mistake.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -10,40 +10,76 @@
...
   };
 
-  void addView(UIView *view)
+  void addEntry(const FBStatefulReusePoolItemEntry &entry)
   {
-    [views addObject:view];
+    _entries.push_back(entry);
   };
 
+  void absorbPendingPool(const FBStatefulReusePoolItem &otherPool, NSInteger maxEntries) {
```
eczarny: Nit: `{` on its own line.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
ocrickard: Turns out I didn't need this! If entries is empty, then entries.begin == entries.end, and we never enter the loop. I'm removing the check.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -10,40 +10,74 @@
...
   {
-    [views addObject:view];
+    _entries.push_back(entry);
   };
 
+  void absorbPendingPool(const FBStatefulReusePoolItem &otherPool, NSInteger maxEntries)
+  {
+    for (const FBStatefulReusePoolItemEntry &entry : otherPool._entries) {
+      if (entry.block == NULL || entry.block()) {
+        // The block was either not present, or evaluated to true, indicating it can join the normal reuse pool
```
eczarny: I'd rephrase this comment to mention why the entry can join the normal reuse pool. The logic is simple enough to follow so I don't see much of a need to reiterate it here. Why that condition is important is a bit more interesting.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -56,6 +90,8 @@ void addView(UIView *view)
 @implementation CKStatefulViewReusePool
 {
   std::unordered_map<std::pair<__unsafe_unretained Class, id>, FBStatefulReusePoolItem, PoolKeyHasher> _pool;
+  std::unordered_map<std::pair<__unsafe_unretained Class, id>, FBStatefulReusePoolItem, PoolKeyHasher> _pendingPool;
```
eczarny: Maybe name this to better indicate what's pending. `_pendingReusePool` is probably better. I may even rename `_pool` to `_reusePool` as well. Up to you.

ComponentKit/StatefulViews/CKStatefulViewReusePool.h
```diff
@@ -10,16 +10,24 @@
...
+
 - (UIView *)dequeueStatefulViewForControllerClass:(Class)controllerClass
                                preferredSuperview:(UIView *)preferredSuperview
                                           context:(id)context;
 
 - (void)enqueueStatefulView:(UIView *)view
          forControllerClass:(Class)controllerClass
-                    context:(id)context;
+                    context:(id)context
+         mayRelinquishBlock:(CKStatefulViewReusePoolPendingMayRelinquishBlock)mayRelinquishBlock;
```
ocrickard: Well, the view can say NO, and it won't be relinquished, so that's not representative. I think this name is fine. It's just an internal piece of infra, and is descriptive enough.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
ocrickard: Meh, this whole thing is a reuse pool. I don't think that improves readability at all. Up to you if you want to change it later though.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
ocrickard: Good point.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
```diff
@@ -131,18 +131,27 @@ - (void)_presentStatefulView
...
+         return YES;
+       }
+       if (!strongSelf->_mounted && [strongSelf canRelinquishStatefulView]) {
+         [strongSelf willRelinquishStatefulView:_statefulView];
+         strongSelf->_statefulView = nil;
+         strongSelf->_statefulViewContext = nil;
+         return YES;
+       }
+       return NO;
+     }];
```
eczarny: Another option is to just have this block call a private instance method so you can avoid the `weakSelf`/`strongSelf` dance.

ComponentKit/StatefulViews/CKStatefulViewReusePool.h
eczarny: Ah, good point. I was reading this block as being responsible for actually doing the relinquishing.

ComponentKit/StatefulViews/CKStatefulViewReusePool.mm
```diff
@@ -10,40 +10,78 @@
...
 
+  void absorbPendingPool(const FBStatefulReusePoolItem &otherPool, NSInteger maxEntries)
+  {
+    for (const FBStatefulReusePoolItemEntry &entry : otherPool._entries) {
+      // In the future, we should consider not evaluating the block here immediately, and letting it move into the
+      // normal reuse pool. That way we can let stateful components hold onto their own views without any
+      // reconfiguration for a longer period of time.
+      if (entry.block == NULL || entry.block()) {
+        // The stateful view component can decide not to allow reuse of its view if the component has re-mounted before
+        // the block is evaluated.
```
eczarny: Awesome! Much clearer.

clementgenzmer:
Does this fix the bug where a statefulView is to be recycled right now (in scroll view full of statefulView)?

ocrickard:
@clementgenzmer yes it does, at the possible cost of more flickers on insertions above the viewport. We'll QE the change. There are several other things we can do if this isn't sufficient. In scrolling video home locally we're finally able to reuse views at a really high rate with this in place.

ocrickard:
Ping @eczarny this has been updated.

eczarny:
Let's try it out.

ocrickard:
@facebook-github-bot import

ocrickard:
Landing this internally, merging here.

PR #694, opened 2016-12-09
Fix bad UICollectionView selector usage

dbworku:
Today, animating the bounds of a component will crash on pre-iOS 9 environments.

The `supplementaryViewForElementKind:atIndexPath:` selector on
`UICollectionView` [was introduced in iOS 9.](https://developer.apple.com/reference/uikit/uicollectionview/1618041-supplementaryviewforelementkind) However, ComponentKit
targets iOS 8.1+.

`CK_AT_LEAST_IOS9` was only available in the `FBSnapshotTestCase` project so I created `CKAvailability.h`, to avoid duplication.

I sanity checked this fix on device running iOS 8.4.1.

### DIFF
--- a/ComponentKit/Base/CKAvailability.h
+++ b/ComponentKit/Base/CKAvailability.h
@@ -17,9 +17,3 @@
 #define CK_AT_LEAST_IOS8_2 (kCFCoreFoundationVersionNumber >= 1142.16)
 #define CK_AT_LEAST_IOS9 (kCFCoreFoundationVersionNumber >= 1223.1)
 #define CK_AT_LEAST_IOS10_BETA_4 (kCFCoreFoundationVersionNumber >= 1345.0)
-
-#if __LP64__
-#define CK_64 1
-#else
-#define CK_64 0
-#endif
--- a/ComponentKit/Base/CKAvailability.h
+++ b/ComponentKit/Base/CKAvailability.h
@@ -1,0 +1,25 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <Availability.h>
+#import <AvailabilityInternal.h>
+
+#import <CoreFoundation/CFBase.h>
+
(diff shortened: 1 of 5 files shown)


eczarny:
@dbworku is the intended behavior then that bounds animations on iOS 8 will not work? I'll try getting to reviewing this a bit later today. Thanks for the fix!

dbworku:
@eczarny Note quite... The intended behavior is that, during bounds animations, we should not attempt to animate the supplementary view on iOS 8.

ComponentKit/Base/CKAvailability.h
```diff
@@ -0,0 +1,25 @@
...
+#define CK_AT_LEAST_IOS8 (kCFCoreFoundationVersionNumber > 847.27)
+#define CK_AT_LEAST_IOS8_2 (kCFCoreFoundationVersionNumber >= 1142.16)
+#define CK_AT_LEAST_IOS9 (kCFCoreFoundationVersionNumber >= 1223.1)
+#define CK_AT_LEAST_IOS10_BETA_4 (kCFCoreFoundationVersionNumber >= 1345.0)
+
+#if __LP64__
+#define CK_64 1
+#else
+#define CK_64 0
+#endif
```
eczarny: I'd drop `CK_64` from here since it's really only needed for snapshot tests.

ComponentKit/Base/CKAvailability.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Availability.h>
+#import <AvailabilityInternal.h>
+
+#import <CoreFoundation/CFBase.h>
+
+#define CK_AT_LEAST_IOS8 (kCFCoreFoundationVersionNumber > 847.27)
+#define CK_AT_LEAST_IOS8_2 (kCFCoreFoundationVersionNumber >= 1142.16)
+#define CK_AT_LEAST_IOS9 (kCFCoreFoundationVersionNumber >= 1223.1)
+#define CK_AT_LEAST_IOS10_BETA_4 (kCFCoreFoundationVersionNumber >= 1345.0)
```
eczarny: Reminds me I need to update this.

eczarny:
> The intended behavior is that, during bounds animations, we should not attempt to animate the supplementary view on iOS 8.

Gotcha. The fallback before is to essentially ignore what was introduced to originally support animating supplementary views. Seems reasonable.

eczarny:
Awesome! I'll ship this once CI passes.

eczarny:
@facebook-github-bot shipit

PR #695, opened 2016-12-11
[HACK][RFC] Add basic and hacky swift interop

ocrickard:
This is definitely not merge-worthy code, but I wanted to give Swift interop a shot over my weekend. This compiles, I haven't gotten to testing it really yet, but it should work.

The tough part I think will be the large surface area of our framework. Perhaps we could focus on a limited subset at first, get that right, and push outwards.

Also, I really dislike how hard it would be to keep our C interfaces in sync. Yet another thing to test. We'd want to code-gen these interfaces from the Obj-C++ AST.


ocrickard:
We're not gonna merge this, I just wanted it up so others could start from here if they want to.

appleguy:
@ocrickard very cool, excited to follow along!

PR #696, opened 2016-12-12
Include component and mounted view class names when logging unexpected view on mount

eczarny:
Knowing the class name of the component and the unexpectedly mounted view should help investigate where these kinds of issues are occurring.

### DIFF
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -154,7 +154,7 @@ struct CKComponentMountInfo {
     _mountInfo->viewContext = {v, {{0,0}, v.bounds.size}};
     return {.mountChildren = YES, .contextForChildren = effectiveContext.childContextForSubview(v, g.didBlockAnimations)};
   } else {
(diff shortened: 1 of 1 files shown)


ComponentKit/Core/CKComponent.mm
```diff
@@ -154,7 +154,7 @@ - (CKComponentViewContext)viewContext
     _mountInfo->viewContext = {v, {{0,0}, v.bounds.size}};
     return {.mountChildren = YES, .contextForChildren = effectiveContext.childContextForSubview(v, g.didBlockAnimations)};
   } else {
-    CKAssertNil(_mountInfo->view, @"Didn't expect to sometimes have a view and sometimes not have a view");
+    CKAssertNil(_mountInfo->view, @"Didn't expect %@ to have a mounted view: %@", [self class], [_mountInfo->view class]);
```
eczarny: I thought about including some information from the view configuration itself but figured this may be enough. I'd be open to including the identifier of the view configuration's view class as well if we think that would be helpful.

eczarny:
@facebook-github-bot import

eczarny:
@facebook-github-bot import

ocrickard:
@facebook-github-bot shipit

PR #697, opened 2016-12-13
Add __covariant to CKComponentController to avoid casting self.component

maxianzhe:
API clients won't need to cast self.component to their CKComponent subclass inside CKComponentController subclasses anymore.

### DIFF
--- a/ComponentKit/Core/CKComponentController.h
+++ b/ComponentKit/Core/CKComponentController.h
@@ -14,7 +14,7 @@
 
 @class CKComponent;
 
-@interface CKComponentController : NSObject
+@interface CKComponentController<__covariant ComponentType:CKComponent *> : NSObject
 
 /** The controller's component is not mounted, but is about to be. */
 - (void)willMount NS_REQUIRES_SUPER;
@@ -58,7 +58,7 @@
 - (void)componentTreeDidDisappear NS_REQUIRES_SUPER;
 
 /** The current version of the component. */
-@property (nonatomic, weak, readonly) CKComponent *component;
+@property (nonatomic, weak, readonly) ComponentType component;
 
 /** The view created by the component, if currently mounted. */
 @property (nonatomic, strong, readonly) UIView *view;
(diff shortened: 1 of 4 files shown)


ocrickard:
So good. You're quick.

ComponentKit/Core/CKComponentController.h
```diff
@@ -14,7 +14,7 @@
 
 @class CKComponent;
 
-@interface CKComponentController : NSObject
+@interface CKComponentController<__covariant ComponentType:CKComponent *> : NSObject
```
ocrickard: Perfect, so ComponentType is defined as a covariant value with superclass CKComponent.

ocrickard:
Would you mind updating the example code and any controllers in the core with the covariant before landing? This looks perfect, but before landing I think it's worth providing examples and drinking our own kool-aid.

maxianzhe:
@ocrickard sure! i will update some example code in the core.

maxianzhe:
@ocrickard Oliver, I only found two Core classes that casts self.component in componentkit codebase.

any docs I should update?

ocrickard:
Hmm, isn't there CKStatefulViewComponentController? I know we don't cast its component, but worth adding the covariant there too.
Re docs, that's fine, let's handle that separately. I'll update it later.

ComponentKit/StatefulViews/CKStatefulViewComponentController.h
```diff
@@ -32,7 +32,7 @@
 
  This controller's corresponding component must subclass CKStatefulViewComponent.
  */
-@interface CKStatefulViewComponentController : CKComponentController
+@interface CKStatefulViewComponentController<__covariant ComponentType:CKComponent *> : CKComponentController
```
ocrickard: Interesting. I almost wonder if this component controller should be covariant not only in the component type, but the view type... Anyway, let's deal with that later.

I'm not positive about this, but shouldn't this interface be:

```
@interface CKStatefulViewComponentController<__covariant ComponentType:CKComponent *> : CKComponentController<ComponentType>
```

ocrickard:
Back to you for one more question.

eczarny:
This is amazing.

ComponentKit/StatefulViews/CKStatefulViewComponentController.h
maxianzhe: @ocrickard Oliver, you are right. I forgot about the inheritance. Will update.

ocrickard:
This is super awesome.

ocrickard:
@facebook-github-bot shipit

maxianzhe:
@ocrickard will this PR close by itself?

ocrickard:
Well, it's supposed to! I'll merge it.

PR #698, opened 2016-12-13
Move animation ivars into lazy unique_ptr

ocrickard:
In investigating component memory consumption, these ivars were consuming a lot of space, but are relatively rarely used. Move them into a lazily-instantiated wrapper struct.

### DIFF
--- a/ComponentKit/Core/CKComponentController.mm
+++ b/ComponentKit/Core/CKComponentController.mm
@@ -13,6 +13,8 @@
 
 #import <ComponentKit/CKAssert.h>
 
+#import <memory>
+
 #import "CKComponentInternal.h"
 #import "CKComponentSubclass.h"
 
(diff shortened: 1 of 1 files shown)


ComponentKit/Core/CKComponentController.mm
```diff
@@ -66,17 +68,37 @@ static void eraseAnimation(CKAppliedComponentAnimationMap &map, CKComponentAnima
...
+};
+
+struct CKComponentControllerAnimationWrapper {
+public:
+  CKComponentControllerAnimationData *operator->() {
+    if (_animationData == nullptr) {
+      _animationData.reset(new CKComponentControllerAnimationData());
+    }
+    return _animationData.get();
+  };
```
eczarny: Seems like a nice way to make this lazy with minimal complexity in the `CKComponentController` implementation. I like this approach.

ComponentKit/Core/CKComponentController.mm
```diff
@@ -66,17 +68,37 @@ static void eraseAnimation(CKAppliedComponentAnimationMap &map, CKComponentAnima
...
   CKComponentControllerState _state;
   BOOL _updatingComponent;
   BOOL _performedInitialMount;
   CKComponent *_previousComponent;
-  CKComponentAnimationID _nextAnimationID;
-  std::vector<CKPendingComponentAnimation> _pendingAnimationsOnInitialMount;
-  CKAppliedComponentAnimationMap _appliedAnimationsOnInitialMount;
-  std::vector<CKPendingComponentAnimation> _pendingAnimations;
-  CKAppliedComponentAnimationMap _appliedAnimations;
+  CKComponentControllerAnimationWrapper _animationData;
```
eczarny: This makes the state managed by the `CKComponentController` much clearer (no need to animation-related state to add clutter). Nice side effect of this work.

ComponentKit/Core/CKComponentController.mm
```diff
@@ -142,32 +164,36 @@ - (void)componentDidMount:(CKComponent *)component
...
+          const CKComponentAnimation &anim = pendingAnimation.animation;
+          [CATransaction begin];
+          CKComponentAnimationID animationID = _animationData->nextAnimationID++;
+          [CATransaction setCompletionBlock:^() {
+            eraseAnimation(_animationData->appliedAnimations, animationID);
+          }];
+          _animationData->appliedAnimations.insert({animationID, {anim, anim.didRemount(pendingAnimation.context)}});
+          [CATransaction commit];
+        }
+        _animationData->pendingAnimations.clear();
```
eczarny: It's a bit of a bummer that this block of code is so similar to the one above. Seems like a potential place for behavior to get out of sync. Likely not a concern of this change but maybe something we can clean up separately?

ComponentKit/Core/CKComponentController.mm
ocrickard: Yeah, totally. I like it.

ocrickard:
@facebook-github-bot import

PR #699, opened 2016-12-14
Include component backtrace when logging unexpected view on mount

eczarny:
Follow up to #696. Including the component backtrace aids in debugging this assertion.


eczarny:
@facebook-github-bot import

PR #701, opened 2016-12-14
Fix crash after assertion fire

ocrickard:
If assertions are disabled, or if they log and continue, then the assertion above this call won't stop execution, and we can get to this for loop where the number of arguments in the type signature and the type encodings don't match. In that case we will hard-crash. Let's not crash if that happens.


ocrickard:
@facebook-github-bot import

PR #702, opened 2016-12-17
Fix equality comparison on the transactional DS config when context is nil

jarman:
When the context is null the equality comparison is always false.


ocrickard:
Looks correct to me.

@facebook-github-bot shipit

ocrickard:
This landed internally. Merging.

PR #703, opened 2016-12-20
Correct identifier used in view action attributes

ocrickard:
When I updated the actions to use scoped/target actions I forgot to update this. What this means is that the identifier for the action only included the selector, so when a new component mounted we wouldn't update the action registered on the view. That means that old actions were being left on the views, which could cause weird problems. This was fine when we were exclusively responder-chain-based since it would always crawl up from the current component, but no longer. Now we take a reference to the target scope or object.

Fix is simple, just include the pointer value of the action at the time of application. This means that we'll update the value when a new target/selector pair is set.


ocrickard:
@facebook-github-bot import

adamjernst:
ship it!

ocrickard:
Landing internally, merging here.

PR #704, opened 2016-12-21
Fix gesture actions still using selector identifier

ocrickard:
I missed this one yesterday. Identifiers for actions should be built with both the selector and its target, or we risk not installing an updated action attribute when components update and we get a new target.


ocrickard:
@facebook-github-bot import

ocrickard:
Landing internally, merging here.

PR #708, opened 2016-12-24
Expose CKAsyncLayerSubclass and mark it as private

conradev:
Supplements #563, allowing consumers of the framework to implement their own async layers


ocrickard:
Hmm, I don't see any problem with this.

ocrickard:
@facebook-github-bot shipit

ocrickard:
This landed internally. Unclear why this isn't closed.

PR #709, opened 2016-12-27
Maintain strong reference to controller while in reuse pool

ocrickard:
If the component controller is deallocated before the block is called, then we won't call willRelinquishStatefulView on the controller at all, leaving important cleanup work undone. Before my cleanup this was a strong ref. I changed it to a weak ref without really thinking through the details. Returning to a strong reference.


ocrickard:
I'm reverting and going to re-apply instead.

PR #714, opened 2017-01-06
Add support for attributed titles to button components

conradev:
`attributedTitles` take precedent over `titles` in all cases

Let me know what you think!


ocrickard:
Sweet! Definitely something we'd want, but generally we'd do this in a different way. We'd take the existing constructor, and instead of taking a map to NSString, we'd use an intermediate C++ struct with 2 constructors, one for NSString, and one for NSAttributedString. This way it's an algebraic type, and all existing code can stay as-is. You can just pass an NSAttributedString where previously we accepted an NSString, and we'll do the right thing.

What do you think of that option?

ocrickard:
BTW, build is failing.

conradev:
I like that idea, it is definitely a better API! Can do that and fix the build.

conradev:
Thoughts on this implementation?

ComponentKit/Components/CKButtonComponent.h
```diff
@@ -18,6 +18,22 @@ struct CKButtonComponentAccessibilityConfiguration {
   NSString *accessibilityLabel;
 };
 
+struct CKButtonTitle {
+  CKButtonTitle() : title(nil), attributedTitle(nil) {};
+  CKButtonTitle(std::nullptr_t v) : title(nil), attributedTitle(nil) {};
+  CKButtonTitle(NSString *t) : title(t), attributedTitle(nil) {};
+  CKButtonTitle(NSAttributedString *t) : title(nil), attributedTitle(t) {};
+
+  NSString *title;
```
ocrickard: nit: Since title is already in the name of the struct, I would tend to name these after their types, so string and attributedString. A little less redundant I think? When I see struct.title.title, I'd have to double-check that the type of title is a string...

ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -228,14 +237,14 @@ static T valueForState(const std::unordered_map<UIControlState, T> &m, UIControl
 }
 
 #if !TARGET_OS_TV // sizeWithFont is not available on tvOS
-static CGSize intrinsicSize(NSString *title, UIFont *titleFont, UIImage *image,
+static CGSize intrinsicSize(CKButtonTitle title, UIFont *titleFont, UIImage *image,
                             UIImage *backgroundImage, UIEdgeInsets contentEdgeInsets)
 {
   // This computation is based on observing [UIButton -sizeThatFits:], which uses the deprecated method
   // sizeWithFont in iOS 7 and iOS 8
```
ocrickard: We should check if this is still true in iOS 9/10.

ocrickard:
This looks pretty good to me. Any chance we could get some snapshot tests for this?

conradev:
Update on this: I implemented screenshot tests, and the size calculation is wrong because I was seeing some truncated text.

Will open IDA and figure it out when I get a chance 👍

eczarny:
Closing due to inactivity. @conradev feel free to reopen once you're able to get back to this (looks like this PR is still a bit of a WIP to address the sizing issue you mentioned).

PR #717, opened 2017-01-09
Fix unused static function warning

ocrickard:
When not in debug mode this static function was never called. Let's just move the #if to not compile it when it's not called. I built both for debug and release to confirm that the build doesn't break.


eczarny:
@facebook-github-bot shipit

ocrickard:
@facebook-github-bot import

ocrickard:
This landed internally. Merging.

PR #718, opened 2017-01-10
Fix next behavior set for legacy send function

ocrickard:


eczarny:
@facebook-github-bot shipit

ocrickard:
Landed internally. Merging.

PR #721, opened 2017-01-13
Add CKComponentViewContext fields documentation

zats:
This adds basic documentation clarifying `frame` coordinates to avoid potential confusion.

### DIFF
--- a/ComponentKit/Core/CKComponent.h
+++ b/ComponentKit/Core/CKComponent.h
@@ -18,7 +18,18 @@
 #import <ComponentKit/CKComponentViewConfiguration.h>
 
 struct CKComponentViewContext {
+  /**
+   View managed by the component.
+   */
   UIView *view;
+  /**
+   Frame of the component in view coordinates.
+
+   @example To calculate component frame in the corresponding window coordinates you can do
+
+      [c.viewContext.view convertRect:c.viewContext.frame toView:nil];
+
+   */
   CGRect frame;
 };
 


ComponentKit/Core/CKComponent.h
```diff
@@ -18,7 +18,18 @@
 #import <ComponentKit/CKComponentViewConfiguration.h>
 
 struct CKComponentViewContext {
+  /**
+   View managed by the component.
+   */
   UIView *view;
+  /**
+   Frame of the component in view coordinates.
```
ocrickard: I'd prefer to be more descriptive here than provide the example.

ocrickard:
If we're gonna add a comment, I think we should go a little deeper. I don't feel too strongly about it though, what do you think?

zats:
I feel like existent comment covers what caused my confusion, but I'd love to add more in depth information if I'd be more familiar with the framework
Any suggestions?

[Sash Zats - Contact Using Hop](http://GetHop.com/?_hmid=1484680946)

On January 17, 2017 at 19:19 GMT, Oliver Clark Rickard <notifications@github.com> wrote:

If we're gonna add a comment, I think we should go a little deeper. I don't feel too strongly about it though, what do you think?

eczarny:
I'm not really sure how much value this documentation adds to what's already available. Maybe instead we should move the more descriptive information in the `-[CKComponent viewContext]` to the `CKComponentViewContext` declaration instead? e.g. high level documentation on `-[CKComponent viewContext]` with more details on `CKComponentViewContext`. That way if somebody finds `CKComponentViewContext` first they'll have a good idea what it is and how it should be used instead of just finding an undocumented struct.

zats:
My thinking was that context documentation should be on the element that might cause confusion. 
As a broader effort to increase codebase documentation coverage, I agree with what you said.

zats:
Nope, I guess it's okay to close it if we are not planing on adding this

PR #724, opened 2017-01-15
Removed copy method calls from quote.m

AdamRobertsEF:
These get copied anyway.

### DIFF
--- a/Examples/WildeGuess/WildeGuess/Quote.h
+++ b/Examples/WildeGuess/WildeGuess/Quote.h
@@ -15,9 +15,9 @@
 
 @interface Quote : NSObject
 
-@property (nonatomic, readonly, strong) NSString *text;
-@property (nonatomic, readonly, strong) NSString *author;
-@property (nonatomic, readonly, assign) QuoteDisplayStyle style;
+@property (nonatomic, readonly, copy) NSString *text;
+@property (nonatomic, readonly, copy) NSString *author;
+@property (nonatomic, readonly) QuoteDisplayStyle style;
 
 - (instancetype)initWithText:(NSString *)text
                       author:(NSString *)author
--- a/Examples/WildeGuess/WildeGuess/Quote.h
+++ b/Examples/WildeGuess/WildeGuess/Quote.h
@@ -15,9 +15,9 @@
 
 @interface Quote : NSObject
 
-@property (nonatomic, readonly, copy) NSString *text;
-@property (nonatomic, readonly, copy) NSString *author;
-@property (nonatomic, readonly) QuoteDisplayStyle style;
+@property (nonatomic, readonly, strong) NSString *text;
+@property (nonatomic, readonly, strong) NSString *author;
+@property (nonatomic, readonly, assign) QuoteDisplayStyle style;
 
 - (instancetype)initWithText:(NSString *)text
                       author:(NSString *)author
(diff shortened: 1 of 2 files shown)


conradev:
The existing code seems correct to me because the arguments don't get copied automatically when using instance variable assignment.

AdamRobertsEF:
You're correct that they don't get copied. Although they don't need to be copied as they get retained. 

Let's say ViewController creates two NSStrings called stringOne and stringTwo and passes these objects into Quote, these objects get retained by Quote. If ViewController changes the values of either stringOne, then a new NSString object is created for stringOne in ViewController, the previous object is released, and the instance of Quote becomes the owner.

NSString aren't mutable, when changing the value a new object is always created. Calling copy wastes cpu time and memory. If we don't use copy, then if the same NSString value is set in multiple instances of Quote, then the same object will be used.

conradev:
Yes! That all makes sense. 

> If we don't use copy, then if the same NSString value is set in multiple instances of Quote, then the same object will be used

`-[NSString copy]` is actually a no-op (`return self`), so even if we do copy the string, the same object is still used. It isn't a no-op for mutable subclasses like `NSMutableString`, though.

I assume the properties were made `copy` to protect against this situation:

```objc
NSMutableString *author = [@"Conrad" mutableCopy];
Quote *quote = [[Quote alloc] initWithText:@"lol" author:author …];
[author appendString:@" Kramer"];
```

AdamRobertsEF:
I've created a test project and a similar class to Quote, the screenshots below show the output, and the code that I've used. This confirms my statement.

screen shot 2017-01-17 at 07 59 36

screen shot 2017-01-17 at 08 02 58

screen shot 2017-01-17 at 08 03 12

screen shot 2017-01-17 at 08 05 14

AdamRobertsEF:
@conradev I've added a check for mutable strings and copies are only made when a NSMutableString is passed, if a NSString is passed it's retained; thanks for your input.

ocrickard:
Would you mind explaining why you want to do this a little bit more? Is there a bug somewhere? This is common practice in most large Obj-C codebases.

ocrickard:
Closing this. If there's something that I'm missing, happy to re-open.

AdamRobertsEF:
In this example project there will be little performance hit, given that this has a very small data set.

It's something to bear in mind when working with large amounts of data, especially if it's cached locally, and frequently repopulating.

Common practice isn't optimal. There are always ways to improve, even if you cannot see much benefit initially.

Thanks
Adam Roberts FIAP

ocrickard:
Hmm, even with a large amount of data, your code is identical in runtime performance. The design of the NSMutableCopying Foundation class clusters is that the immutable variant no-ops on copy, as @conradev notes. You only incur a cost if the input is mutable, but this proposed solution still copies in that case, so is identical.

Anyway, that discussion is somewhat orthogonal to the changes here. This is in an example project that is small, dealing with strings, and if copying mutable strings is what's slowing down your project, then a re-architect is likely in need anyway.

ComponentKit takes a particularly strong approach to immutability because it is a multithreaded UI framework, so where in certain other places you may want to make a tradeoff and not copy a huge mutable array of data (again, I'd suggest re-architecting, but still), we'd still be in favor of copying everywhere to enforce that every object that is shared among threads truly has its own, immutable copy of the data.

PR #710, opened 2016-12-27
Revert "Merge pull request #693 from ocrickard/pending-reuse-pool"

ocrickard:
This reverts commit 42cb8f37f9eecd60c513c3dfc2f2de2bb7b9e522, reversing
changes made to 9df31ebdbd78c8d6b72f48cb93a56738360d1296.

While we're reviewing fixes to stateful views, we decided to revert this diff during the holidays.

Reverts:
https://github.com/facebook/componentkit/pull/693

PR #731, opened 2017-01-23
Bring back the pending reuse pool

ocrickard:
Reverts https://github.com/facebook/componentkit/pull/710, and adds back in https://github.com/facebook/componentkit/pull/709/files, which should fix the issues.


ocrickard:
This merged internally. Merging.

PR #735, opened 2017-01-28
Add noexcept specifications

swolchok:
From the first diff in the stack:

Presents a guarantee to the compiler that these functions will not
throw. For reviewers' reference, the consequence of violating a
`noexcept` specification is that `std::terminate` is called.

`NSAssert` seems to throw on error, and `CKAssert` wraps `NSAssert`,
so I considered using `noexcept(!CK_ASSERTIONS_ENABLED)` on those
functions that use `CKAssert`. However, local testing revealed that
`NSAssert` failure in a `noexcept` function doesn't result in an
unusable exception backtrace, at least on Mac. (It doesn't seem to
change the exception crash message at all.)

I did a profile build of WildeGuess for simulator (having difficulty
building for device right now) and observed a 1.4% code size
improvement with this diff, which seems to support my hypothesis that
these `noexcept` specifications provide additional optimization
opportunities.

### DIFF
--- a/ComponentKit/Core/CKComponentAnimation.h
+++ b/ComponentKit/Core/CKComponentAnimation.h
@@ -34,10 +34,10 @@ struct CKComponentAnimation {
    @param animation A CAAnimation to apply on the component's layer.
    @param layerPath A key path to a sublayer of the component's view. Defaults to nil.
    */
-  CKComponentAnimation(CKComponent *component, CAAnimation *animation, NSString *layerPath = nil);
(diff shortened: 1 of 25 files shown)


eczarny:
@facebook-github-bot import

eczarny:
Sounds promising! Thanks for the PR @swolchok. I'm importing your changes to kick of our internal builds.

@ocrickard seems like we may have a reasonable code size win on our hands.

ocrickard:
Soooo excited for this code-size win. Thanks @swolchok! I'm pretty onboard with this if we can show the size of the impact on the builds.

eczarny:
I'll land the internal change then merge this once that's successful.

PR #737, opened 2017-02-01
Only disturb the collection view if a cell changes size

conradev:
I have another change for `-[CKCollectionViewTransactionalDataSource transactionalComponentDataSource:didModifyPreviousState:byApplyingChanges:]`.

The gist here is that if I change the background color of a component, or otherwise change the content without changing the size, it shouldn't call `-[UICollectionView performBatchUpdates:completion:]`. Calling that method invalidates the `UICollectionViewLayout`, which causes the layout to be recomputed. This change makes it so that simple state updates should be a lot faster.

Let me know if I need to make any changes!

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -30,6 +30,8 @@ CKTransactionalComponentDataSourceListener
   __weak id<CKSupplementaryViewDataSource> _supplementaryViewDataSource;
   CKTransactionalComponentDataSourceState *_currentState;
   CKComponentDataSourceAttachController *_attachController;
+  NSMapTable<UICollectionViewCell *, NSIndexPath *> *_cellToIndexPathMap;
+  NSMapTable<NSIndexPath *, UICollectionViewCell *> *_indexPathToCellMap;
   NSMapTable<UICollectionViewCell *, CKTransactionalComponentDataSourceItem *> *_cellToItemMap;
 }
 @end
@@ -52,6 +54,8 @@ CKTransactionalComponentDataSourceListener
     
     _attachController = [[CKComponentDataSourceAttachController alloc] init];
     _supplementaryViewDataSource = supplementaryViewDataSource;
+    _cellToIndexPathMap = [NSMapTable weakToStrongObjectsMapTable];
+    _indexPathToCellMap = [NSMapTable strongToWeakObjectsMapTable];
     _cellToItemMap = [NSMapTable weakToStrongObjectsMapTable];
   }
   return self;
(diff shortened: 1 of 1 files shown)


conradev:
Hold up on reviewing this, I found a few issues while testing it out.

conradev:
Fixed the issues. Should probably write tests for this functionality

eczarny:
Any updates on the tests @conradev? In the meantime I'll try and get around to reviewing this later today or tomorrow.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -114,16 +118,32 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
...
+      CKTransactionalComponentDataSourceItem *oldItem = [_currentState objectAtIndexPath:indexPath];
+      CKTransactionalComponentDataSourceItem *newItem = [state objectAtIndexPath:indexPath];
+      sizeChanged = !CGSizeEqualToSize(oldItem.layout.size, newItem.layout.size);
+      if (sizeChanged)
+        break;
+    }
+    
     void (^applyUpdatedState)(CKTransactionalComponentDataSourceState *) = ^(CKTransactionalComponentDataSourceState *updatedState) {
-      [_collectionView performBatchUpdates:^{
+      if (sizeChanged) {
```
eczarny: This check doesn't seem quite right. What if there are no size changes in any of the updates but there are insertions or removals? I imagine we would also want to check `changesIncludeNonUpdates`:

```
if (changesIncludeNonUpdates || (changesIncludeOnlyUpdates && updatesWithBoundsChange)) {
```

You could probably get way with omitting the `changesIncludeOnlyUpdates` check but I think being explicit is helpful.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -114,16 +118,32 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
         break;
     }
     
+    // If none of the cells changed size, we can remount the updated cells directly
+    // without notifying the collection view
+    BOOL sizeChanged = NO;
```
eczarny: I think `updatesWithBoundsChange` is more fitting.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -138,7 +158,15 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
     CKComponentBoundsAnimationApply(boundsAnimation, ^{
       for (NSIndexPath *indexPath in changes.updatedIndexPaths) {
         CKTransactionalComponentDataSourceItem *item = [state objectAtIndexPath:indexPath];
+          
+        // There is a race condition that causes this method to return nil
+        // between when the cell is first requested and when the cell is
+        // placed in the collection view. Thus we fall back on our mapping.
```
eczarny: Great comment.

I'd like to understand the problem a bit more. Is this issue orthogonal to your optimization or is it something you noticed as a result of the optimization? Just trying to gauge whether or not I need to consider these changes together or if I can think about them separately.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -138,7 +158,15 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
...
         CKCollectionViewDataSourceCell *cell = (CKCollectionViewDataSourceCell *)[_collectionView cellForItemAtIndexPath:indexPath];
+        if (!cell) {
+          cell = (CKCollectionViewDataSourceCell *)[_indexPathToCellMap objectForKey:indexPath];
+        }
```
eczarny: I'd prefer we avoid mutating `cell`:

```
CKCollectionViewDataSourceCell *const cell = (CKCollectionViewDataSourceCell *)([_collectionView cellForItemAtIndexPath:indexPath] ?: [_indexPathToCellMap objectForKey:indexPath]);
```

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -138,7 +158,15 @@ - (void)transactionalComponentDataSource:(CKTransactionalComponentDataSource *)d
...
+
```
eczarny: Nit: I'd drop the newline.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
conradev: This entire block is inside of a `if (changesIncludeOnlyUpdates) { … }`

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
conradev: This is something I noticed as a result of the optimization. If I issue two updates in quick succession, i.e. an insert and then an update, the update would not find a cell to mount, and silently fail.

Previously, updates were only performed after calling `-[UICollectionView performBatchUpdates:completion:]`, after which the issue doesn't seem to occur.

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
eczarny: Oops, missed that. Thanks for clarifying @conradev!

eczarny:
Closing due to inactivity. Feel free to reopen @conradev once you are able to include some tests.

PR #738, opened 2017-02-03
Explicitly instantiate some commonly used templates.

yiding:
Doing so prevents instantiations of methods of these templates in every
compilation unit that use them.

These templates were picked empirically based on statistics of which
symbols are duplicated across object files. This reduces the size of
generated static library archives in a large app by 3.6%.

One can examine duplicate symbols in object files by doing something like this:

```
find buck-out/gen -name '*.a' | xargs -L1 objdump -t | grep ' gw ' | grep '__TEXT,__text' | c++filt | cut -f2 |  sort | uniq -c | sort -n
```


eczarny:
@facebook-github-bot import

eczarny:
Thanks @yiding! I'l get this merged in today.

ocrickard:
Fantastic, thanks!

PR #740, opened 2017-02-04
Speculative fix for duplicate stateful views in reuse pool

ocrickard:
@conradev this is my stab at fixing the issue you reported. Mind patching it and testing if it fixes your issue?

### DIFF
--- a/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
+++ b/ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
@@ -131,13 +131,13 @@
 
 - (void)_relinquishStatefulViewIfPossible
 {
-  if ([self canRelinquishStatefulView]) {
+  if ([self canRelinquishStatefulView] && _statefulView) {
     [[CKStatefulViewReusePool sharedPool]
      enqueueStatefulView:_statefulView
      forControllerClass:[self class]
      context:_statefulViewContext
      mayRelinquishBlock:^BOOL{
-       if (!self->_mounted && [self canRelinquishStatefulView]) {
+       if (self->_statefulView && !self->_mounted && [self canRelinquishStatefulView]) {
          [self willRelinquishStatefulView:self->_statefulView];
          self->_statefulView = nil;
          self->_statefulViewContext = nil;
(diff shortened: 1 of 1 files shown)


ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
```diff
@@ -131,13 +131,13 @@ - (void)_presentStatefulView
...
 {
-  if ([self canRelinquishStatefulView]) {
+  if ([self canRelinquishStatefulView] && _statefulView) {
     [[CKStatefulViewReusePool sharedPool]
      enqueueStatefulView:_statefulView
      forControllerClass:[self class]
      context:_statefulViewContext
      mayRelinquishBlock:^BOOL{
-       if (!self->_mounted && [self canRelinquishStatefulView]) {
+       if (self->_statefulView && !self->_mounted && [self canRelinquishStatefulView]) {
```
ocrickard: Here's the issue: _relinquishStatefulViewIfPossible can be called multiple times. I missed that the previous version of this code was using the `if (_statefulView)` check as a way to guarantee that only one block won and actually enqueued the view into the reuse pool.

So what would happen here is that the stateful view could be enqueued into the reuse pool as many times as this method was called, which could result in a stateful view being used for more than one controller at a time.

ComponentKit/StatefulViews/CKStatefulViewComponentController.mm
```diff
@@ -131,13 +131,13 @@ - (void)_presentStatefulView
 
 - (void)_relinquishStatefulViewIfPossible
 {
-  if ([self canRelinquishStatefulView]) {
+  if ([self canRelinquishStatefulView] && _statefulView) {
```
conradev: Might want to flip this boolean check

ocrickard: Sure, good point.

ocrickard:
Confirmed with @conradev that this fixes the issue that he's seeing. Since it's safe, I'd like to merge this.

ocrickard:
@facebook-github-bot import

ocrickard:
Landed internally, build is green. Merging.

PR #741, opened 2017-02-06
RFC: Component triggers

ocrickard:
This is a concept that we've been talking about for awhile. I had a need for this in one of my other projects, and decided to prototype the implementation. I think the documentation speaks for itself, so check out the header.

I wanted feedback on the approach before I write all the unit tests.

### DIFF
--- a/ComponentKit/Utilities/CKComponentTrigger.h
+++ b/ComponentKit/Utilities/CKComponentTrigger.h
@@ -77,7 +77,7 @@
    
      - (void)activateTrigger
      {
-       _handle->invoke(YES);
+       _handle->invoke(self, YES);
      }
  */
 
--- a/ComponentKit/Utilities/CKComponentTrigger.h
+++ b/ComponentKit/Utilities/CKComponentTrigger.h
@@ -1,0 +1,169 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <memory>
+
+#import <ComponentKit/CKComponentScope.h>
+#import <ComponentKit/CKComponentAction.h>
+
+/**
+ A component trigger is the opposite of an action. It is used by a parent component to trigger some side effect in a
+ child component.
+ 
+ Similar to an action, it is created in a parent scope, and is threaded down to a child component. Unlike an action
+ though, triggers activate **downward** from a parent component onto a child. Actions are triggered by a child, and
+ cause side-effects in parent components.
+ 
+ Triggers are an alternative to manually exposing methods on parent components that invoke on child components. Instead
+ triggers can be passed through many layers of components.
+ 
+ Triggers use the same pattern as component actions to allow parameters to be passed to targets. The first argument to
+ a trigger selector will always be the sender of the trigger invocation, the next arguments types are determined by
+ the variadic template types. This allows passing of data down to the trigger invocation method.
+ 
+ Triggers passed into a component **must** be resolved once, and only once. Resolution assigns a target and selector
+ with the trigger for invocation. Here's an example:
+ 
+     + (instancetype)newWithTriggerHandle:(CKTypedComponentTriggerHandle<BOOL>)someTriggerHandle
+     {
+       CKComponentScope scope(self);
+       if (someTriggerHandle) {
+         someTriggerHandle.resolve(scope, @selector(someMethodWithSender:booleanParameter:));
+       }
+       return [super new...];
+     }
+ 
+ Component triggers may be resolved in three different ways:
+ 
+   1. (Preferred) Scope resolution. Allows a CKComponent to register itself or its controller with the trigger.
+      If the selector is not implemented on the component itself, it will be invoked on the controller associated with
+      that component. At runtime in DEBUG mode the trigger will validate that your component or controller implements
+      the selector, and that the parameter types match.
+ 
+       CKComponentScope scope(self);
+       someTriggerHandle.resolve(scope, @selector(someMethodWithSender:booleanParameter:));
+
+   2. (Discouraged) Target resolution. Allows an arbitrary object to receive a trigger invocation. This object is weakly
+      held by the trigger.
+ 
+       Foo *foo = ...;
+       someTriggerHandle.resolve(foo, @selector(someMethodWithSender:booleanParameter:));
+ 
+   3. No-op resolution. Since triggers *must* be resolved, this allows a component to resolve the trigger with no
+      target. This means when the trigger is invoked, nothing will be called.
+ 
+ Triggers are created in parent scopes through acquiring a CKTypedComponentTriggerHandle.
+
+     + (instancetype)new
+     {
+       CKTypedComponentTriggerHandle<BOOL> handle = CKTypedComponentTriggerHandle<BOOL>::acquire();
+       FooComponent *c = [super newWithComponent:
+                          [BarComponent
+                           newWithTriggerHandle:handle]];
+       if (c) {
+         c->_handle = handle;
+       }
+       return c;
+     }
+   
+     - (void)activateTrigger
+     {
+       _handle->invoke(YES);
+     }
+ */
+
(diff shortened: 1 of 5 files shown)


ComponentKit/Utilities/CKComponentTrigger.h
```diff
@@ -0,0 +1,169 @@
...
+ cause side-effects in parent components.
+ 
+ Triggers are an alternative to manually exposing methods on parent components that invoke on child components. Instead
+ triggers can be passed through many layers of components.
+ 
+ Triggers use the same pattern as component actions to allow parameters to be passed to targets. The first argument to
+ a trigger selector will always be the sender of the trigger invocation, the next arguments types are determined by
+ the variadic template types. This allows passing of data down to the trigger invocation method.
+ 
+ Triggers passed into a component **must** be resolved once, and only once. Resolution assigns a target and selector
```
eczarny: > Triggers passed into a component **must** be resolved once, and only once.

I'm a little concerned by this restriction. Any descendant of the parent can resolve a trigger. The runtime assertion helps but at the end of the day I feel like this may be a common issue. Worst case triggers for particular components may just stop working if another descendant resolves it instead. In that case which one should actually "win"?

ComponentKit/Utilities/CKComponentTrigger.h
```diff
@@ -0,0 +1,169 @@
...
+ with the trigger for invocation. Here's an example:
+ 
+     + (instancetype)newWithTriggerHandle:(CKTypedComponentTriggerHandle<BOOL>)someTriggerHandle
+     {
+       CKComponentScope scope(self);
+       if (someTriggerHandle) {
+         someTriggerHandle.resolve(scope, @selector(someMethodWithSender:booleanParameter:));
+       }
```
eczarny: Can we avoid the nil check boilerplate somehow?

ComponentKit/Utilities/CKComponentTrigger.h
```diff
@@ -0,0 +1,169 @@
...
+      the selector, and that the parameter types match.
+ 
+       CKComponentScope scope(self);
+       someTriggerHandle.resolve(scope, @selector(someMethodWithSender:booleanParameter:));
+
+   2. (Discouraged) Target resolution. Allows an arbitrary object to receive a trigger invocation. This object is weakly
+      held by the trigger.
+ 
+       Foo *foo = ...;
+       someTriggerHandle.resolve(foo, @selector(someMethodWithSender:booleanParameter:));
```
eczarny: If this is discouraged why even offer it as an option?

ComponentKit/Utilities/CKComponentTrigger.h
```diff
@@ -0,0 +1,169 @@
...
+ 
+   3. No-op resolution. Since triggers *must* be resolved, this allows a component to resolve the trigger with no
+      target. This means when the trigger is invoked, nothing will be called.
```
eczarny: Can you provide an example when this form of resolution would be necessary? I feel like a trigger must always be properly resolved, otherwise the contract with the parent invoking the trigger is broken.

ocrickard: I think that's fair. @adamjernst mentioned wanting this. I agree I would prefer to remove it.

ComponentKit/Utilities/CKComponentTrigger.h
ocrickard: Sure, we could. I was providing a symmetric API with component actions, but considering the use-case is different, I'm cool with ignoring this case.

Logic becomes simpler if I do that! Also binary bloat gets smaller.

ComponentKit/Utilities/CKComponentTrigger.h
ocrickard: Yeah, I can make it just no-op.

ComponentKit/Utilities/CKComponentTrigger.h
ocrickard: Yeah, totally fair. We talked about providing a `fork` operator that parent components can call before passing the handle into more than one child. I'm not sure what else we can do here though with the assertion. I think we want the "one and only one" contract, and if you want more than that then you should have to manually duplicate the handle.

eczarny: Yeah, I think this is the biggest challenge here. For the time being this may be the best we can do, as long as we can provide fairly helpful context in the assertion. Can we maybe call out where else resolution took place in the event of a "collision"?

ComponentKit/Utilities/CKComponentTrigger.h
adamjernst: nuke it! i agree let's wait til it's needed

ComponentKit/Utilities/CKComponentTrigger.h
adamjernst: yeah i agree!

ComponentKit/Utilities/CKComponentTrigger.h
GotEmB: Is the reason for wanting to impose this restriction a solution to the race condition / state-update clash that we were talking about with signals?

To me, I'm feature comparing this to signals and one thing that I was looking into building is view attribute bindings (with the help of the other diff for supporting any attribute values).

Basically if we can bind a signal / component trigger as a value directly to a view / layer attribute selector, we'll be able to manipulate view / layer properties with the signal / trigger without touching the component tree.

ocrickard: I'm not really sure what that issue is @GotEmB. I'd love to hear more though.

GotEmB: @ocrickard, this is for binding a component trigger to a view property as I talked about on the internal note. Basically that'd mean a single producer and several consumers.

ocrickard:
Ping can I get a review here @eczarny @adamjernst @GotEmB

ComponentKit/Utilities/CKComponentTriggerInternal.h
```diff
@@ -0,0 +1,81 @@
...
+
+  friend class CKComponentTriggerBase;
+};
+
+template <typename... T>
+class CKTypedComponentTrigger;
+
+template <typename... T>
+class CKTypedComponentTriggerTarget : public CKComponentTriggerTargetBase {
+  static_assert(std::is_same<
```
GotEmB: If we use an IMP rather than an invocation (like in #747), we'll be able to send in any type of values (ex: `std::vector<int>`).

ComponentKit/Utilities/CKComponentTrigger.h
```diff
@@ -0,0 +1,101 @@
...
+CKTypedComponentTrigger<T...> CKComponentTriggerAcquire()
+{
+  return CKTypedComponentTrigger<T...>(std::static_pointer_cast<CKComponentTriggerTargetBase>(std::make_shared<CKTypedComponentTriggerTarget<T...>>()));
+}
+
+/** Convenience typedefs for triggers that have no arguments. */
+typedef CKTypedComponentTrigger<> CKComponentTrigger;
+
+extern template class CKTypedComponentTrigger<>;
+extern template class CKTypedComponentTrigger<id>;
```
GotEmB: I'm guessing that you'll be explicitly instantiating this template with common types like `int`, `float`, ... before this gets merged.

ComponentKit/Utilities/CKComponentTriggerInternal.h
ocrickard: Yeah, I kind of like this just using the same as actions, and for actions we have specified that you may specify as many parameters as you choose.

ComponentKit/Utilities/CKComponentTrigger.h
ocrickard: Yeah, we could.

ComponentKit/Utilities/CKComponentTriggerInternal.h
GotEmB: Cool, we can revisit this for both actions and triggers together separately.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ocrickard:
We've come to a consensus that this is something we'd like to avoid. It's always possible to implement a method on a component, or pass in an announcer to achieve the same result with more friction. We don't think removing the friction here is a good idea since it encourages people to avoid doing react state updates in favor of direct calls on triggers.

PR #742, opened 2017-02-08
Don't trip debug asserts when the selector of a CKComponentAction is NULL

conradev:
Running the action is a no-op, so checking the validity of it should be too.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -242,6 +242,10 @@ CKComponentViewAttributeValue CKComponentActionAttribute(CKTypedComponentAction<
 #if DEBUG
 static void checkMethodSignatureAgainstTypeEncodings(SEL selector, NSMethodSignature *signature, const std::vector<const char *> &typeEncodings)
 {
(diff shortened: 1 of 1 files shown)


ocrickard:
Looks good to me.

ocrickard:
@facebook-github-bot shipit

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -242,6 +242,9 @@ - (void)handleControlEventFromSender:(UIControl *)sender withEvent:(UIEvent *)ev
 #if DEBUG
 static void checkMethodSignatureAgainstTypeEncodings(SEL selector, NSMethodSignature *signature, const std::vector<const char *> &typeEncodings)
 {
+  if (selector == NULL)
```
ocrickard: Argh, wait wait wait. Sorry, I missed that there's a missing {} here. We strictly include {} everywhere in our codebase to avoid things like goto fail; goto fail; :P

conradev:
Ah, yes, oops – fixed!

ocrickard:
@facebook-github-bot import

PR #743, opened 2017-02-09
Fix logic flaw in CKCollectionViewTransactionalDataSource

conradev:
If a changeset was applied with an update and a remove positioned before that update, the wrong components would be mounted on the cells. I'm actually surprised that this bug has existed for so long!

I wrote a fix for it and an associated test that now passes. `CKCollectionViewTransactionalDataSourceTests` currently punches a hole through the test coverage...

I'm not sure if my proposed fix is the best one, but it works. I can write tests for `-newIndexPathForPreviousIndexPath:` if we decide to go with it , but I am also open to better suggestions 😄

### DIFF
--- a/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
+++ b/ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
@@ -71,12 +71,14 @@ CKTransactionalComponentDataSourceListener
 static void applyChangesToCollectionView(UICollectionView *collectionView,
                                          CKComponentDataSourceAttachController *attachController,
                                          NSMapTable<UICollectionViewCell *, CKTransactionalComponentDataSourceItem *> *cellToItemMap,
-                                         CKTransactionalComponentDataSourceState *currentState,
+                                         CKTransactionalComponentDataSourceState *state,
                                          CKTransactionalComponentDataSourceAppliedChanges *changes)
 {
-  [changes.updatedIndexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, BOOL *stop) {
-    if (CKCollectionViewDataSourceCell *cell = (CKCollectionViewDataSourceCell *) [collectionView cellForItemAtIndexPath:indexPath]) {
-      attachToCell(cell, [currentState objectAtIndexPath:indexPath], attachController, cellToItemMap);
+  NSDictionary<NSIndexPath *, NSIndexPath *> *mapping = CKComputeFinalUpdatedIndexPathsForAppliedChanges(changes);
+  [mapping enumerateKeysAndObjectsUsingBlock:^(NSIndexPath *indexPathBeforeUpdate, NSIndexPath *indexPathAfterUpdate, BOOL *stop) {
+    CKCollectionViewDataSourceCell *cell = (CKCollectionViewDataSourceCell *)[collectionView cellForItemAtIndexPath:indexPathBeforeUpdate];
+    if (cell) {
+      attachToCell(cell, [state objectAtIndexPath:indexPathAfterUpdate], attachController, cellToItemMap);
     }
   }];
   [collectionView deleteItemsAtIndexPaths:[changes.removedIndexPaths allObjects]];
--- a/ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
+++ b/ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
@@ -11,7 +11,6 @@
 #import <objc/runtime.h>
 #import "CKComponentInternal.h"
 
-#import "CKComponentDataSourceAttachController.h"
 #import "CKComponentDataSourceAttachControllerInternal.h"
 
 @implementation CKComponentDataSourceAttachController
--- a/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
+++ b/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
@@ -33,3 +33,5 @@
                                  userInfo:(NSDictionary *)userInfo NS_DESIGNATED_INITIALIZER;
 
 @end
+
+extern NSDictionary<NSIndexPath *, NSIndexPath *> *CKComputeFinalUpdatedIndexPathsForAppliedChanges(CKTransactionalComponentDataSourceAppliedChanges *changes);
--- a/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
+++ b/ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
@@ -89,3 +89,76 @@
 }
 
 @end
+
+
+NSDictionary<NSIndexPath *, NSIndexPath *> *CKComputeFinalUpdatedIndexPathsForAppliedChanges(CKTransactionalComponentDataSourceAppliedChanges *changes)
+{
+  NSMutableDictionary<NSIndexPath *, NSIndexPath *> *mapping = [NSMutableDictionary new];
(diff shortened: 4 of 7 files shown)


ocrickard:
Defer to @eczarny on this one.

I'd love to just write a simple datasource that uses the IGListKit diffing algorithm before enqueueing changesets with the transactional datasource.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
```diff
@@ -48,6 +48,53 @@ - (instancetype)initWithUpdatedIndexPaths:(NSSet *)updatedIndexPaths
   return self;
 }
 
+- (NSIndexPath *)newIndexPathForPreviousIndexPath:(NSIndexPath *)indexPath
+{
+  if ([_removedIndexPaths containsObject:indexPath])
+    return nil;
```
eczarny: Please use braces, even for these simple checks:

```
if ([_removedIndexPaths containsObject:indexPath]) {
  return nil;
}
```

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.mm
```diff
@@ -48,6 +48,53 @@ - (instancetype)initWithUpdatedIndexPaths:(NSSet *)updatedIndexPaths
...
+  __block NSInteger section = indexPath.section;
+  NSInteger item = indexPath.item;
+  if ([_removedSections containsIndex:section])
+    return nil;
+
+  for (NSIndexPath *sourceIndexPath in _movedIndexPaths.keyEnumerator) {
+    if (sourceIndexPath.section == section && sourceIndexPath.item < indexPath.item) {
+      item--;
+    }
+  }
```
eczarny: Let's include comments explaining each step. Makes reading through this a bit easier.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
```diff
@@ -32,4 +32,6 @@
                        insertedIndexPaths:(NSSet *)insertedIndexPaths
                                  userInfo:(NSDictionary *)userInfo NS_DESIGNATED_INITIALIZER;
 
+- (NSIndexPath *)newIndexPathForPreviousIndexPath:(NSIndexPath *)indexPath;
```
eczarny: Instead of adding this as an instance method on `CKTransactionalComponentDataSourceAppliedChanges` could this be implemented as a standalone function with its own set of tests?

ComponentKitTests/TransactionalDataSource/CKCollectionViewTransactionalDataSourceTests.mm
```diff
@@ -14,44 +14,109 @@
...
   NSIndexPath *indexPath = [NSIndexPath indexPathForItem:3 inSection:4];
   NSString *viewKind = @"foo";
   
-  [[self.mockSupplementaryViewDataSource expect] collectionView:self.mockCollectionView viewForSupplementaryElementOfKind:viewKind atIndexPath:indexPath];
-  [self.dataSource collectionView:self.mockCollectionView viewForSupplementaryElementOfKind:viewKind atIndexPath:indexPath];
+  [[self.mockSupplementaryViewDataSource expect] collectionView:self.collectionView viewForSupplementaryElementOfKind:viewKind atIndexPath:indexPath];
+  [self.dataSource collectionView:self.collectionView viewForSupplementaryElementOfKind:viewKind atIndexPath:indexPath];
+}
+
+- (void)testRemovingAndUpdatingSimultaneously {
```
eczarny: Nit: Keep opening braces on their own lines.

eczarny:
> If a changeset was applied with an update and a remove positioned before that update, the wrong components would be mounted on the cells.

Thinking about this more I'm curious. Does applying updates last help?

conradev:
Applying updates last wouldn't work because `changes.updatedIndexPaths` is set of pre-update index paths. The update needs to apply new items to old index paths.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
conradev: It could be, yeah! I want to write tests either way. What do you think it should be called?

eczarny:
> Applying updates last wouldn't work because changes.updatedIndexPaths is set of pre-update index paths. The update needs to apply new items to old index paths.

That's a good point. Let me think on it a bit more. In the meantime let's move forward with what you have here.

eczarny:
@conradev let's move ahead with this fix. Thanks for all the help!

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceAppliedChanges.h
eczarny: Maybe something like this:

```
NSDictionary<NSIndexPath *, NSIndexPath *> *CKComputeFinalUpdatedIndexPathForAppliedChanges(CKTransactionalComponentDataSourceAppliedChanges *appliedChanges);
```

The result can be a dictionary from original updated index paths to the final updated index paths.

eczarny:
Back to you @conradev to move this logic out of `CKTransactionalComponentDataSourceAppliedChanges` and into a function.

conradev:
Pushed what I have, still need to write tests

ComponentKit/DataSources/CKCollectionViewTransactionalDataSource.mm
```diff
@@ -71,12 +71,14 @@ - (void)applyChangeset:(CKTransactionalComponentDataSourceChangeset *)changeset
...
                                          CKComponentDataSourceAttachController *attachController,
                                          NSMapTable<UICollectionViewCell *, CKTransactionalComponentDataSourceItem *> *cellToItemMap,
-                                         CKTransactionalComponentDataSourceState *currentState,
+                                         CKTransactionalComponentDataSourceState *state,
                                          CKTransactionalComponentDataSourceAppliedChanges *changes)
 {
-  [changes.updatedIndexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, BOOL *stop) {
-    if (CKCollectionViewDataSourceCell *cell = (CKCollectionViewDataSourceCell *) [collectionView cellForItemAtIndexPath:indexPath]) {
-      attachToCell(cell, [currentState objectAtIndexPath:indexPath], attachController, cellToItemMap);
+  NSDictionary<NSIndexPath *, NSIndexPath *> *mapping = CKComputeFinalUpdatedIndexPathsForAppliedChanges(changes);
```
eczarny: Nit: `mapping` -> `updatedIndexPathsToFinalUpdatedIndexPaths`. A bit more verbose but it's a bit more explicit.

ComponentKit/DataSources/Common/CKComponentDataSourceAttachController.mm
```diff
@@ -11,7 +11,6 @@
 #import <objc/runtime.h>
 #import "CKComponentInternal.h"
 
-#import "CKComponentDataSourceAttachController.h"
```
eczarny: Why remove this? Although it's imported indirectly from the internal header it's better to be explicit.

conradev: Makes sense! I can undo that. It's a personal habit (assuming internal headers import regular headers)

eczarny:
@conradev any updates on the tests? Would love to get this fix merged.

eczarny:
Closing due to inactivity. @conradev feel free to reopen once the tests are ready.

PR #746, opened 2017-02-10
Avoid inline construction of vector in parameter lists

yiding:
ComponentKit frequently uses Objective C++ methods that takes in lists as parameters. These parameters are usually declared as `const std::vector<Foo> &`. This is suboptimal for two reasons:

1. In cases where the parameters specified as an initializer list, this initializer list must be converted into a vector before being passed into this method. This incurs a (potentially inlined) vector construction and destruction operation that bloats the binary.
2. In cases where the parameter is passed in as a pre-constructed vector and stored, it must be copied due to the incoming vector being a mere const reference.

This change introduces `CKContainerWrapper`, which can be used as a drop-in replacement for methods which currently takes in `const vector<T> &`, but will not (in my testing at least) inline initializer-list to vector construction. It also allows a vector to be moved into the field, so that it can be later moved out, without declaring another method which takes `vector<T> &&` as a parameter.

Note that when used with maps, the unordered_map constructor is normally not inlined anyways, so this is less of a size win, but the argument for "free" move construction still applies.

Sample code generated using -Os for:
```
[CKStackLayoutComponent newWithView:... size:... style:... children:{item1, item2}
```

Before:
```
	xorpd	%xmm0, %xmm0
Ltmp184:
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/vector:432:7
	movapd	%xmm0, -624(%rbp)
	movq	$0, -608(%rbp)
Ltmp185:
	.## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/vector:1283:9
Ltmp68:
	leaq	-624(%rbp), %rdi
Ltmp186:
	##DEBUG_VALUE: vector:this <- %RDI
	##DEBUG_VALUE: vector:__il [bit_piece offset=64 size=64] <- 2
	movl	$2, %esi
	callq	std::__1::vector<CKStackLayoutComponentChild, std::__1::allocator<CKStackLayoutComponentChild> >::allocate(unsigned long)
Ltmp187:
Ltmp69:
## BB#33:
	leaq	-232(%rbp), %rbx
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/initializer_list:93:57
Ltmp188:
	leaq	-104(%rbp), %r12
Ltmp189:
	##DEBUG_VALUE: __construct_at_end<const CKStackLayoutComponentChild *>:__last <- %R12
	##DEBUG_VALUE: __construct_range_forward<const CKStackLayoutComponentChild *, CKStackLayoutComponentChild *>:__end1 <- %R12
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1586:56
	movq	-616(%rbp), %r14
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1501:57
Ltmp190:
	movq	%r14, %r15
Ltmp191:
LBB0_34:                                ## =>This Inner Loop Header: Depth=1
	##DEBUG_VALUE: construct<CKStackLayoutComponentChild, const CKStackLayoutComponentChild &>:__p <- %R15
	## buck-out/gen/VendorLib/ComponentKit/ComponentKit#headers,iphonesimulator-x86_64/ComponentKit/CKStackLayoutComponent.h:74:8
	movq	(%rbx), %rdi
	callq	*_objc_retain@GOTPCREL(%rip)
	movq	%rax, (%r15)
	movq	56(%rbx), %rax
	movq	%rax, 56(%r15)
	movq	48(%rbx), %rax
	movq	%rax, 48(%r15)
	movq	40(%rbx), %rax
	movq	%rax, 40(%r15)
	movq	32(%rbx), %rax
	movq	%rax, 32(%r15)
	movq	24(%rbx), %rax
	movq	%rax, 24(%r15)
	movq	8(%rbx), %rax
	movq	16(%rbx), %rcx
	movq	%rcx, 16(%r15)
	movq	%rax, 8(%r15)
Ltmp192:
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1585:29
	addq	$64, %rbx
Ltmp193:
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1585:13
	addq	$64, %r15
Ltmp194:
	cmpq	%r12, %rbx
	jne	LBB0_34
## BB#35:
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1586:56
Ltmp195:
	subq	$-128, %r14
	## APPLE_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/char++/v1/memory:1585:59
	movq	%r14, -616(%rbp)
```

After:
```
Ltmp187:
	## buck-out/gen/VendorLib/ComponentKit/ComponentKit#headers,iphonesimulator-x86_64/ComponentKit/CKParam.h:30:98
Ltmp68:
	leaq	-616(%rbp), %rdi
Ltmp188:
	##DEBUG_VALUE: VectorParam:this <- %RDI
	##DEBUG_VALUE: VectorParam:items [bit_piece offset=64 size=64] <- 2
	leaq	-232(%rbp), %rsi
	movl	$2, %edx
	callq	CK::VectorParam<CKStackLayoutComponentChild>::VectorParam(std::initializer_list<CKStackLayoutComponentChild>)
```

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -83,10 +83,10 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
   CGSize _intrinsicSize;
 }
 
-+ (instancetype)newWithTitles:(const std::unordered_map<UIControlState, NSString *> &)titles
-                  titleColors:(const std::unordered_map<UIControlState, UIColor *> &)titleColors
-                       images:(const std::unordered_map<UIControlState, UIImage *> &)images
-             backgroundImages:(const std::unordered_map<UIControlState, UIImage *> &)backgroundImages
++ (instancetype)newWithTitles:(CKContainerWrapper<std::unordered_map<UIControlState, NSString *>> &&)titlesParam
+                  titleColors:(CKContainerWrapper<std::unordered_map<UIControlState, UIColor *>> &&)titleColorsParam
+                       images:(CKContainerWrapper<std::unordered_map<UIControlState, UIImage *>> &&)imagesParam
+             backgroundImages:(CKContainerWrapper<std::unordered_map<UIControlState, UIImage *>> &&)backgroundImagesParam
                     titleFont:(UIFont *)titleFont
                      selected:(BOOL)selected
                       enabled:(BOOL)enabled
@@ -95,6 +95,10 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
 {
+  const auto titles = titlesParam.take();
+  const auto titleColors = titleColorsParam.take();
+  const auto images = imagesParam.take();
+  const auto backgroundImages = backgroundImagesParam.take();
   static const CKComponentViewAttribute titleFontAttribute = {"CKButtonComponent.titleFont", ^(UIButton *button, id value){
     button.titleLabel.font = value;
   }};
--- a/ComponentKit.xcodeproj/project.pbxproj
+++ b/ComponentKit.xcodeproj/project.pbxproj
@@ -376,6 +376,8 @@
 		B761C8AE1CB36BF700CDD03F /* CKTransactionalComponentDataSourceChangesetTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = B761C8AD1CB36BF700CDD03F /* CKTransactionalComponentDataSourceChangesetTests.mm */; };
 		B761C8B11CB36FA200CDD03F /* CKTransactionalComponentDataSourceAppliedChangesTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = B761C8B01CB36FA200CDD03F /* CKTransactionalComponentDataSourceAppliedChangesTests.mm */; };
 		B7E2E59F1D077098002A4442 /* CKVectorHelperTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = B7E2E59E1D077098002A4442 /* CKVectorHelperTests.mm */; };
+		CDCC9DD41E568C2C0005D52E /* CKContainerWrapper.h in Headers */ = {isa = PBXBuildFile; fileRef = CDCC9DD31E568C2C0005D52E /* CKContainerWrapper.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		CDCC9DD51E568C4B0005D52E /* CKContainerWrapper.h in Headers */ = {isa = PBXBuildFile; fileRef = CDCC9DD31E568C2C0005D52E /* CKContainerWrapper.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		D000B81D1CC45427005A27CF /* OCMock.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = D04977E31CC4145E0046CBCC /* OCMock.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		D000B81E1CC45440005A27CF /* FBSnapshotTestCase.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D000B8101CC41535005A27CF /* FBSnapshotTestCase.framework */; };
 		D000B81F1CC45446005A27CF /* FBSnapshotTestCase.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D000B8101CC41535005A27CF /* FBSnapshotTestCase.framework */; };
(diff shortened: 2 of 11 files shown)


ocrickard:
Wow, pretty sweet find here. Build is failing though.

ocrickard:
I took a brief look at the build failure, which is complaining:

> 'ComponentKit/CKParam.h' file not found

You likely need to change the header visibility of CKParam.h in the Xcode Project for both iOS and TVOS

yiding:
Yep, forgot to do that


________________________________

ocrickard:
Hmm, now this is broken for all builds.

yiding:
I seem to have not committed the xcode changes last time......

ComponentKit/Utilities/CKParam.h
```diff
@@ -0,0 +1,62 @@
...
+  private:
+    ContainerType _value;
+  };
+
+  /**
+   * \see VectorParam
+   */
+  template<class K, class V>
+  class MapParam {
+    typedef std::unordered_map<K, V> ContainerType;
```
adamjernst: is it feasible to make this templatized on ContainerType instead of simply hardcoding it?

i am not actually concerned with the flexibility that provides — i just think that for someone relatively familiar with C++ but unfamiliar with ComponentKit, this:

```
Wrapper<std::unordered_map<K, V>>
```

is easier to read than `WeirdWrapper<K, V>`

@eczarny thoughts?

ComponentKit/Utilities/CKParam.h
```diff
@@ -0,0 +1,62 @@
...
+   *   inlining helps reduce the size of the compiled output.
+   *
+   * Additionally, this type allows vector to be moved in, as if the parameter
+   *   has a rvalue-reference overload (e.g. `foo(vector<int> &&)`). In ObjC++
+   *   methods, where overloads are not allowed, this gives the benefit of a
+   *   rvalue-reference overload at the cost of an additional move operation.
+   */
+  template<class T>
+  class VectorParam {
+    typedef std::vector<T> ContainerType;
```
adamjernst: Same as my other comment.

ComponentKit/Utilities/CKParam.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <initializer_list>
+#include <vector>
+#include <unordered_map>
+
+namespace CK {
+  /**
+   * Helper class that's useful for function parameters that's likely to take
```
adamjernst: "that are likely"

ComponentKit/Utilities/CKParam.h
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * \see VectorParam
+   */
+  template<class K, class V>
+  class MapParam {
+    typedef std::unordered_map<K, V> ContainerType;
+  public:
+    MapParam() {}
+    MapParam(std::initializer_list<typename ContainerType::value_type> items) : _value(items) {}
```
adamjernst: why doesn't C++ simply inline this construction too? i'm confused.

yiding: For vectors, the constructors are declared with __attribute__((__always_inline_)). The wrapper cause the inlining to happen in the wrapper's constructor, which is not forcefully inlined.

unordered_map constructor (the main one, not the delegating ones) is not forcefully inlined in the aformentioned manner. The compiler, however, will instead inline this MapParam constructor, since its body is simply a call to the (not-forcefully inlined) unordered_map constructor, resulting in no additional indirection.

ComponentKit/Utilities/CKParam.h
yiding: I'm fine with doing it that way, since the map and vector wrappers have the same interface. We can always add a typedef for map and vector cases if we feel the name is excessively long. I'm not sure about just calling it `Wrapper` though, maybe `ContainerParam` or `ContainerWrapper`

adamjernst: `CKContainerWrapper`? Since we're in Objective-C++, we've gradually settled on CK-prefixing C++ classes. Originally we used C++ namespaces but that just looked too weird in Objective-C++.

ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -95,6 +95,10 @@ + (instancetype)newWithTitles:(const std::unordered_map<UIControlState, NSString
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
 {
+  std::unordered_map<UIControlState, NSString *> titles = titlesParam.take();
+  std::unordered_map<UIControlState, UIColor *> titleColors = titleColorsParam.take();
+  std::unordered_map<UIControlState, UIImage *> images = imagesParam.take();
+  std::unordered_map<UIControlState, UIImage *> backgroundImages = backgroundImagesParam.take();
```
swolchok: nit: why not use `auto`?

ComponentKit/Utilities/CKParam.h
```diff
@@ -0,0 +1,62 @@
...
+   */
+  template<class T>
+  class VectorParam {
+    typedef std::vector<T> ContainerType;
+  public:
+    VectorParam() {}
+    VectorParam(std::initializer_list<typename ContainerType::value_type> items) : _value(items) {}
+    VectorParam(ContainerType &&items) : _value(std::move(items)) {}
+    VectorParam(const ContainerType &items) : _value(items) {}
+    VectorParam(const VectorParam<T> &) = delete;
```
swolchok: I'm a little fuzzy on the exact rules here, despite just having read them at http://en.cppreference.com/w/cpp/language/move_assignment . Do you have to explicitly default the move assignment operator because of this deleted copy ctor?

yiding: I don't think it's required, I put it here for clarity.  As I understand it the intent of the move constructor rules is that trivial classes (i.e. POD and aggregates) can get move constructors without extra boilerplate and legacy code gets move constructors "for free". In this case I wish to explicitly state my intention of having this class be only movable and not copyable, so I explicitly specify the move constructor's existence, and the copy constructor's non-existence.

ComponentKit/Components/CKButtonComponent.mm
yiding: I don't mind the typing, but I can certainly do that if you think it improves readability.

yiding:
Updated to use `auto` in places, and made implementation fully generic (and usages more verbose).

ComponentKit/Components/CKButtonComponent.mm
```diff
@@ -95,6 +95,10 @@ + (instancetype)newWithTitles:(const std::unordered_map<UIControlState, NSString
                    attributes:(const CKViewComponentAttributeValueMap &)passedAttributes
    accessibilityConfiguration:(CKButtonComponentAccessibilityConfiguration)accessibilityConfiguration
 {
+  const auto titles = titlesParam.take();
```
adamjernst: eeeek, not sure i love this. what's the reason for not just having automatic conversion?

adamjernst:
Looks good to me, apart from `.take()` which I don't love but I trust you.

adamjernst:
will wait for Travis, then merge. Ping me if i forget

ComponentKit/Components/CKButtonComponent.mm
yiding: This operation moves the value out of the wrapped container in the same manner as `std::move` would.

A conversion that simply returns a const-reference makes it extremely easy to create an inadvertent copy. A conversion that actually moves the container is extremely confusing.

yiding:
This failure seems questionable, how do I tell it to go again?

eczarny:
@yiding I restarted the job.

yiding:
Looks good now.

yiding:
Ping?

eczarny:
@adamjernst? I can work with @yiding to merge this if you like.

adamjernst:
@eczarny i'm good with it (obviously the damage to readability is a bummer, but shrug).

eczarny:
@facebook-github-bot import

PR #747, opened 2017-02-10
Support custom types in view attributes

GotEmB:
Currently `CKComponentViewAttribute` uses `CKBoxedValue` to store values which only lets you store `id` types and values that can be wrapped as an `NSValue`.

This proposed change lets you store any kind of value (primitive, custom structs, id types, etc...) within `CKComponentViewAttribute` while remaining completely compatible with the current API.


ocrickard:
Rad. I'll try and review this over the weekend.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -74,52 +113,216 @@ struct CKComponentViewAttribute {
...
+      BOOL isEqualTo(const ValueBase &other) const
+      {
+        return this->typeName() == other.typeName() && CK::is_equal<T>()(_value, static_cast<const Value<T> &>(other)._value);
+      }
+
+      void performSetter(id object, SEL setter) const
+      {
+        const CachedSetter &set = CachedSetterInvocation(object, setter);
+        T value = _value;
+        [set.invocation setArgument:&value atIndex:2];
```
ocrickard: If you're using an invocation then the type should be trivially-constructible, right? What happens if I pass a std::vector<CGRect> here.

ocrickard:
Looks pretty good to me. My main concern is that I'm not sure how this works for more complex types. I'd love to see tests for those cases if they're supported, or as I suspect, they likely aren't supported, in which case a static_assert would be helpful.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -56,14 +69,41 @@ struct CKComponentViewAttribute {
    |        |          | Updater is called if the attribute was previously applied and the value changes.              |
    |--------|----------|-----------------------------------------------------------------------------------------------|
    */
+  template<typename T, typename U, typename = typename std::enable_if<std::is_convertible<T, id>::value>::type>
```
ocrickard: OK, so we're templatizing the view type? Seems nice.

ocrickard: nit: I prefer more descriptive type names for more than one template param.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -56,14 +69,41 @@ struct CKComponentViewAttribute {
...
-                           void (^upd)(id view, id oldValue, id newValue) = nil) :
+                           void (^app)(T view, U value),
+                           void (^unapp)(T view, U value),
+                           void (^upd)(T view, U oldValue, U newValue)) :
   identifier(ident),
-  applicator(app),
-  unapplicator(unapp),
-  updater(upd) {};
+  applicator([=](T view, U value) { app(view, value); }),
+  unapplicator(unapp ? ApplicatorFunc([=](T view, U value) { unapp(view, value); }) : ApplicatorFunc(nullptr)),
```
ocrickard: Why wrap these in a C++ lambda? You're using std::function right? std::function accepts blocks directly.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -74,52 +114,205 @@ struct CKComponentViewAttribute {
...
+  namespace ViewAttribute {
 
-private:
-  id __actual;
+    // Singleton object to signify non-ID-type object. For internal use only.
+    extern id nonIDObject;
 
-};
+    /**
+     An abstract base type for the value type. This is a hack to store multiple templated objects in a non
```
ocrickard: Meh, I wouldn't call it a hack.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -74,52 +114,205 @@ struct CKComponentViewAttribute {
...
+      {
+        return CK::hash(_value);
+      }
+
+      BOOL isEqualTo(const ValueBase &other) const
+      {
+        return this->typeName() == other.typeName() && CK::is_equal<id>()(_value, static_cast<const Value<id> &>(other)._value);
+      }
+
+      void performSetter(id object, SEL setter) const;
```
ocrickard: Out-of-line definition in a template? How does this work?

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
```diff
@@ -30,16 +30,17 @@ CKComponentViewAttributeValue CKComponentDelegateAttribute(SEL selector,
...
-      @YES  // Bogus value, we don't use it.
+      nil  // Bogus value, we don't use it.
     };
   }
 
   return {
     {
       std::string(sel_getName(selector)) + CKIdentifierFromDelegateForwarderSelectors(selectors),
-      ^(UIView *view, id value){
+      ^(UIView *view, CK::ViewAttribute::BoxedValue){
```
ocrickard: Can we at least typedef the non-namespaced versions of this? CKViewAttributeBoxedValue. We're trying to avoid the CK:: usage since it looks weird in Obj-C

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -74,52 +113,216 @@ struct CKComponentViewAttribute {
...
+        return !(*this == other);
+      }
+
+      void performSetter(id object, SEL setter) const
+      {
+        _value->performSetter(object, setter);
+      }
+
+    private:
+      std::shared_ptr<ValueBase> _value;
```
ocrickard: Hmm, not really. I don't think we've made the switch more broadly yet right?

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -194,11 +187,8 @@ namespace CK {
 
       void performSetter(id object, SEL setter) const
       {
-        const CachedSetter &set = CachedSetterInvocation(object, setter);
-        T value = _value;
-        [set.invocation setArgument:&value atIndex:2];
-        [set.invocation setSelector:setter];
-        [set.invocation invokeWithTarget:object];
+        const auto setterIMP = (void (*)(id, SEL, T))[object methodForSelector:setter];
```
ocrickard: Curious what happens here if I have a method signature that expects a const T &, and then you pass me T by value?

ComponentKit/Core/CKComponentViewAttribute.h
ocrickard: Ahh, my bad, this is in a template specialization. Missed that.

ComponentKit/Core/CKComponentViewAttribute.h
```diff
@@ -74,52 +114,205 @@ struct CKComponentViewAttribute {
...
+
+      BOOL isEqualTo(const ValueBase &other) const
+      {
+        return this->typeName() == other.typeName() && CK::is_equal<T>()(_value, static_cast<const Value<T> &>(other)._value);
+      }
+
+      void performSetter(id object, SEL setter) const
+      {
+        const auto setterIMP = (void (*)(id, SEL, T))[object methodForSelector:setter];
+        setterIMP(object, setter, _value);
```
ocrickard: What are the implications more broadly of moving us to use IMPs instead of invocations?

ocrickard:
Let's see if internal tests pass with this.

@facebook-github-bot import

ocrickard:
Few questions I'd like to talk about before pushing this.

ComponentKit/Core/CKComponentViewAttribute.h
ocrickard: Thinking about this more, by using IMPs, don't we disallow message forwarding entirely? That's probably OK, since if you need forwarding you can use the block syntax. This should be faster than invocations.

I'd like to add some type-checking here for DEBUG mode though. I'd grab the type signature from the object and verify that the type encoding for _value matches `@encode(T)`? Also add an assertion if setterIMP is NULL and no-op so we don't crash without a good debug message?

ComponentKit/Core/CKComponentViewAttribute.h
GotEmB: We currently compile using C++14 in our FB codebase.

ComponentKit/Core/CKComponentViewAttribute.h
GotEmB: I'm modifying the method signature to accept any callable type (function pointers, blocks, lambdas, functors).

Accepting an `std::function` directly here breaks brace initialization.

In addition, I still need to wrap this with a lambda as storing a block with signature `void (^)(UIView *, ...)` in an `std::function< void(id, ...) >` results in an `EXC_BAD` crash on the `std::function` destructor.

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
GotEmB: We definitely can, but this shouldn't be used in any user-code anyways. If a user is defining their own property, it should look something like:

```
const CKComponentViewAttributeValue customViewAttribute = {
  "custom-view-attribute",
  ^(CustomView *view, CustomType value) {
    view.customValue = value;
  }
};
```

Do you still think it's worth exposing this non-namespaced?

ComponentKit/Core/CKComponentViewAttribute.h
GotEmB: Added a test doing so and it works!

Is the `IMP` ignoring the typecast type (some kind of type erasure) and still taking in `_value` by reference?

ComponentKitTests/CKComponentViewAttributeTests.mm
```diff
@@ -81,15 +129,15 @@ - (void)testThatMountingViewWithPrimitiveAttributeActuallyAppliesAttributeToView
   CKComponent *testComponent = [CKComponent newWithView:{[CKNSNumberView class], {
     {@selector(setSelected:), YES},
     {CKComponentViewAttribute::LayerAttribute(@selector(setOpacity:)), @0.5},
-    {@selector(setTag:), 2},
+    {@selector(setTag:), (NSInteger)2},
```
GotEmB: Not really happy about me having to do this. Sending an `int` to a `long` setter does work as intended (due to the typecast we do to the `IMP`). But checking type encodings in debug mode fails the assert here.

GotEmB:
@facebook-github-bot import

GotEmB:
@facebook-github-bot import

ComponentKit/Utilities/CKComponentDelegateAttribute.mm
ocrickard: Ahh, that makes more sense. No need then.

ComponentKit/Core/CKComponentViewAttribute.h
ocrickard: lol, then go ahead and switch us over.

ocrickard:
@GotEmB I'll review the updates in the next day or so. In the meantime can you check the binary size impact of this change?

GotEmB:
@ocrickard, yep will do. I'm blocked by some build errors within fbobjc though (see imported phabricator diff). I should be able to fix them separately and rebase this import over it.

GotEmB:
@facebook-github-bot import

GotEmB:
Dropping this PR.

PR #749, opened 2017-02-13
[experimental] Introducing an enqueue state API

eczarny:
I wanted to explore an API that allows you to enqueue a state update that does not trigger a complete reflow of the component hierarchy. The idea here is to permit "lazy" state updates that aren't applied until a proper state update is scheduled.

There isn't anything here that can't be done without component controllers. Given that, and the somewhat confusing behavior of enqueued state updates, I don't recommend merging these changes. This pull request is completely experimental and for exploratory purposes only.

### DIFF
--- a/ComponentKit/Core/CKUpdateMode.h
+++ b/ComponentKit/Core/CKUpdateMode.h
@@ -10,7 +10,14 @@
 
 #import <Foundation/Foundation.h>
 
+/** The update mode is used to inform ComponentKit how to apply changes. */
 typedef NS_ENUM(NSUInteger, CKUpdateMode) {
+  /** Apply the update off the main thread. */
   CKUpdateModeAsynchronous,
+  /**
+   Apply the update on the main thread.
+   The synchronous update mode will not immediately block the executing thread when apply updates. Instead ComponentKit
+   will schedule the update to be performed on the next tick of the main thread's run loop.
+   */
   CKUpdateModeSynchronous,
 };
--- a/ComponentKit/HostingView/CKComponentHostingView.mm
+++ b/ComponentKit/HostingView/CKComponentHostingView.mm
@@ -148,14 +148,23 @@ struct CKComponentHostingViewInputs {
 
 #pragma mark - CKComponentStateListener
 
-- (void)componentScopeHandleWithIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier
-                            rootIdentifier:(CKComponentScopeRootIdentifier)rootIdentifier
-                     didReceiveStateUpdate:(id (^)(id))stateUpdate
+- (void)componentScopeHandleWithIdentifier:(int32_t)globalIdentifier
+                            rootIdentifier:(int32_t)rootIdentifier
+        didReceiveStateUpdateToBeScheduled:(id (^)(id))updateBlock
                                       mode:(CKUpdateMode)mode
 {
   CKAssertMainThread();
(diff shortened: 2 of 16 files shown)


ComponentKit/HostingView/CKComponentHostingView.mm
```diff
@@ -148,14 +148,23 @@ - (void)updateContext:(id<NSObject>)context mode:(CKUpdateMode)mode
...
 - (void)componentScopeHandleWithIdentifier:(CKComponentScopeHandleIdentifier)globalIdentifier
                             rootIdentifier:(CKComponentScopeRootIdentifier)rootIdentifier
-                     didReceiveStateUpdate:(id (^)(id))stateUpdate
+         didReceiveStateUpdateToBeEnqueued:(id (^)(id))updateBlock
                                       mode:(CKUpdateMode)mode
 {
   CKAssertMainThread();
-  _pendingInputs.stateUpdates.insert({globalIdentifier, stateUpdate});
-  [self _setNeedsUpdateWithMode:mode];
+  _pendingInputs.stateUpdates.insert({globalIdentifier, updateBlock});
```
eczarny: This is naive and likely doesn't work, but it gets the idea across.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSource.mm
```diff
@@ -157,11 +157,23 @@ - (void)componentScopeHandleWithIdentifier:(CKComponentScopeHandleIdentifier)glo
...
+{
+  CKAssertMainThread();
   if (mode == CKUpdateModeAsynchronous) {
-    _pendingAsynchronousStateUpdates[rootIdentifier].insert({globalIdentifier, stateUpdate});
+    _pendingAsynchronousStateUpdates[rootIdentifier].insert({globalIdentifier, updateBlock});
   } else {
-    _pendingSynchronousStateUpdates[rootIdentifier].insert({globalIdentifier, stateUpdate});
+    _pendingSynchronousStateUpdates[rootIdentifier].insert({globalIdentifier, updateBlock});
   }
 }
```
eczarny: This definitely doesn't work, you can verify that with my test in the sample app. The problem is this breaks the logic on line 155. Once you enqueue a state update the entire mechanism breaks. To make this work any enqueued updates would need to be tracked separately.

ComponentKit/Core/CKUpdateMode.h
```diff
@@ -10,7 +10,14 @@
...
 typedef NS_ENUM(NSUInteger, CKUpdateMode) {
+  /** Apply the update off the main thread. */
   CKUpdateModeAsynchronous,
+  /**
+   Apply the update on the main thread.
+   The synchronous update mode will not immediately block the executing thread when apply updates. Instead ComponentKit
+   will schedule the update to be performed on the next tick of the main thread's run loop.
+   */
   CKUpdateModeSynchronous,
 };
```
ocrickard: We should land these documentation changes.

eczarny: That's why I kept the changes around. :) I'll get these changes into a proper pull request.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSource.mm
ocrickard: lol yup, so simple.

eczarny:
Closing since this should not be merged.

PR #753, opened 2017-02-14
CKOverlayLayoutComponent renders to nil if component is nil

adamjernst:
Match `CKBackgroundLayoutComponent`, which was already doing the same. Also seize the opportunity for code style cleanup.


eczarny:
Nice! Seems reasonable to me.

eczarny:
@facebook-github-bot shipit

PR #754, opened 2017-02-15
Include the memory address of the lowest common ancestor in component scope collisions

eczarny:
The additional information will help make it clear which instance of a particular component in the component backtrace is the lowest common ancestor.


eczarny:
@facebook-github-bot import

PR #755, opened 2017-02-15
Crash when the transactional data source encounters a bad changeset

eczarny:
In most cases a bad changeset will result in a crash well after it has been applied to the transactional data source. Typically once the dependent collection view (or table view) processes that invalid changes. Crashing when encountering a bad changeset allows the transactional data source to fail faster with a bit more context at the ready.


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSource.mm
```diff
@@ -260,17 +260,14 @@ static void verifyChangeset(CKTransactionalComponentDataSourceChangeset *changes
                             CKTransactionalComponentDataSourceState *state,
                             NSArray<id<CKTransactionalComponentDataSourceStateModifying>> *pendingAsynchronousModifications)
 {
-#if CK_ASSERTIONS_ENABLED
```
eczarny: I'd appreciate thoughts on this. Determining if the changeset is valid involves some work, though I don't believe the performance characteristics are poor enough to warrant this check.

ocrickard: I'm pretty OK with doing this all the time. Let's measure the impact on performance and see if it's a problem.

ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSource.mm
```diff
@@ -260,17 +260,14 @@ static void verifyChangeset(CKTransactionalComponentDataSourceChangeset *changes
...
-            CKHumanReadableBadChangesetOperationType(badChangesetOperationType),
-            changeset,
-            state,
-            pendingAsynchronousModifications);
-#endif
+  if (badChangesetOperationType != CKBadChangesetOperationTypeNone) {
+    NSString *const humanReadableBadChangesetOperationType = CKHumanReadableBadChangesetOperationType(badChangesetOperationType);
+    [NSException raise:NSInternalInconsistencyException
+                format:@"Bad operation: %@\n*** Changeset:\n%@\n*** Data source state:\n%@\n*** Pending data source modifications:\n%@", humanReadableBadChangesetOperationType, changeset, state, pendingAsynchronousModifications];
+  }
```
ocrickard: Nice message.

ocrickard:
I'm on board with this change. I'd really like us to perf test it though.

adamjernst:
Perf test and let's go!

eczarny:
Hoping to perf test this next week.

eczarny:
Finally getting around to benchmarking `CKIsValidChangesetForState`, I don't have access to a device to run the benchmarks on so results are from my 15-inch 2016 MacBook Pro (on Monday I will run these benchmarks on my iPhone 5 test device):

https://gist.github.com/eczarny/d3b9f980ac4c6e09e7961ce464b85a2b
https://gist.github.com/eczarny/4a16f25f2a04c58bc0f277375fb56440

I only really focused on insertions, both with and without pending asynchronous modifications to fold into an initial state. Based on these results I wouldn't be surprised if verification of basic insertions would be somewhere between 6-12us, lower if there are fewer pending asynchronous modifications to apply. We'll see what the results look like on a lower end device.

eczarny:
Updated both gists with data from an actual device. The same general trends in performance apply, with results on device being noticeably worse (as expected).

For basic insertions we're still looking at microseconds, somewhere around 100us for insertions of 10 items. Performance certainly suffers as more and more pending asynchronous modifications need to be folded into the current state.

Results here can be on the oder of hundreds of microseconds, though that can creep up into milliseconds when you're talking about 15-20 pending asynchronous modifications to apply.

eczarny:
@facebook-github-bot import

eczarny:
@facebook-github-bot import

eczarny:
@facebook-github-bot import

eczarny:
Merged, though this behavior is disabled for now until we can further understand the potential performance impact.

PR #757, opened 2017-02-15
Pass accessibility configuration by reference

iKenWoo:
Fixes a small bug where accessibility labels aren't being passed through correctly. Also keeps it consistent with some of the recent changes to pass structs by reference.

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.h
+++ b/ComponentKit/Components/CKButtonComponent.h
@@ -36,6 +36,6 @@ struct CKButtonComponentAccessibilityConfiguration {
                        action:(const CKTypedComponentAction<UIEvent *> &)action
                          size:(const CKComponentSize &)size
                    attributes:(const CKViewComponentAttributeValueMap &)attributes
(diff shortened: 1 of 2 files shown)


ocrickard:
> Fixes a small bug where accessibility labels aren't being passed through correctly

Could you elaborate? And if this caused a bug, is it possible to write a failing unit test to include here?

iKenWoo:
To be honest, I haven't narrowed down the exact cause of the issue. 

When several button components were on screen, some of them would correctly have accessibility labels, and some of them wouldn't even though they were created the exact same way. When I set breakpoints at both the call-site and the constructor, I would see a valid value in the call-site, but inside the constructor the accessibilityLabel of the struct was nil. 

Marking this param as pass by reference fixed this. I'll try to come up with a test case to repro this consistently.

ocrickard:
Alright, I looked at the code a bit more and there should be no reason for this to fix a bug. I'm not opposed to the change, but if this is intended as a bug fix I think we should get more information, since I don't think this will fix it.

iKenWoo:
I'm going to close this for now.

PR #759, opened 2017-02-17
removed unused scenePath object from CK signatures

robspychala:
codemod -m --extensions h,m,mm 'scenePath'

and one by one removed all the instances of this signature

This is an API breaking change.

Tested in Xcode CMD-U.


eczarny:
Seems reasonable to me. I'll accept, and begin importing the change, but it will likely break internal builds until we remove any internal references to these parameters.

eczarny:
@facebook-github-bot import

PR #762, opened 2017-02-21
[ComponentKit] Add __kindof for CKComponentViewContext

supercfanx:
One often has to perform a cast when reading back the view from view context like this:

`
FooView *fooView = (FooView *)aComponent.viewContext.view;
`

Adding `__kindof` spares this cast and reduces noise in the code.


eczarny:
Neat! I'll check this out today.

eczarny:
@facebook-github-bot import

PR #764, opened 2017-02-26
Exposing component ID that caused reflow

ksenks:


ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceUpdateStateModification.mm
```diff
@@ -42,13 +42,15 @@ - (CKTransactionalComponentDataSourceChange *)changeFromState:(CKTransactionalCo
 
   NSMutableArray *newSections = [NSMutableArray array];
   NSMutableSet *updatedIndexPaths = [NSMutableSet set];
+  __block NSInteger globalIdentifier = 0;
```
eczarny: Nit: I'd use `CKComponentScopeHandleIdentifier` instead of `NSInteger`.

ComponentKit/TransactionalDataSources/Common/Internal/CKTransactionalComponentDataSourceUpdateStateModification.mm
```diff
@@ -42,13 +42,15 @@ - (CKTransactionalComponentDataSourceChange *)changeFromState:(CKTransactionalCo
...
   [[oldState sections] enumerateObjectsUsingBlock:^(NSArray *items, NSUInteger sectionIdx, BOOL *sectionStop) {
     NSMutableArray *newItems = [NSMutableArray array];
     [items enumerateObjectsUsingBlock:^(CKTransactionalComponentDataSourceItem *item, NSUInteger itemIdx, BOOL *itemStop) {
       const auto stateUpdatesForItem = _stateUpdates.find([[item scopeRoot] globalIdentifier]);
       if (stateUpdatesForItem == _stateUpdates.end()) {
         [newItems addObject:item];
       } else {
+        globalIdentifier = (stateUpdatesForItem->second).begin()->first;
```
eczarny: Any concern here that `(stateUpdatesForItem->second).begin()` may be empty? I think the behavior here would be undefined in that case, which would be less than desirable.

eczarny:
@facebook-github-bot import

eczarny:
@facebook-github-bot import

PR #766, opened 2017-03-01
Update 18-datasource-changeset-api.md

avnerbarr:
The bottom line of the code isn't visible in chrome

componentkit___changeset_api


avnerbarr:
I added an extra "enter" to the code block. I think it will add enough margin to the bottom portion so that the code will appear clearly.

PR #767, opened 2017-03-01
Fix changeset verification with a move whose original index path is non-existent after update

supercfanx:
Here is an example of what this diff is fixing. Consider a list being updated from `[A, B, C, D]` to `[D, A]`.

The changeset would be `[delete(from:[1,2]), move(from:3, to:0)]`.

Because this algorithm apply these changes in place with move handled in the last, at the time when moves are being verified, index 3 no longer exists.

The fix is to keep a copy of the original state and always verify fromIndexPath or fromSection with that.


eczarny:
@facebook-github-bot shipit

PR #768, opened 2017-03-03
Moving CKStackLayout to Flexbox

gkassabli:
This diff replaces CKStackLayoutComponent internal layout engine with Yoga.
Yoga is added as a submodule and included in build file


Carthage/Checkouts/yoga/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "lib/gtest/googletest"]
+	path = lib/gtest/googletest
+	url = https://github.com/google/googletest.git
```
ocrickard: I'd defer to @eczarny on this stuff.

Carthage/Checkouts/yoga/docs/_includes/nav/header_nav.html
```diff
@@ -0,0 +1,34 @@
...
+  document.addEventListener('slide', function (e) {
+    document.body.classList.toggle('sliderActive');
+  }, false);
+  var headerNav = document.getElementById('header_nav');
+  var headerNavExpander = document.getElementById('header_nav_expander');
+  headerNavExpander.addEventListener('click', function(e) {
+    headerNav.classList.toggle('navSlideoutActive');
+    document.dispatchEvent(event);
+  }, false);
+</script>
```
ocrickard: It's a bummer that our repo would contain all these things like the yoga website.

Carthage/Checkouts/yoga/lib/fb/include/fb/fbjni.h
```diff
@@ -0,0 +1,24 @@
...
+#include <fb/ALog.h>
+#include <fb/fbjni/Common.h>
+#include <fb/fbjni/Exceptions.h>
+#include <fb/fbjni/ReferenceAllocators.h>
+#include <fb/fbjni/References.h>
+#include <fb/fbjni/Meta.h>
+#include <fb/fbjni/CoreClasses.h>
+#include <fb/fbjni/Iterator.h>
+#include <fb/fbjni/Hybrid.h>
+#include <fb/fbjni/Registration.h>
```
ocrickard: And a whole jni library?

ComponentKit/LayoutComponents/CKStackLayoutComponent.mm
```diff
@@ -14,48 +14,328 @@
 
 #import <ComponentKit/CKMacros.h>
 
+#import "yoga/Yoga.h"
```
ocrickard: Curious why we decided to do the imports like this.

ComponentKit/LayoutComponents/CKStackLayoutComponent.mm
```diff
@@ -14,48 +14,328 @@
...
+  @property (nonatomic) CKComponentLayout componentLayout;
+  @property (nonatomic) float width;
+  @property (nonatomic) float height;
+  @property (nonatomic) YGMeasureMode widthMode;
+  @property (nonatomic) YGMeasureMode heightMode;
+  @property (nonatomic) CGSize parentSize;
+  @property (nonatomic) CKStackLayoutAlignSelf align;
+  @property (nonatomic) NSInteger zIndex;
+  
+  @end
```
ocrickard: these `@end` look indented.

ComponentKit/LayoutComponents/CKStackLayoutComponent.mm
```diff
@@ -14,48 +14,328 @@
...
+    YGNodeFreeRecursive(layoutNode);
+    return {self, {width, height}, childrenLayout};
+  }
+  
+  /*
+   layoutCache is passed by reference so that we are able to allocate it in one thread
+   and mutate it within that thread
+   Layout cache shouldn't be exposed publicly
+   */
+- (YGNodeRef)ygNode:(CKSizeRange)constrainedSize cache:(NSArray<CKFlexboxChildCachedLayout *> **)layoutCache
```
ocrickard: I think this would be better named `buildNode...`

gkassabli:
@ocrickard The idea was to add Yoga as a submodule, I might not executed it the way I wanted :) I've updated the diff to reflect only having Yoga as submodule

ocrickard:
Oh a submodule! Interesting! Honestly I'm out of my depth here. @eczarny what do you think the direction we should go with?

eczarny:
Adding Yoga as a Carthage dependency should include it as a submodule, which makes sense to me. We may also need to update the getting started documentation to run `carthage bootstrap` after cloning the repository in order to build.

Cartfile.private
```diff
@@ -1,3 +1,4 @@
 github "erikdoe/ocmock" ~> 3.3
 github "facebook/ios-snapshot-test-case" ~> 2.1
 github "jspahrsummers/xcconfigs" ~> 0.9
+github "facebook/yoga" ~> 1.2
```
eczarny: I'm not sure this dependency can go in this file. Technically Yoga is a public dependency. Anybody depending on ComponentKit will also need to get Yoga as well.

eczarny:
@gkassabli what are the benefits of having Carthage use a submodule? I only really thought it was useful if we were making changes within the submodule itself, which in this case I doubt we would be. If that that isn't correct (i.e. it would be useful when working on Yoga within the context of ComponentKit) then a submodule would be fine.

eczarny:
Now that ComponentKit depends on Yoga we may also be necessary to update our getting started documentation, particularly in the README.

https://github.com/Carthage/Carthage#running-a-project-that-uses-carthage

eczarny:
Hm. Looking at https://github.com/facebook/componentkit/tree/master/Carthage/Checkouts it appears that all the dependencies there are submodules. Might as well stick to that pattern here, so I'm on board @gkassabli.

Do you need to include the submodule in this pull request as well? I'm not seeing it be added under `Carthage/Checkouts`.

gkassabli:
@eczarny The idea is that Carthage would automatically update yoga as needed so that we don't have to manage that

conradev:
To follow up, the reason I [originally](https://github.com/facebook/componentkit/pull/534#issuecomment-211153130) made the dependencies submodules was so that you don't need Carthage to use the project at all 😮

gkassabli:
Right, I've added Yoga with the same logic - you can either use it as a submodule or part of Carthage

gkassabli:
@dshahidehpour just updated this to the latest version. btw, this needs review!

gkassabli:
Closing in favor of general sync PR

PR #769, opened 2017-03-03
Allow passing components into hierarchy logger

zats:
Sometimes when dumping components hierarchy all you have is a reference to a component and not a view. This allows to dump hierarchy starting with a component and not a view

### DIFF
--- a/ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
+++ b/ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
@@ -52,6 +52,9 @@ static NSString *const indentString = @"| ";
 
 + (NSString *)componentHierarchyDescriptionForView:(UIView *)view searchUpwards:(BOOL)upwards showViews:(BOOL)showViews
 {
+  if ([view isKindOfClass:[CKComponent class]]) {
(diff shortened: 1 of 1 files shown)


ComponentKit/Debug/CKComponentHierarchyDebugHelper.mm
```diff
@@ -52,6 +52,9 @@ + (NSString *)componentHierarchyDescriptionForView:(UIView *)view searchUpwards:
 
 + (NSString *)componentHierarchyDescriptionForView:(UIView *)view searchUpwards:(BOOL)upwards showViews:(BOOL)showViews
 {
+  if ([view isKindOfClass:[CKComponent class]]) {
+    view = ((CKComponent *)view).viewContext.view;
+  }
```
natansh: This is hacky. The functionality you care about can likely be implemented fully in chisel where you can do the same checks to get the view. It might be a bit less convenient to evaluate commands in python, but definitely doable.

zats: Cool, I'll close this one and open pull request to chisel then 👍

zats:
Moved it to https://github.com/facebook/chisel/pull/183

Issue #621, opened 2016-08-26
Unmount doesn't call attribute unapplicator


jaltreuter:
We're not calling the unapplicator in `[CKComponent unmount]`, despite the [comment in CKComponentInternal which states that as its contract](https://github.com/facebook/componentkit/blob/master/ComponentKit/Core/CKComponentInternal.h#L48). The only time I can find us calling unapplicator() is actually when applying a new set of attributes.

jaltreuter:
Here's a unit test that currently fails:

```
- (void)testThatUnmountingViewWithAttributeRequiringUnapplicationCallsUnapplicator
{
  NSMutableSet *appliedValues = [NSMutableSet set];
  NSMutableSet *unappliedValues = [NSMutableSet set];
  CKComponentViewAttribute attrWithUnapplicator("test", ^(id view, id val){
    [appliedValues addObject:val];
  }, ^(id view, id val){
    [unappliedValues addObject:val];
  });

  CKComponent *testComponent1 = [CKComponent newWithView:{[UIView class], {{attrWithUnapplicator, @1}}} size:{}];

  CKComponentLifecycleManager *m = [[CKComponentLifecycleManager alloc] init];
  [m updateWithState:{
    .layout = [testComponent1 layoutThatFits:{{0, 0}, {10, 10}} parentSize:kCKComponentParentSizeUndefined]
  }];

  UIView *container = [[UIView alloc] init];
  [m attachToView:container];

  XCTAssertEqualObjects(appliedValues, [NSSet setWithObject:@1], @"Expected @1 to be applied");
  XCTAssertEqualObjects(unappliedValues, [NSSet set], @"Expected nothing to be unapplied");

  [m detachFromView];

  XCTAssertEqualObjects(unappliedValues, [NSSet setWithObject:@1], @"Expected @1 to be unapplied");
}
```

ocrickard:
Yep, so sadly it's the docs on that internal header that are incorrect:

https://github.com/facebook/componentkit/blob/master/ComponentKit/Core/CKComponentInternal.h#L48

It took me awhile to remember, but I realized today that the unapplicator is actually called only when the view leaves the reuse pool:

https://github.com/facebook/componentkit/blob/master/ComponentKit/Core/CKComponentViewAttribute.h#L29

I'm looking for an alternate solution to the long press gesture recognizer problem. It's clear that the recognizer attribute is flawed.

jaltreuter:
[...]

PR #770, opened 2017-03-03
Remove incorrect comment for unmounting

ocrickard:
Fixes https://github.com/facebook/componentkit/issues/621. This comment is simply incorrect. Unapplication occurs on *mounting* not on *unmounting*. I'm open to moving this comment, but couldn't think of a good place to put it.

### DIFF
--- a/ComponentKit/Core/CKComponentInternal.h
+++ b/ComponentKit/Core/CKComponentInternal.h
@@ -45,7 +45,6 @@
  Unmounts the component:
  - Clears the references to supercomponent and superview.
  - If the component has a _mountedView:
-   - Calls the unapplicator for any attributes that have one.
    - Clears the view's reference back to this component in ck_component.
    - Clears _mountedView.
  */


eczarny:
I'm not sure how much value these comments provide?

The unapplicator itself should call out its contract, notably when it is expected to be called. Maybe that's a better place for this comment (e.g. explaining that attributes are unapplied on mount).

eczarny:
Back to you @ocrickard.

ocrickard:
> I'm not sure how much value these comments provide?

Yep, agreed.

I'll move the documentation for this specific one to the unapplicator callsite.

ocrickard:
@eczarny I looked to move it, but actually there's quite a bit of documentation in the unapplicator callsite:

https://github.com/facebook/componentkit/blob/82e3a0405cd4a9463cafc63b2e685722af20291b/ComponentKit/Core/CKComponentViewAttribute.h#L43-L57
https://github.com/facebook/componentkit/blob/82e3a0405cd4a9463cafc63b2e685722af20291b/ComponentKit/Core/CKComponentViewAttribute.h#L35-L37

I think we're good here. The documentation at the unmount callsite I believe is just bad.

eczarny:
@facebook-github-bot shipit

ocrickard:

PR #771, opened 2017-03-03
Fix blind call to targetForAction:withSender:

ocrickard:
So what happened here is that we added targetForAction:withSender: to CKComponents which is part of the UIResponder API, but when I made it possible to use non-CKComponents or responders as the target, I didn't catch that targetForAction:withSender: can't be called blindly.

This was working without crashing because I was testing it against views and view controllers, which subclass UIResponder and implement this method.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -88,7 +88,9 @@ std::string CKTypedComponentActionBase::identifier() const noexcept
 
 NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id target, CKComponent *sender) noexcept
 {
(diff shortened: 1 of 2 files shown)


ocrickard:
An alternative to this is that I could force the target/selector form of action to take a UIResponder.

eczarny:
> An alternative to this is that I could force the target/selector form of action to take a UIResponder.

Not a bad idea, but I'll leave that to you. This fix seems reasonable to me in the interim though.

eczarny:
I'll ship it once CI passes.

eczarny:
@facebook-github-bot import

eczarny:
No idea how long Travis will take. Building internally for now.

ocrickard:
Merged internally, CI is green, merging.

PR #773, opened 2017-03-05
Add more test related to move to CKTransactionalComponentDataSourceChangesetModificationTests

supercfanx:
Add another test case to cover a move related edge case, the same case that I fixed in https://github.com/facebook/componentkit/pull/767 for the verification algorithm.


eczarny:
@facebook-github-bot shipit

eczarny:
Thanks @supercfanx!

PR #774, opened 2017-03-09
Tweak object layout of CKTypedComponentActionBase to reduce code size

yiding:
The old layout involved two weak objc pointers, so the resulting destructor needs to call `objc_destroyWeak` twice (and associated offset calculations to find the fields). Since these objects are constructed (and destroyed) often, the inlined destruction cost is noticeable.

Since only one of the two weak fields are ever populated, I collapsed them into a single field, and use the enum to interpret it. I also moved the field to the start of the object so that no offset calculation is necessary. This makes the destructor effectively a single `objc_destroyWeak` call. I leave this inline-able since nothing would be gained by forcing another function call.

Here's an example of the destructor portion of the generated code of a function that creates 3 of these temporaries to pass to another function:

Before:
```
        add     r4, sp, #48
        add.w   r0, r4, #8
        bl      _objc_destroyWeak
        adds    r0, r4, #4
        bl      _objc_destroyWeak
        add.w   r0, r8, #8
        bl      _objc_destroyWeak
        add.w   r0, r8, #4
        bl      _objc_destroyWeak
        add     r4, sp, #80
        add.w   r0, r4, #8
        bl      _objc_destroyWeak
        adds    r0, r4, #4
        bl      _objc_destroyWeak
```

After:
```
        add     r0, sp, #48
        bl      _objc_destroyWeak
        add     r0, sp, #60
        bl      _objc_destroyWeak
        mov     r0, r8
        bl      _objc_destroyWeak
```

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -29,14 +29,13 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
 bool CKTypedComponentActionBase::operator==(const CKTypedComponentActionBase& rhs) const
 {
   return (_variant == rhs._variant
-          && CKObjectIsEqual(_target, rhs._target)
-          && CKObjectIsEqual(_scopeHandle, rhs._scopeHandle)
+          && CKObjectIsEqual(_targetOrScopeHandle, rhs._targetOrScopeHandle)
           && _selector == rhs._selector);
 }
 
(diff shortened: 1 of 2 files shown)


eczarny:
Build failures seem relevant to the changes here.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot shipit

ocrickard:
Landed internally, merging.

PR #777, opened 2017-03-09
Drop support for CocoaPods

eczarny:
Supporting CocoaPods has always been a pain, but we've managed. Unfortunately it looks like we may have hit a fairly big blocker: [CocoaPods 1.x is now needed to publish new ComponentKit release](http://blog.cocoapods.org/Sharding/) but only [CocoaPods 0.39.0 can correctly lint Objective-C++](https://github.com/CocoaPods/CocoaPods/issues/5152).

Moving forward ComponentKit will only support Carthage. Anybody looking to use ComponentKit in a project that doesn't use Carthage can simply add ComponentKit as a submodule and manually add it as a dependency in their Xcode project.


ocrickard:
Yeah, it's a bummer that obj-c++ support has been dropped. Oh well.

eczarny:
There are likely workarounds but, ugh, they aren't pretty and it just isn't sustainable for us to support it any longer.

adamjernst:
@facebook-github-bot shipit

dnkoutso:
FWIW in CP 1.2.1.beta.1 a flag was added (https://github.com/CocoaPods/CocoaPods/issues/5670) in the validator to skip linking, which matches the behavior of 0.39 that used to allow publishing without linking. I tried this with your podspec and it works:

```
pod spec lint ComponentKit.podspec --verbose --no-clean --skip-import-validation
```

Output:

```
** BUILD SUCCEEDED **

 -> ComponentKit (0.2)
    - NOTE  | xcodebuild:  warning: no rule to process file 'ComponentKit/ComponentKit/ComponentKit.xcconfig' of type text.xcconfig for architecture i386
    - NOTE  | xcodebuild:  warning: no rule to process file 'ComponentKit/ComponentKit/ComponentKit.xcconfig' of type text.xcconfig for architecture x86_64

Analyzed 1 podspec.

ComponentKit.podspec passed validation.
```

I understand there are plenty of other reasons to not support CocoaPods anymore but there are plenty of projects probably that use it and could consume ComponentKit as a pod.

Issue #622, opened 2016-08-26
Gesture recognizers can fire on unmounted views


jaltreuter:
When a UILongPressGestureRecognizer is registered on a component's view, since the gesture isn't instantaneous it's possible for the component to be unmounted mid-gesture. Then, when I lift my finger, [CKComponentGestureActionForwarder](https://github.com/facebook/componentkit/blob/master/ComponentKit/Utilities/CKComponentGestureActions.mm#L182) fires a component action. `recognizer.view` is a view that exists in the view hierarchy, hidden, but it's not linked to any component in the component hierarchy so `recognizer.view.ck_component` is nil and [CKComponentActionSend](https://github.com/facebook/componentkit/blob/master/ComponentKit/Utilities/CKComponentAction.mm#L37) throws an assertion.

Since we intentionally don't call the unapplicator block until the view leaves the reuse pool, two options come to mind:
1. Add a check in CKComponentGestureActionForwarder to not send the gesture action if `recognizer.view.ck_component` is nil
2. Add a new attribute lifecycle block which is called when a view enters the reuse pool (which may require a second block called when it exits the recycle pool)

PR #780, opened 2017-03-13
CKGestureAction will not fire if the view has a nil ck_component

aCorrado:
It solves #622.

I went with solution number 1, because in this case it's quite effective.

Adding 2 new lifecycle blocks for attributes (solution 2) is, for me, not the way to go because interactions between applicator, unapplicator, updater, and the new blocks will be quite complex. (glad to be proven wrong on this)

Another approach that I would follow is to create a subclass (substruct??) of CKComponentViewAttribute (maybe CKComponentViewAttributeUnmountRemovable) that can is removed at unmount not when the view is remounted.
I cannot see any other obvious attribute that would benefit of this procedure (that's why it's more a proposal and i have not implemented), but if we found any other we should think about what's the best approach.

Please let me know what do you think about it.

### DIFF
--- a/ComponentKit/Utilities/CKComponentGestureActions.mm
+++ b/ComponentKit/Utilities/CKComponentGestureActions.mm
@@ -179,8 +179,10 @@ CKComponentViewAttributeValue CKComponentGestureAttribute(Class gestureRecognize
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
-  // If the action can be handled by the sender itself, send it there instead of looking up the chain.
-  [recognizer ck_componentAction].send(recognizer.view.ck_component, CKComponentActionSendBehaviorStartAtSender, recognizer);
+  if (recognizer.view.ck_component) {
+    // If the action can be handled by the sender itself, send it there instead of looking up the chain.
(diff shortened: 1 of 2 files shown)


ComponentKit/Utilities/CKComponentGestureActions.mm
```diff
@@ -179,8 +179,10 @@ + (instancetype)sharedInstance
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
-  // If the action can be handled by the sender itself, send it there instead of looking up the chain.
-  [recognizer ck_componentAction].send(recognizer.view.ck_component, CKComponentActionSendBehaviorStartAtSender, recognizer);
+  if (recognizer.view.ck_component) {
+    // If the action can be handled by the sender itself, send it there instead of looking up the chain.
+    [recognizer ck_componentAction].send(recognizer.view.ck_component, CKComponentActionSendBehaviorStartAtSender, recognizer);
+  }
```
ocrickard: Looks great to me!

ocrickard:
@facebook-github-bot shipit

ocrickard:

PR #781, opened 2017-03-15
Remove -layoutThatFits:parentSize from Docs.

dshahidehpour:
We only expose `layoutThatFits:parentSize` in `CKComponentSubclass`, not on the base object. Removing it from the docs.


ryanrhee:
@adamjernst @ocrickard @eczarny ?

PR #785, opened 2017-03-23
Add appearance API on hosting views

supercfanx:
Currently appearance callbacks (`componentTreeWillAppear` and `componentTreeDidDisappear`) on the component controller are not called when a component is hosted in a hosting view. This change provides a way for users to explicitly trigger appearance events on the hosting view, which will then be funneled to the component controller.

This has to be done manually since there is no easy way to infer the visibility change of a view (there are so many factors like alpha value of ancestor, covered by other views etc)


ocrickard:
Looks pretty reasonable to me.

ocrickard:
@facebook-github-bot shipit

ocrickard:
Landed internally. Merging.

PR #786, opened 2017-03-28
Make Nil Actions no-op, remove assert.

dshahidehpour:
Currently, if you have a component that takes an action like so

```
[CKFooComponent newWithAction:nil];
```

this will trigger multiple asserts within `CKComponentAction.mm`. This diffs removes those asserts, and simply causes the action to no-op. I've updated the docs, and added a test to verify the new behavior.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -73,14 +73,14 @@ std::string CKTypedComponentActionBase::identifier() const noexcept
 
 NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id target, CKComponent *sender) noexcept
 {
+  // If we have a nil selector, we bail early.
+  if (selector == nil) {
+    return nil;
+  }
+
   id responder = ([target respondsToSelector:@selector(targetForAction:withSender:)]
                   ? [target targetForAction:selector withSender:target]
                   : target);
-
-  // If we have no responder to start, bail early.
-  if (responder == nil) {
-    return nil;
-  }
 
   // This is not performance-sensitive, so we can just use an invocation here.
   NSMethodSignature *signature = [responder methodSignatureForSelector:selector];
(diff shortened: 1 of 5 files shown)


ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -76,8 +76,12 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
   id responder = ([target respondsToSelector:@selector(targetForAction:withSender:)]
                   ? [target targetForAction:selector withSender:target]
                   : target);
-  CKCAssertNotNil(responder, @"Unhandled component action %@ following responder chain %@",
-                  NSStringFromSelector(selector), _CKComponentResponderChainDebugResponderChain(target));
+
+  // If we have no responder to start, bail early.
+  if (responder == nil) {
```
ocrickard: How are you differentiating between a nil responder, and a weak responder that has been deallocated? A straight-up nil target I think is fine to remove the assert. However, targets are held weakly, and a common error is them being deallocated before the action fires, which would now not be caught at all, right?

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -76,8 +76,12 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
   id responder = ([target respondsToSelector:@selector(targetForAction:withSender:)]
                   ? [target targetForAction:selector withSender:target]
                   : target);
-  CKCAssertNotNil(responder, @"Unhandled component action %@ following responder chain %@",
-                  NSStringFromSelector(selector), _CKComponentResponderChainDebugResponderChain(target));
+
```
ocrickard: Why don't we just early-return if there's a nil selector here? There are real cases that are bugs when the target/responder are nil that we still want to assert on, right?

ocrickard: The *selector* on the other hand should never be nil if this is a valid action.

dshahidehpour:
checking for a nil selector sounds a lot better! I'll update the diff.

Thanks @ocrickard!

dshahidehpour:
@facebook-github-bot shipit

dshahidehpour:
@facebook-github-bot import

PR #787, opened 2017-03-28
Make scopes generic to component type

ocrickard:
This still needs some cleanup, but the tests are passing, and this demonstrates the approach.

We have 2 base classes for components, the standard CKComponent type, then another type specifically for data fetching. We'd like to be able to take the implementation of scopes, and reuse it there so things like component actions and triggers work seamlessly.

I implemented this in 3 different ways, and this way was my favorite. For reference here's also what I tried:

1. Obj-C Generics + select templates. This was a real pain because these two generic systems don't work together.
2. After the above failed, I tried just straight using C++ templates, but the code for this was ugly in my opinion, and I was fairly certain would result in binary bloat.
3. This approach, which uses protocols for components, and controllers, for the methods that are needed.

In all of the above 3 implementations I've had to make the component appearance/bounds animation events generic. I've introduced the idea of a "predicate" that is a normal C function pointer that is passed to a scope root on its initialization. All components and controllers will register with the scope root on construction. After this process, we record every component and controller in the tree that match the predicate. Later, when we want to announce something, we give the scope root the predicate we're interested in, and it allows us to rapidly enumerate over all components or controllers which matched that predicate.

CKComponent and CKComponentController both declare conformance with the protocol.

TODO:

1. Needs tests for the generic predicate system. It's tested through the appearance events and bounds animations right now, but probably some dedicated tests here would help exercise the corner cases of the API.
2. Possibly extend the tests of the bounds animation system. There appear to be minimal tests there.
3. Clean up, naming, copyright notices, etc.


ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,63 @@
...
+     CKComponent *oldComponent = (CKComponent *)component;
+     id scopeFrameToken = [oldComponent scopeFrameToken];
+     if (scopeFrameToken) {
+       [scopeFrameTokenToOldComponent setObject:oldComponent forKey:scopeFrameToken];
+     }
+   }];
+  
+  __block CKComponentBoundsAnimation boundsAnimation {};
+  [newRoot
+   enumerateComponentsMatchingPredicate:&CKComponentBoundsAnimationPredicate
```
ocrickard: Here's an example of the enumeration system I mentioned.

ocrickard: Before, all of this was implemented inside the scope root itself. Now it's implemented in a few functions above the scope root using a generic API.

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
```diff
@@ -0,0 +1,17 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKComponentScopeRoot.h>
+
+BOOL CKComponentControllerAppearanceEventPredicate(id<CKScopedComponentController> controller);
+BOOL CKComponentControllerDisappearanceEventPredicate(id<CKScopedComponentController> controller);
+
+void CKComponentControllerAnnounceAppearance(CKComponentScopeRoot *scopeRoot);
+void CKComponentControllerAnnounceDisappearance(CKComponentScopeRoot *scopeRoot);
```
ocrickard: These should rarely, if ever be used directly. Generally they're configured for you when you're using a scope root.

ComponentKit/Core/Scope/CKComponentScopeRootFactory.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+@protocol CKComponentStateListener;
+
+@class CKComponentScopeRoot;
+
+CKComponentScopeRoot *CKComponentScopeRootWithListener(id<CKComponentStateListener> listener);
```
ocrickard: This is the normal factory that should be used with CKComponent implementations. For working ranges implementation we'd like to override the predicates for controllers to make the tree traversal more efficient, so may need a way to inject this somehow into the hosting view and datasource. Thoughts?

ComponentKit/Core/Scope/CKComponentScopeRootFactory.mm
```diff
@@ -0,0 +1,25 @@
...
+{
+  return [CKComponentScopeRoot
+   rootWithListener:listener
+   componentPredicates:{
+     &CKComponentBoundsAnimationPredicate
+   }
+   componentControllerPredicates:{
+     &CKComponentControllerAppearanceEventPredicate,
+     &CKComponentControllerDisappearanceEventPredicate
+   }];
```
ocrickard: The predicates are just C function pointers.

ComponentKit/Core/Scope/CKScopedComponent.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Oliver Rickard on 3/27/17.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol CKScopedComponent <NSObject>
+
+@end
```
ocrickard: I was surprised that the scope infra actually doesn't care about the component type at all, thus this protocol is empty. I still want to keep it so we get type annotations on the id types for which are the components, and which are the controllers in the interfaces of the scope infra. Straight id's are unreadable.

ComponentKit/Core/Scope/CKScopedComponentController.h
```diff
@@ -0,0 +1,17 @@
...
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol CKScopedComponent;
+
+@protocol CKScopedComponentController <NSObject>
+
+- (instancetype)initWithComponent:(id<CKScopedComponent>)component;
```
ocrickard: The scope infra actually constructs component controllers at the moment. I considered pulling this out and passing in a component controller factory, but decided against for now. May need that for section components though.

ocrickard:
@facebook-github-bot import

ComponentKit/Core/CKBuildComponent.h
```diff
@@ -0,0 +1,27 @@
...
+
+struct CKBuildComponentResult {
+  CKComponent *component;
+  CKComponentScopeRoot *scopeRoot;
+  CKComponentBoundsAnimation boundsAnimation;
+};
+
+CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
+                                        const CKComponentStateUpdateMap &stateUpdates,
+                                        CKComponent *(^function)(void));
```
eczarny: I'm a big fan of moving this function to its own header and source file.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -38,9 +38,7 @@ + (CKComponentScopeHandle *)handleForComponent:(CKComponent *)component
 
   CKComponentScopeHandle *handle = currentScope->stack.top().frame.handle;
   if ([handle acquireFromComponent:component]) {
-    if (CKSubclassOverridesSelector([CKComponent class], [component class], @selector(boundsAnimationFromPreviousComponent:))) {
-      [currentScope->newScopeRoot registerBoundsAnimationComponent:component];
-    }
+    [currentScope->newScopeRoot registerComponent:component];
```
eczarny: Cool, so register components with the scope root regardless of what it's actually being used for.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -26,7 +27,7 @@
  handle corresponding to the current scope. Otherwise it returns nil.
  This is only meant to be called when constructing a component and as part of the implementation itself.
  */
-+ (instancetype)handleForComponent:(CKComponent *)component;
++ (instancetype)handleForComponent:(id<CKScopedComponent>)component;
```
eczarny: Github won't let me add a comment in the implementation file but shouldn't we update the signature in the implementation file as well?

ComponentKit/Core/Scope/CKComponentScopeRoot.h
```diff
@@ -10,18 +10,23 @@
 
 #import <Foundation/Foundation.h>
 
+#import <vector>
+
 #import <ComponentKit/CKComponentBoundsAnimation.h>
```
eczarny: Is this still needed?

ComponentKit/Core/Scope/CKComponentScopeRoot.mm
```diff
@@ -9,126 +9,109 @@
...
 #import "CKComponentScopeRoot.h"
-#import "CKComponentScopeRootInternal.h"
 
 #import <libkern/OSAtomic.h>
 
-#import "CKComponentController.h"
-#import "CKComponentInternal.h"
+#import "CKScopedComponent.h"
+#import "CKScopedComponentController.h"
 #import "CKComponentSubclass.h"
```
eczarny: Another import that can be removed?

ComponentKit/Core/Scope/CKScopedComponentController.h
eczarny: I think you made the right call. That change alone is fairly substantial, it would make reviewing these changes a bit more difficult.

ComponentKit/Core/Scope/CKComponentScopeRootFactory.h
eczarny: Hm. I'm not sure we would necessarily want to override the predicates but instead rely on a separate set of predicates. Right? There shouldn't be a need to replace the appearance event predicates when incorporating additional working range predicates.

I'm having a bit of trouble seeing how we could leverage this new scope API outside where it's being used internally by ComponentKit. Let's sync up on this when you have some time.

ComponentKit/Core/Scope/CKComponentScopeRootFactory.mm
eczarny: Makes sense, though this does make it a bit harder to implement more efficient traversals. It may be necessary to also pass along some state that can be used for caching the predicate results. What do you think?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
ocrickard: yep

ComponentKit/Core/Scope/CKComponentScopeRootFactory.h
ocrickard: Yeah, the predicates we'd want to add would be used in addition to the current predicates for appearance and bounds animations.

ComponentKit/Core/Scope/CKComponentScopeRootFactory.mm
ocrickard: Oh to be clear, this is the efficient traversal. We build up the list of matching components on component generation, so by nature this is cached. It happens once for a scope root, and then the results are used over the lifetime.

ComponentKit/Core/CKBuildComponent.h
```diff
@@ -0,0 +1,27 @@
+//
+//  CKBuildComponent.h
+//  ComponentKit
+//
+//  Created by Oliver Rickard on 3/28/17.
```
adamjernst: fix header

ComponentKit/Core/CKBuildComponent.h
```diff
@@ -0,0 +1,27 @@
...
+//
+//
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKComponentScopeTypes.h>
+#import <ComponentKit/CKComponentBoundsAnimation.h>
+
+@protocol CKScopedComponent;
```
adamjernst: remove; not used in header

ComponentKit/Core/CKBuildComponent.mm
adamjernst: so so good.

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,63 @@
...
+  NSMapTable *scopeFrameTokenToOldComponent = [NSMapTable strongToStrongObjectsMapTable];
+  [previousRoot
+   enumerateComponentsMatchingPredicate:&CKComponentBoundsAnimationPredicate
+   block:^(id<CKScopedComponent> component) {
+     CKComponent *oldComponent = (CKComponent *)component;
+     id scopeFrameToken = [oldComponent scopeFrameToken];
+     if (scopeFrameToken) {
+       [scopeFrameTokenToOldComponent setObject:oldComponent forKey:scopeFrameToken];
+     }
+   }];
```
adamjernst: you might want to use a std::unordered_map. if you wanted you could even use __unsafe_unretained for the keys since they are guaranteed not to go out of scope during this function's execution.

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,63 @@
...
+  return boundsAnimation;
+}
+
+CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
+                                        const CKComponentStateUpdateMap &stateUpdates,
+                                        CKComponent *(^function)(void))
+{
+  CKThreadLocalComponentScope threadScope(previousRoot, stateUpdates);
+  // Order of operations matters, so first store into locals and then return a struct.
+  CKComponent *component = function();
```
adamjernst: const

ComponentKit/Core/CKComponentBoundsAnimationPredicates.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Oliver Rickard on 3/28/17.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKComponentScopeRoot.h>
+
+BOOL CKComponentBoundsAnimationPredicate(id<CKScopedComponent> controller);
```
adamjernst: docs?

ComponentKit/Core/CKBuildComponent.mm
adamjernst: eh actually don't do this

ComponentKit/Core/CKComponentControllerInternal.h
```diff
@@ -11,8 +11,9 @@
 #import <UIKit/UIKit.h>
 
 #import <ComponentKit/CKComponentController.h>
+#import <ComponentKit/CKScopedComponentController.h>
 
-@interface CKComponentController ()
+@interface CKComponentController () <CKScopedComponentController>
```
adamjernst: seems kinda mysterious to me, not knowing what it is

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
adamjernst: can we just hide these inside the file that uses them instead of making them public?

ComponentKit/Core/Scope/CKComponentScopeHandle.h
adamjernst: ahh so this is why the new protocol exists. we can reuse it for… other similar systems.

should we call this `CKScopedObject` instead? `CKScopedElement`? Some other generic name?

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
ocrickard: Sadly we have several places that use these.

ComponentKit/Core/CKComponentBoundsAnimationPredicates.h
dshahidehpour: Not to woodshed too much, but, can we reword this function's name?

`CKComponentHasCustomImplementationForBoundsAnimation()`

or something?

small nit: component, not controller.

ComponentKit/Core/Scope/CKComponentScopeRoot.mm
```diff
@@ -9,126 +9,109 @@
...
-  for (const auto &announceableEvent : announceableEvents()) {
-    if (CKSubclassOverridesSelector([CKComponentController class], [controller class], announceableEvent.second)) {
-      _eventRegistration.insert({{announceableEvent.first, controller}});
+  if (!component) {
+    // Handle this gracefully so we don't have a bunch of nils being passed to predicates.
+    return;
+  }
+  for (const auto &predicate : _componentPredicates) {
+    if (predicate(component)) {
+      _registeredComponents.insert({predicate, component});
```
dshahidehpour: How does this work? Since the same predicate serves as the same key for multiple components, won't the insertion fail?

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
dshahidehpour: I find this method a little confusing because you are providing a scope root. Could we name it something like:

`CKScopeRootAnnounceApperanceToControllers(CKComponentScopeRoot *scopeRoot)`

ComponentKit/Core/CKBuildComponent.mm
ocrickard: Yeah, this is just the old code.

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,63 @@
...
+       CKComponent *oldComponent = [scopeFrameTokenToOldComponent objectForKey:scopeFrameToken];
+       if (oldComponent) {
+         const CKComponentBoundsAnimation ba = [newComponent boundsAnimationFromPreviousComponent:oldComponent];
+         if (ba.duration != 0) {
+           boundsAnimation = ba;
+         }
+       }
+     }
+   }];
+  return boundsAnimation;
```
ocrickard: lol yep! That's the contract with bounds animations. Only a single one gets executed.

ComponentKit/Core/Scope/CKComponentScopeRoot.h
ocrickard: Nope, good eye.

ComponentKit/Core/Scope/CKComponentScopeRoot.mm
ocrickard: Probably

ComponentKit/Core/Scope/CKComponentScopeRoot.mm
ocrickard: lol yeah, this was supposed to be a multimap, not a straight unordered_map. Good find.

ComponentKit/Core/CKBuildComponent.mm
dshahidehpour: wowow

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
ocrickard: I think you're right.

ComponentKit/Core/CKBuildComponent.h
dshahidehpour: I think we can still remove this

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,65 @@
...
+#import "CKComponentSubclass.h"
+#import "CKThreadLocalComponentScope.h"
+
+static CKComponentBoundsAnimation boundsAnimationFromPreviousScopeRoot(CKComponentScopeRoot *newRoot, CKComponentScopeRoot *previousRoot)
+{
+  NSMapTable *const scopeFrameTokenToOldComponent = [NSMapTable strongToStrongObjectsMapTable];
+  [previousRoot
+   enumerateComponentsMatchingPredicate:&CKComponentBoundsAnimationPredicate
+   block:^(id<CKScopedComponent> component) {
+     CKComponent *oldComponent = (CKComponent *)component;
```
dshahidehpour: The `CKScopedComponent` protocol is currently empty, maybe we should consider adding `scopeFrameToken` as a required property, then we can avoid the cleanup of this cast in the future?

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -0,0 +1,65 @@
...
+  __block CKComponentBoundsAnimation boundsAnimation {};
+  [newRoot
+   enumerateComponentsMatchingPredicate:&CKComponentBoundsAnimationPredicate
+   block:^(id<CKScopedComponent> component) {
+     CKComponent *newComponent = (CKComponent *)component;
+     id scopeFrameToken = [newComponent scopeFrameToken];
+     if (scopeFrameToken) {
+       CKComponent *oldComponent = [scopeFrameTokenToOldComponent objectForKey:scopeFrameToken];
+       if (oldComponent) {
+         const CKComponentBoundsAnimation ba = [newComponent boundsAnimationFromPreviousComponent:oldComponent];
```
dshahidehpour: maybe it would be worth adding to the `CKScopedComponent`, future implementations can just return a empty implementation.

Do you think that's too much stuff to add to the protocol?

ComponentKit/Core/CKBuildComponent.h
```diff
@@ -0,0 +1,29 @@
...
+
+struct CKBuildComponentResult {
+  CKComponent *component;
+  CKComponentScopeRoot *scopeRoot;
+  CKComponentBoundsAnimation boundsAnimation;
+};
+
+CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
+                                        const CKComponentStateUpdateMap &stateUpdates,
+                                        CKComponent *(^function)(void));
```
dshahidehpour: I know you are just moving code around, but can we give this a better name? `creationBlock`, maybe?

ComponentKit/Core/CKComponentBoundsAnimationPredicates.h
dshahidehpour: +1

ComponentKit/Core/Scope/CKComponentControllerAppearanceEvents.h
dshahidehpour: +1 to this still

ComponentKit/Core/Scope/CKComponentScopeRootFactory.h
dshahidehpour: Just playing devil's advocate, how does someone know to use this factory function, or the initializer for `CKComponentScopeRoot`? Can we make it more clear that this isn't a "plain ol' CKComponentScopeRoot?"

`CKComponentScopeRootWithDefaultPredicates(id<CKComponentStateListner> listener)`?

or, since these predicates are baked into the infra

```
CKComponentScopeRoot::rootWithDefaultPredicates(id<CKComponentStateListener> listener)
```

something like that?

ComponentKit/Core/CKBuildComponent.mm
ocrickard: Ahh, interesting! That's a good observation. Let's see if our other implementation has the same problem.

ocrickard:
I smoke-tested this internally, should be ready to merge now. Accept anyone?

ocrickard:
@facebook-github-bot import

ocrickard:
Alright, this landed internally. Merging.

PR #789, opened 2017-04-04
Use firmer language when describing scope collisions

eczarny:
It isn't obvious what the impact of a scope collision is based on the language used in the error. I've updated the error's description to be a bit firmer in language, describing the possibility of strange and unexpected behavior (along with some simple, yet actionable, next steps).


dshahidehpour:
@facebook-github-bot shipit

PR #791, opened 2017-04-05
[componentkit] make descriptions more readable for debugging

hutley:
^^ so that when they print in logs its easier to decipher


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceChangeset.mm
```diff
@@ -7,11 +7,14 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  *
  */
-
+#import "CKTransactionalComponentDataSourceChangeset.h"
 #import "CKTransactionalComponentDataSourceChangesetInternal.h"
 
+#import <UIKit/UITableView.h>
```
adamjernst: would prefer we don't depend on UITableView, but just do it manually, but oh well

adamjernst:
@facebook-github-bot shipit

PR #794, opened 2017-04-17
Check viewClass being a subclass of a UIView

zats:
This should help avoiding silly bugs like the one I had when I passed a `XComponent` instead instead of `XComponentView`.

I couldn't figure out if there's an earlier point where we can to assert on the `viewClass`?

### DIFF
--- a/ComponentKit/Core/CKComponentViewConfiguration.mm
+++ b/ComponentKit/Core/CKComponentViewConfiguration.mm
@@ -20,7 +20,9 @@ CKComponentViewClass::CKComponentViewClass() noexcept : factory(nil) {}
 
 CKComponentViewClass::CKComponentViewClass(Class viewClass) noexcept :
 identifier(class_getName(viewClass)),
-factory(^{return [[viewClass alloc] init];}) {}
+factory(^{ return [[viewClass alloc] init]; }) {
+  CKCAssert([viewClass isSubclassOfClass:[UIView class]], @"%@ is not a subclass of UIView", viewClass);
+}
 
 static CKComponentViewReuseBlock blockFromSEL(SEL sel) noexcept
 {


ComponentKit/Core/CKComponentViewConfiguration.mm
```diff
@@ -20,7 +20,12 @@
 
 CKComponentViewClass::CKComponentViewClass(Class viewClass) noexcept :
 identifier(class_getName(viewClass)),
-factory(^{return [[viewClass alloc] init];}) {}
+factory(^{
+  #if DEBUG
+    CKAssertTrue([viewClass isSubclassOfClass:[UIView class]], @"%@ is not a subclass of UIView", viewClass);
+  #endif
+  return [viewClass isSubclassOfClass:[UIView class]] ? [[viewClass alloc] init] : nil;
+}) {}
```
adamjernst: put the assert in the `{}` block (which is the constructor body) instead of in the factory. that'll run sooner — at the point this `CKComponentViewClass` object is instantiated instead of the point where the factory is invoked.

zats: Is this enough though? When we are using component kit as a precompiled dependency won't it be compiled without `DEBUG`?

adamjernst: no, it is never used as a "precompiled dependency." However, don't bother with #DEBUG, CKAssert is compiled out in non-debug builds for you (in the configuration we use internally)

zats:
I'm not that good with C++, please check if it's the right place to assert

ComponentKit/Core/CKComponentViewConfiguration.mm
```diff
@@ -20,7 +20,11 @@
 
 CKComponentViewClass::CKComponentViewClass(Class viewClass) noexcept :
 identifier(class_getName(viewClass)),
-factory(^{return [[viewClass alloc] init];}) {}
+factory(^{ return [[viewClass alloc] init]; }) {
+  #if DEBUG
```
ocrickard: Remove the #if debugs here

adamjernst: ++

zats: Should we remove it from https://github.com/facebook/componentkit/blob/ed357f67df3d398cccc536a0c7b280e3e0b62ea8/ComponentKit/Core/CKComponentViewConfiguration.mm#L59, too?

ComponentKit/Core/CKComponentViewConfiguration.mm
```diff
@@ -20,7 +20,9 @@
 
 CKComponentViewClass::CKComponentViewClass(Class viewClass) noexcept :
 identifier(class_getName(viewClass)),
-factory(^{return [[viewClass alloc] init];}) {}
+factory(^{ return [[viewClass alloc] init]; }) {
+  CKAssertTrue([viewClass isSubclassOfClass:[UIView class]], @"%@ is not a subclass of UIView", viewClass);
```
gkassabli: You probably wanted to use CKAssert here - CKAssertTrue doesn't take description

ComponentKit/Core/CKComponentViewConfiguration.mm
eczarny: I don't think it would hurt. Looks like the `DEBUG` check @zats mentions is from the initial commit. @adamjernst any idea if we need to keep that `DEBUG` check or if we can remove it too?

eczarny:
Build is failing.

zats:
Ugh, sorry, such a tiny change caused so many troubles 😳

zats:
Passing now 👍

PR #795, opened 2017-04-18
Add a block interface for actions

ocrickard:
We originally had an interface like this for JS interop, but we removed it due to binary size. Let's try adding this back and see what it does since @ashwinb needs this for JS interop in CK.

This is a test diff to see binary size impact. I'm mostly concerned about additional objc_storeStrong in every receiving callsite of a component action.

BTW, using templates we can actually handle the block promotion/demotion relatively nicely, since each time the block is converted, we know the number of arguments before/after so we can wrap the blocks with a new template pack.... however at the cost of additional code bloat at the conversion callsites, and unlike selectors when promoting/demoting we actually slice off the values as you convert.

Thus, I think it's probably safer/better in the long-run to remove the option to promote block actions, and a necessity to delete the demotion side. I've added asserts so we can know where this happens, and I was considering just nilling out the block when this happens to avoid undefined behavior at runtime.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -133,16 +133,12 @@ public:
   /** We support promotion from actions that take no arguments. */
   template <typename... Ts>
   CKTypedComponentAction<Ts...>(const CKTypedComponentAction<> &action) noexcept : CKTypedComponentActionBase(action) {
-    if (_variant == CKTypedComponentActionVariant::Block) {
-      void (^oldBlock)(CKComponent *) = (void (^)(CKComponent *))action._block;
-      _block = ^(CKComponent *sender, Ts...) { oldBlock(sender); };
-    }
     // At runtime if we provide more arguments to a block on invocation than accepted by the block, the behavior is
     // undefined. If you hit this assert, it means somewhere in your code you're doing this:
     // CKTypedComponentAction<BOOL, int> = ^(CKComponent *sender) {
     // To fix the error, you must handle all arguments:
     // CKTypedComponentAction<BOOL, int> = ^(CKComponent *sender, BOOL foo, int bar) {
-    CKCAssert(_variant != CKTypedComponentActionVariant::Block, @"Block actions should not be promoted, you will not receive arguments.");
+    CKCAssert(_variant != CKTypedComponentActionVariant::Block, @"Block actions should not take fewer arguments than defined in the declaration of the action, you are depending on undefined behavior and will cause crashes.");
   };
 
   /**
@@ -151,7 +147,7 @@ public:
    */
   template<typename... Ts>
   explicit CKTypedComponentAction<>(const CKTypedComponentAction<Ts...> &action) noexcept : CKTypedComponentActionBase(action) {
-    CKCAssert(_variant != CKTypedComponentActionVariant::Block, @"Block actions cannot be promoted, you are depending on undefined behavior and will cause crashes.");
+    CKCAssert(_variant != CKTypedComponentActionVariant::Block, @"Block actions cannot take fewer arguments than provided in the declaration of the action, you are depending on undefined behavior and will cause crashes.");
   };
 
   ~CKTypedComponentAction() {};
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -72,10 +72,6 @@
              ...
              - (void)methodWithSender:(CKComponent *)sender {}
  
- Option 3 - (HIGHLY DISCOURAGED) Block action. We support passing blocks to actions through the actionForBlock() static
-            function, however it is likely to lead to retain cycles unless carefully managed. Avoid this unless there
-            is no other option.
- 
  When using the action, simply use the send() function with the sender, an optional behavior parameter, and the
  arguments defined in the declaration of the action.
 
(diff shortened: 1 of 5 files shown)


ocrickard:
Whoops, didn't mean to close.

@facebook-github-bot import

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -95,6 +95,9 @@ class CKTypedComponentAction : public CKTypedComponentActionBase {
                 CKTypedComponentActionBoolPack<(std::is_trivially_constructible<T>::value || std::is_pointer<T>::value)...>,
                 CKTypedComponentActionBoolPack<(CKTypedComponentActionDenyType<T>::value)...>
                 >::value, "You must either use a pointer (like an NSObject) or a trivially constructible type. Complex types are not allowed as arguments of component actions.");
+  
+  CKTypedComponentAction<T...>(void(^block)(CKComponent *, T...)) noexcept : CKTypedComponentActionBase((dispatch_block_t)block) {};
```
ocrickard: We hide the constructor since really we'd like to avoid people using this. It's meant to be used in infrastructure where nothing else will work.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -57,11 +61,11 @@ class CKTypedComponentActionBase {
 
   bool operator==(const CKTypedComponentActionBase& rhs) const;
 
-private:
   // Destroying this field calls objc_destroyWeak. Since this is the only field
   // that runs code on destruction, making this field the first field of this
   // object saves an offset calculation instruction in the destructor.
   __weak id _targetOrScopeHandle;
+  dispatch_block_t _block;
```
ocrickard: If binary bloat of this is too high, then we could consider taking the scope handle/target and moving it *into* the block? That may have other side-effects though, and I'm not sure exactly what would happen to our demotion/promotion callsites.

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -125,21 +132,39 @@ class CKTypedComponentAction : public CKTypedComponentActionBase {
 
   /** We support promotion from actions that take no arguments. */
   template <typename... Ts>
-  CKTypedComponentAction<Ts...>(const CKTypedComponentAction<> &action) noexcept : CKTypedComponentActionBase(action) { };
+  CKTypedComponentAction<Ts...>(const CKTypedComponentAction<> &action) noexcept : CKTypedComponentActionBase(action) {
+    if (_variant == CKTypedComponentActionVariant::Block) {
```
ocrickard: It sucks checking the variant here. It'd be great if we could bump this out. Perhaps all this would move into a templated function in the internal header that this just calls to keep the inline bloat here to a minimum.

ComponentKit/Utilities/CKComponentActionInternal.h
dshahidehpour: What if we `union`-ized these properties?

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -49,18 +51,23 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
       return _targetOrScopeHandle;
     case CKTypedComponentActionVariant::ComponentScope:
       return ((CKComponentScopeHandle *) _targetOrScopeHandle).responder;
+    case CKTypedComponentActionVariant::Block:
+      CKCFailAssert(@"Should not be asking for target for block action.");
```
dshahidehpour: woodshed: "Block actions should never ask for a target"

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
```diff
@@ -44,7 +44,9 @@ + (instancetype)newWithQuote:(Quote *)quote
    [CKStackLayoutComponent
     newWithView:{
       [UIView class],
-      {CKComponentTapGestureAttribute(@selector(didTap))}
+      {CKComponentTapGestureAttribute(CKTypedComponentAction<UIGestureRecognizer *>::actionFromBlock(^(CKComponent *sender, UIGestureRecognizer *recognizer) {
+        NSLog(@"action!");
```
dshahidehpour: nit: just flagging in case your didn't want to keep it

ComponentKit/Utilities/CKComponentActionInternal.h
ocrickard: ARC-managed id types I don't think can be kept in a union?

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
ocrickard: yeah, definitely have to remove.

ComponentKit/Utilities/CKComponentActionInternal.h
dshahidehpour: Yeah, you're right.

Would it be out of the question to convert CKComponentAction to a NSObject (not in this diff, just, in general)

ocrickard: I'm thinking that may be what I do if this has a big impact on binary size. We'd switch to an Obj-C wrapper for these params that's constructed with a C function. Then as we pass around this struct it really just becomes a single strong pointer in the actual copy constructors.

ocrickard:
Binary size looks OK! Let's clean this up and land it then.

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

ocrickard:
@facebook-github-bot import

PR #798, opened 2017-04-20
Expose Custom Controller Implementation, add more typing to ScopeHandle

dshahidehpour:
This builds upon @ocrickard's change to make CKComponentScope more less reliant on concrete types (i.e. `id<CKScopeComponent>` instead of `CKComponent`). This builds upon it and decouples `CKComponentController` concrete implementation from `CKComponentScopeHandle`

Sorry there are a million commits on this, it'll be more clear once the other PRs have landed. Here's what you should be interested in: e81dccd

### DIFF
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -297,7 +297,7 @@ static void *kRootComponentMountedViewKey = &kRootComponentMountedViewKey;
 
 + (Class<CKScopedComponentController>)controllerClass
 {
-  const Class componentClass = self.class;
+  const Class componentClass = self;
 
   if (componentClass == [CKComponent class]) {
     return Nil; // Don't create root CKComponentControllers as it does nothing interesting.
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -166,10 +166,11 @@ struct CKComponentMountInfo {
 {
   CKAssertMainThread();
   if (_mountInfo != nullptr) {
-    [_scopeHandle.controller componentWillUnmount:self];
+    CKComponentController *const controller = (CKComponentController *)_scopeHandle.controller;
+    [controller componentWillUnmount:self];
     [self _relinquishMountedView];
     _mountInfo.reset();
-    [_scopeHandle.controller componentDidUnmount:self];
+    [controller componentDidUnmount:self];
   }
 }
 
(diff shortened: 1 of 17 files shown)


dshahidehpour:
@facebook-github-bot import

ComponentKit/Core/CKComponent.mm
```diff
@@ -166,10 +166,11 @@ - (void)unmount
 {
   CKAssertMainThread();
   if (_mountInfo != nullptr) {
-    [_scopeHandle.controller componentWillUnmount:self];
+    CKComponentController *const controller = (CKComponentController *)_scopeHandle.controller;
```
eczarny: Is there a way we can use Objective-C generics on the scope handle to avoid these casts?

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -261,12 +261,14 @@ static void checkMethodSignatureAgainstTypeEncodings(SEL selector, NSMethodSigna
 void _CKTypedComponentDebugCheckComponentScope(const CKComponentScope &scope, SEL selector, const std::vector<const char *> &typeEncodings) noexcept
 {
 #if DEBUG
+  const CKComponentScopeHandle *scopeHandle = scope.scopeHandle();
```
eczarny: `const` should follow the asterisk here:

```
CKComponentScopeHandle *const scopeHandle = scope.scopeHandle();
```

ComponentKit/Core/CKComponentController.h
```diff
@@ -89,3 +91,14 @@
 + (instancetype)new NS_UNAVAILABLE;
 
 @end
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+Class<CKScopedComponentController> CKComponentControllerClassFromComponentClass(Class<CKScopedComponent> componentClass) noexcept;
```
eczarny: We're probably going to want documentation here, and on any other public API.

ComponentKit/Core/CKComponent.mm
ocrickard: That's a good idea. I think scope handles having a generic interface could really help us avoid casts since most of the interop with the scope infra is through the scope handles.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -65,3 +66,9 @@
 - (id)responder;
 
 @end
+
+@interface CKComponentScopeHandle (Debug)
+
+- (CKScopedComponentControllerClassGenerator)controllerClassGenerator;
+
+@end
```
ocrickard: As noted internally, can we move this to a debug-only header?

ComponentKit/Core/CKComponent.mm
dshahidehpour: Love it.

eczarny:
@dshahidehpour `const CKComponentScopeHandle *scopeHandle = scope.scopeHandle();` still isn't quite right but the rest of this looks amazing! Accepting here since it was accepted internally.

PR #801, opened 2017-04-27
Trivial blocks

ocrickard:
I used is_trivially_constructible, but really what we need here are things that work with NSInvocation. dispatch blocks should be totally fine to pass to NSInvocation, it has fully support in the language. Yet, is_trivially_constructible returns the false type.

This allows CKTypedComponentAction<dispatch_block_t>, but still forbids CKTypedComponentAction<std::vector<int>>, since that would result in undefined behavior when calling the method with a complex type. We allow references, pointers, things that convert to id, and trivially constructible types.


indragiek:
Awesome, thanks for implementing!

ocrickard:
@facebook-github-bot import

PR #802, opened 2017-04-28
Remove int+long CKTypedComponentAction constructors

toulouse:
* `NULL` can possibly be `0`, `0L`, or `nullptr_t`, depending on implementation or whether using C or C++.
* `CKTypedComponentAction` has multiple constructors which accept one parameter:
  * `SEL`: essentially `char *`, which is pointer-sized
  * `int`: probably pointer sized
  * `long`: pointer sized
  * `dispatch_block_t`: IIRC, either an object (Obj-C) or a lambda (C++).
  * `std::nullptr_t`: typesafe `NULL`

A `NULL` result in a ternary expression that provides a value to a struct with a `CKTypedComponentAction<CKComponent *>` field is seemingly allowed by clang but may result in undefined behavior.

Basically: the ternary was supposed to return a `CKTypedComponentAction<CKComponent *>`, i.e. a value, not a pointer. Sensibly, the non-`NULL` case calls the copy constructor, so it makes sense that the `NULL` case would call a constructor appropriate for `NULL`. The question is...which one? There are several which are pointer-sized, including two numeric.

The answer is that the constructor that takes `long` is chosen. I don't think that this is a satisfying result. The conclusion of this is that it seems prudent to remove the `int` and `long` constructors (as they serve no purpose other to accept `NULL` for convenience), preserve the `std::nullptr_t` constructor, and modify all `CKComponentAction` parameters to use `nullptr` instead of `NULL`.

This may be a slightly breaking change for the API in two ways:

* `NULL` is no longer accepted as an explicit action argument
  * However, the remediation is very easy - change that `NULL` to a `nullptr`.
* Some arguments may be ambiguous
  * It's not a bug, it's a feature! ;) The ambiguity should surface the previously-undefined behavior, allowing developer judgment to resolve it with a cast, `nullptr` replacement, or other modification.

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -165,7 +165,7 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
                             std::move(attributes),
                             {
                               .accessibilityLabel = accessibilityConfiguration.accessibilityLabel,
-                              .accessibilityComponentAction = enabled ? CKComponentAction(action) : NULL
+                              .accessibilityComponentAction = enabled ? CKComponentAction(action) : nullptr
                             }
                           }
                           size:size];
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -133,9 +133,10 @@ public:
     return CKTypedComponentAction<T...>(block);
   }
 
-  /** Allows conversion from NULL actions. */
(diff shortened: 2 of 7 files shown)


toulouse:
fixing

ocrickard:
This is fantastic. Thanks for doing this!

ocrickard:
@facebook-github-bot import

ocrickard:
@toulouse I'm importing this, you should probably commandeer the diff internally if anything doesn't build with this.

ocrickard:
Merged internally. Thanks for this!

PR #803, opened 2017-05-02
CKComponentKey: allow parents to assign keys to children

adamjernst:
We want the ability for parents to assign keys to their children, instead of requiring children to assign keys to themselves using `CKComponentScope`. The new `CKComponentKey` struct allows doing this.

Example usage:

```
[CKStackLayoutComponent
 newWithChildren:CK::map(models, ^(XXPerson *person) {
   CKComponenentKey key(person.uniqueIdentifier);
   return [XXPersonComponent newWithPerson:person];
 }]
```

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeFrame.mm
@@ -22,20 +22,23 @@
 #import "CKMacros.h"
 #import "CKThreadLocalComponentScope.h"
 
-typedef struct _CKStateScopeKey {
+struct CKStateScopeKey {
   Class __unsafe_unretained componentClass;
   id identifier;
+  NSArray<id> *keys;
 
-  bool operator==(const _CKStateScopeKey &v) const {
-    return (CKObjectIsEqual(this->componentClass, v.componentClass) && CKObjectIsEqual(this->identifier, v.identifier));
+  bool operator==(const CKStateScopeKey &v) const {
+    return (CKObjectIsEqual(this->componentClass, v.componentClass)
+            && CKObjectIsEqual(this->identifier, v.identifier)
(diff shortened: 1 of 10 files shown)


GotEmB:
This looks awesome!

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -22,19 +22,22 @@
...
-} _CKStateScopeKey;
+};
 
 namespace std {
   template <>
-  struct hash<_CKStateScopeKey> {
-    size_t operator ()(_CKStateScopeKey k) const {
+  struct hash<CKStateScopeKey> {
+    size_t operator ()(CKStateScopeKey k) const {
       NSUInteger subhashes[] = { [k.componentClass hash], [k.identifier hash] };
```
ocrickard: No influence on the hash of the state scope key?

ocrickard: (with obligatory note that NSArray's hash is its count)

ocrickard:
This looks pretty good to me. Just one question about the array of keys not being included in the hash of the state scope key. Probably not an issue, but if the other params start colliding more, then we can start just depending on linear comparisons of the arrays when looking up state. If we had a huge hierarchy of components, could be a problem. As-is today, probably fine.

ocrickard:
Also, tests would be nice

ocrickard:
(accidentally closed, sorry)

ComponentKit/Core/Scope/CKComponentScopeFrame.mm
adamjernst: Oh heh. Will fix.

ocrickard:

PR #805, opened 2017-05-03
Mark promotion constructor explicit

ocrickard:
I wanted to remove this constructor anyway since it's confusing. It allows a component declared in its header as a CKComponentAction to any other action. Then at runtime, that action can just ignore all the params. This gets real tough with block actions, which we don't want to be able to just ignore parameters. We'd like to start failing to compile if anyone is depending on this. Instead we'll force them to do the conversion explicitly.

So this means this code will no longer compile:

```
+ (instancetype)newWithAction:(CKComponentAction)action
{
  return [super newWithComponent:
    [SomeOtherComponent
     newWithTypedAction:action]]; // action type is CKTypedComponentAction<UIEvent *>
```

Instead, you must do this to allow promotion:

```
+ (instancetype)newWithAction:(CKComponentAction)action
{
  return [super newWithComponent:
    [SomeOtherComponent
     newWithTypedAction:CKTypedComponentAction<UIEvent *>(action)]];
```

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -138,9 +138,9 @@ public:
   CKTypedComponentAction(long s) noexcept : CKTypedComponentActionBase() {};
   CKTypedComponentAction(std::nullptr_t n) noexcept : CKTypedComponentActionBase() {};
 
-  /** We support promotion from actions that take no arguments. */
+  /** We support promotion from actions that take no arguments, but only when explicit. */
   template <typename... Ts>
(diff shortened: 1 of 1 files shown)


ocrickard:
@facebook-github-bot import

ocrickard:
Alright, so almost all of the usages internally for this is in CKComponentActionSend(). So I'm gonna push through and remove CKComponentActionSend, and just force people to use .send().

eczarny:
@ocrickard merge when ready.

ocrickard:
This is on my plate for next half. Gonna be a huge codemod.

ocrickard:
Using the new sync here.

PR #807, opened 2017-05-05
Optimize CKComponentKey

adamjernst:
In internal benchmarks the previous implementation caused a performance regression, which makes sense since it was fairly wasteful. Attempt to be more efficient.


ComponentKit/Core/Scope/CKComponentScopeFrame.mm
```diff
@@ -22,23 +23,35 @@
...
 
+static bool keyVectorsEqual(const std::vector<id<NSObject>> &a, const std::vector<id<NSObject>> &b)
+{
+  if (a.size() != b.size()) {
+    return false;
+  }
+  return std::equal(a.begin(), a.end(), b.begin(), [](id<NSObject> x, id<NSObject> y){
+    return CKObjectIsEqual(x, y) == YES; // be pedantic here becuase BOOL != bool
+  });
+}
```
adamjernst: Decided to skip crazy template specialization stuff here. I want the explicitness of definitely using CKObjectIsEqual.

PR #808, opened 2017-05-05
Don't compare to YES explicitly

adamjernst:
On second thought, this is dangerous:

http://stackoverflow.com/questions/1934580/is-comparing-a-bool-against-yes-dangerous


dshahidehpour:
@facebook-github-bot shipit

PR #813, opened 2017-05-14
Remove CKComponent class assertion in CKComponentScopeFrame.

dshahidehpour:
Now that we have `CKScopedComponent` this assert will falsely fire. Removing.


dshahidehpour:
@facebook-github-bot shipit

PR #814, opened 2017-05-16
Keep an accurate responder chain thru memoization layer via CKScopedResponderManager

dshahidehpour:
All Unit Tests in place are passing, I'll add some more in the morning.


dshahidehpour:
I'm taking this back to my woodshed, will reopen in a bit!

PR #815, opened 2017-05-16
Introduce concept of nested thread local scopes.

dshahidehpour:
If another `CKThreadLocalComponentScope` is created while another is in scope, it is overridden. This changes behavior so that the newly created scope takes precedence as `currentScope` for the duration of its' lifecycle. When it is destructed, the previous `currentScope` is restored.

### DIFF
--- a/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
+++ b/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
@@ -31,5 +31,5 @@ public:
   std::stack<std::vector<id<NSObject>>> keys;
 
 private:
-  CKThreadLocalComponentScope *previousScope;
+  CKThreadLocalComponentScope *const previousScope;
 };
--- a/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
+++ b/ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
@@ -29,17 +29,7 @@ public:
   const CKComponentStateUpdateMap stateUpdates;
   std::stack<CKComponentScopeFramePair> stack;
   std::stack<std::vector<id<NSObject>>> keys;
-};
-
-/**
- Temporarily overrides the current thread's component scope.
- Use for testing and advanced integration purposes only.
- */
-class CKThreadLocalComponentScopeOverride {
-public:
-  CKThreadLocalComponentScopeOverride(CKThreadLocalComponentScope *scope) noexcept;
-  ~CKThreadLocalComponentScopeOverride();
 
 private:
-  CKThreadLocalComponentScope *const previousScope;
+  CKThreadLocalComponentScope *previousScope;
 };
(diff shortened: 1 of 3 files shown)


ComponentKit/Core/Scope/CKThreadLocalComponentScope.mm
```diff
@@ -27,36 +27,46 @@ static pthread_key_t _threadKey() noexcept
   return thread_key;
 }
 
+static std::stack<CKThreadLocalComponentScope *> *_threadLocalStack() noexcept
```
ocrickard: nit: we avoid the underscore in static functions nowadays.

ComponentKit/Core/Scope/CKThreadLocalComponentScope.mm
```diff
@@ -27,36 +27,46 @@ static pthread_key_t _threadKey() noexcept
...
+{
+  return static_cast<std::stack<CKThreadLocalComponentScope *> *>(pthread_getspecific(_threadKey()));
```
ocrickard: This should work just fine, but another way to do this is to just grab the thread-local scope pointer and hold onto it in a member of CKThreadLocalComponentScope, then re-set it in the destructor.

ComponentKit/Core/Scope/CKThreadLocalComponentScope.h
```diff
@@ -30,16 +30,3 @@ class CKThreadLocalComponentScope {
...
- Temporarily overrides the current thread's component scope.
- Use for testing and advanced integration purposes only.
- */
-class CKThreadLocalComponentScopeOverride {
-public:
-  CKThreadLocalComponentScopeOverride(CKThreadLocalComponentScope *scope) noexcept;
-  ~CKThreadLocalComponentScopeOverride();
-
-private:
-  CKThreadLocalComponentScope *const previousScope;
```
ocrickard: Yeah, why not just do this?

ComponentKit/Core/Scope/CKThreadLocalComponentScope.mm
```diff
@@ -27,36 +27,46 @@ static pthread_key_t _threadKey() noexcept
...
 {
-  CKCAssert(CKThreadLocalComponentScope::currentScope() == nullptr, @"CKThreadLocalComponentScope already exists");
   stack.push({[newScopeRoot rootFrame], [previousScopeRoot rootFrame]});
   keys.push({});
-  pthread_setspecific(_threadKey(), this);
+
+  auto threadLocalStack = _threadLocalStack();
+  if (threadLocalStack == nullptr) {
+    threadLocalStack = new std::stack<CKThreadLocalComponentScope *>();
+    pthread_setspecific(_threadKey(), threadLocalStack);
```
ocrickard: There is thread-local storage in C++, not sure if we can use it here. I've never used it: http://en.cppreference.com/w/cpp/keyword/thread_local

ocrickard:
Looks correct to me, but it does seem a bit more complex than the override approach which I kinda think is simpler and accomplishes the same thing, right?

dshahidehpour:
@ocrickard Great idea!

ocrickard:
Looks great to me!

dshahidehpour:
@facebook-github-bot shipit

PR #816, opened 2017-05-16
Allow component subclasses to obtain a reference to the view used to render animations

eczarny:
`-[CKComponent viewForAnimation]` and `-[CKCompositeComponent viewForAnimation]` can, and often are, used to support more complex animations in a component hierarchy. Given this reality move the declaration of this method out of `CKComponentInternal.h` and into `CKComponentSubclass.h`.


eczarny:
@facebook-github-bot import

ocrickard:
Yep, been meaning to do this.

PR #817, opened 2017-05-17
Modify ScopeHandle so it always returns a responder that is not stale

dshahidehpour:


ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -27,6 +27,7 @@ @implementation CKComponentScopeHandle
   BOOL _acquired;
   BOOL _resolved;
   CKComponent *__weak _acquiredComponent;
+  id __weak _responder;
```
dshahidehpour: cc: @adamjernst @ocrickard

dshahidehpour:
I *could* have kept it open, but I just decided to close it.  I need to write some tests for, and my approach of passing along a block is not going to work.

PR #818, opened 2017-05-17
Modify ScopeHandle so it always returns a responder that is not stale

dshahidehpour:


ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -62,8 +63,17 @@
...
 
 /**
- Provides a responder corresponding with this scope handle. The controller will assert if called before resolution.
+ Provides a block that, when called, will return the responder associated with the "current" component generation.
  */
-- (id)responder;
+- (CKScopedResponder *)scopedResponder;
+
+/**
+ Used by the infra. Don't call unless you know what you are doing.
```
ocrickard: A description of what this could be used for/how to use it would help us maintain this I think.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -62,8 +63,17 @@
 @property (nonatomic, readonly) CKComponentScopeHandleIdentifier globalIdentifier;
 
 /**
- Provides a responder corresponding with this scope handle. The controller will assert if called before resolution.
+ Provides a block that, when called, will return the responder associated with the "current" component generation.
```
ocrickard: Not quite sure I understand why we need this change.

ComponentKit/Core/CKComponent.mm
```diff
@@ -193,6 +193,7 @@ - (void)_relinquishMountedView
 
 - (void)childrenDidMount
 {
+  [_scopeHandle assignNewResponder];
```
ocrickard: Wow, this is so much simpler than what I was thinking we'd do. Love it.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -19,6 +19,30 @@
 #import "CKScopedComponentController.h"
 #import "CKThreadLocalComponentScope.h"
 
+@interface CKScopedResponder ()
+@property (nonatomic, weak, readwrite) id responder;
+@end
+
+@implementation CKScopedResponder
+
+- (BOOL)isEqual:(id)object
```
ocrickard: why?

ComponentKit/Core/Scope/CKComponentScopeHandle.h
dshahidehpour: (y)

ComponentKit/Core/Scope/CKComponentScopeHandle.h
dshahidehpour: Didn't, I'll revert.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
dshahidehpour: feels like standard procedure for NSObject, but I can kill it

ocrickard:
Alright, I've been thinking about this: what if the scoped responder associated with a scope handle held a threadsafe queue of weak component generations. When a new generation of components is created, the new scope handles push themselves onto the scoped responder queue that's associated with that scope. When we go to get a responder, we just take the first non-nil element in that queue as the responder.

As older generations are torn down, they are automatically dropped from the queue.

This has a few nice properties:

1. No new lifecycle event, yet it works before mount.
2. More localized and requires no changes to existing systems I think?

One source of complexity will be in ensuring forward directionality of the scoped responders. Each new generation's queue should somehow start at themselves, but then all subsequent generation should be added to the back of their queues. Maybe some sort of persistent data structure like a singly linked list with nodes holding weak pointers?

dshahidehpour:
> When a new generation of components is created, the new scope handles push themselves onto the scoped responder queue that's associated with that scope.

@ocrickard so if I'm understanding you, the Scoped Responder will hold a queue of weak references to Scope Handles. Everytime a new handle is generated (with the same) global identifier, it is added to the queue. When we ask for the responder, we traverse the queue until we find a handle with a non-nil view, correct?

dshahidehpour:
@ocrickard I think I'm implemented what you suggested, and I'm not sure it will work.

Checkout this branch: https://github.com/dshahidehpour/componentkit/commit/0bc044df1f3cfe0e0750c30d08ef99ee5a9bbd31

Here's what I see being the flaw:

```
- (void)test
{
  CKComponent *(^block)(void) = ^ {
    return [TestComponentWithScope new];
  }
  const CKBuildComponentResult firstResult = CKBuildComponent(........);
  
  @autoreleasepool {
     const CKBuildComponentResult secondResult= CKBuildComponent(firstResult.scopeRoot, {}, block);
     XCTAssertEquals([firstResult.component.handle responder], secondResult.component]);
  }
  XCTAssertEquals([firstResult.component.handle responder], firstResult.component);
}
```

You can see that if we generate another hierarchy somewhere, once the new handle acquires a view, it is essentially the first responder. We don't really want to update the responder chain until the new component is mounted, otherwise all the handles with the same global identifier would temporarily point to the wrong component.

dshahidehpour:
Talked with @adamjernst offline. Changed it to use the "oldest" component that has an acquired.

dshahidehpour:
@facebook-github-bot import

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -59,7 +67,8 @@ - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
                  rootIdentifier:rootIdentifier
                  componentClass:componentClass
                           state:initialStateCreator ? initialStateCreator() : [componentClass initialState]
-                     controller:nil]; // controllers are built on resolution of the handle
+                     controller:nil
+                scopedResponder:[CKScopedResponder new]]; // controllers are built on resolution of the handle
```
adamjernst: nit: comment up one line

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -76,6 +86,9 @@ - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
     _componentClass = componentClass;
     _state = state;
     _controller = controller;
+
+    _scopedResponder = scopedResponder;
+    [scopedResponder addHandleToChain:self];
```
adamjernst: (y)

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -171,10 +186,50 @@ - (void)resolve
...
+    });
+  }
+}
+
 - (id)responder
 {
-  CKAssert(_resolved, @"Asking for responder from scope handle before resolution:%@", NSStringFromClass(_componentClass));
-  return _acquiredComponent;
+  __block id<CKScopedComponent> responder = nil;
+  dispatch_sync(_responderQueue, ^{
```
adamjernst: Nuke the dispatch_sync. Just use std::mutex for synchronization.

adamjernst:
Use std::mutex instead of dispatch queues for synchronization. Otherwise, looks good!

adamjernst:
ship it!

PR #819, opened 2017-05-19
Use NSHashTable to avoid registering duplicate component/controller

supercfanx:
Currently, if we register the same component or controller twice (which could happen in certain cases when a new scope handle is created before the controller is resolved?), appearance callbacks will be invoked twice. Fix this by using a `NSHashTable` to store the component/controller.

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeRoot.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeRoot.mm
@@ -17,8 +17,8 @@
 #import "CKInternalHelpers.h"
 #import "CKThreadLocalComponentScope.h"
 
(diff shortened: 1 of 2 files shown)


ocrickard:
Awesome, thanks for fixing this!

ocrickard:
Merge when you're ready!

supercfanx:
@facebook-github-bot shipit

supercfanx:
@ocrickard I don't see a merge option here, maybe I don't have the privilege?

dshahidehpour:
@facebook-github-bot shipit

adamjernst:
@facebook-github-bot shipit

PR #821, opened 2017-05-31
[CKComponentAction] Support block based action promotion/demotion

hehe-wz:


ocrickard:
Looks like the sample project is failing to build now.

ocrickard:
Sorry I've been slow to review this. @adamjernst brought up that a currying approach is probably preferable, so I'd really like to consider going with that instead. I'm planning on implementing soonish.

danzimm:
I have now implemented the currying approach you spoke of @ocrickard 

Just realized I should have had you on the diff - weih took care of the review for me

ocrickard:
Alright, I believe this has been addressed for the demote side. We probably just need to remove the assert for promotion, it should be just fine.

PR #822, opened 2017-06-02
Fix Scoped Component Actions so that they have unique identifiers.

dshahidehpour:
We are seeing a problem in the Facebook app because Scoped Component Actions that use the same selector are overriding each other within `CKComponentAttributeValueMap`. This fixes that!

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.h
@@ -69,6 +69,11 @@
 
 @end
 
+typedef int32_t CKScopedResponderUniqueIdentifier;
+
 @interface CKScopedResponder : NSObject
+
+@property (nonatomic, readonly) CKScopedResponderUniqueIdentifier uniqueIdentifier;
+
 - (id)responder;
 @end
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.h
@@ -73,7 +73,8 @@ typedef int32_t CKScopedResponderUniqueIdentifier;
 
 @interface CKScopedResponder : NSObject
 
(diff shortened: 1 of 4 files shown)


ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -69,6 +69,11 @@
 
 @end
 
+typedef int32_t CKScopedResponderUniqueIdentifier;
+
 @interface CKScopedResponder : NSObject
+
+@property (nonatomic, readonly) CKScopedResponderUniqueIdentifier uniqueIdentifier;
```
eczarny: Nit: I'd include `assign` to be explicit.

ComponentKit/Core/Scope/CKComponentScopeHandle.h
```diff
@@ -69,6 +69,11 @@
...
+
 - (id)responder;
 @end
```
eczarny: Nit: A newline above this line would be great.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -201,6 +201,16 @@ @implementation CKScopedResponder
   std::mutex _mutex;
 }
 
+- (instancetype)init
+{
+  if (self = [super init]) {
+    static CKScopedResponderUniqueIdentifier nextIdentifier = 0;
+    _uniqueIdentifier = OSAtomicIncrement32(&nextIdentifier);
```
ocrickard: I guess this is maybe better than just using the pointer value of the responder, which would probably work as well without introducing as much code.

ocrickard:
> Has this always been a problem or is it something that surfaces now given new usage patterns?

I think this issue was introduced when @dshahidehpour split the targetOrScopeHandle member var into two values again, but forgot to fix up the identifier.

ocrickard:
Looks good to me! Merge when ready.

PR #824, opened 2017-06-02
Only create a scoped responder the first time we need it.

dshahidehpour:
Great idea by @ocrickard. Instead of creating a responder when a handle with a new identifier is created, we wait until the first time someone requests a responder.

I tested this in News Feed, and it reduced the number of allocations by 8-10x (depending on the content in feed). Woot!


ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -190,6 +190,13 @@ - (void)resolve
 
 - (CKScopedResponder *)scopedResponder
 {
+  CKAssertFalse(_resolved);
```
ocrickard: Looks pretty reasonable to me! I may move this assertion inside the if statement...

PR #825, opened 2017-06-06
[Accessibility] Add support for hint, value and traits in accessibility context

ikenwoo:
Accessibility hint, value and traits are common properties used for accessibility. 

This will add support for them in `CKComponentAccessibilityContext` so that they won't be applied as view attributes when accessibility isn't enabled. This will also help get rid of mixed accessibility scenarios like: 

```Objective-C++
[CKComponent newWithView:{[UIView view]}, 
{
  {@selector(setAccessibilityHint:), @"Double-tap to open"}, 
  {@selector(setAccessibilityTraits:), @(UIAccessibilityTraitButton)}
}, 
{
  .isAccessibilityElement = @YES, 
  .accessibilityLabel = @"Open"
}}];
```

This also updates the existing tests and adds two tests that ensures that information in an `accessibilityContext` is only applied as view properties when accessibility is enabled.


dshahidehpour:
@facebook-github-bot import

PR #829, opened 2017-06-07
Prevent Scoped Responder from returning an old component that is still alive.

dshahidehpour:
We are seeing an issue where old component generations are still being kept around (whether from retain cycle, or some funny client code) that is causing us to choose an responder that is incorrect.

Here's a scenario:
Within `CKScopedResponder` we keep a vector of `__weak CKComponentScopeHandle *`

In the event an old component generation wasn't dealloced. We could see something like this:
```
{ ScopeHandleA, nil, nil, nil, nil, nil ScopeHandleF, ScopeHandleG }
```
Currently, we simply choose the oldest handle as a responder (`ScopeHandleA`), when in reality, we want `ScopeHandleF`.

Now, when we create a `CKComponentAction` we store a bookmark to the oldest possible component we would want to look at. So, after this commit lands we would create an action that says, "Give me the oldest ScopeHandle that is alive, but start at `ScopeHandleF`".

This fixes our bug internally.

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -227,10 +227,38 @@
   _handles.push_back(handle);
 }
 
-- (id)responder
+- (CKScopedResponderKey)keyForHandle:(CKComponentScopeHandle *)handle
+{
+  static const CKScopedResponderKey notFoundKey = INT_MAX;
+
+  if (handle == nil) {
+    return notFoundKey;
+  }
+
+  std::lock_guard<std::mutex> l(_mutex);
+  auto result = std::find(_handles.begin(), _handles.end(), handle);
+
+  if (result == _handles.end()) {
+    CKFailAssert(@"This scope handle is not associated with this Responder.");
+    return notFoundKey;
+  }
+
+  // Returning the index of an element in a vector: https://stackoverflow.com/a/15099743
+  return (int)std::distance(_handles.begin(), result);
+}
+
+- (id)responderForKey:(CKScopedResponderKey)key
 {
   std::lock_guard<std::mutex> l(_mutex);
-  for (const auto &handle: _handles) {
+
+  const size_t numberOfHandles = _handles.size();
+  if (key < 0 || key >= numberOfHandles) {
+    CKFailAssert(@"Invalid key \"%d\" for responder with %d handles", key, numberOfHandles);
+    return nil;
+  }
+
+  for (int i = key; i < numberOfHandles; i++) {
+      const auto handle = _handles[i];
       const id<CKScopedComponent> responder = handle.acquiredComponent;
       if (responder != nil) {
         return responder;
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -253,7 +253,7 @@
 
   const size_t numberOfHandles = _handles.size();
   if (key < 0 || key >= numberOfHandles) {
-    CKFailAssert(@"Invalid key \"%d\" for responder with %d handles", key, numberOfHandles);
+    CKFailAssert(@"Invalid key \"%d\" for responder with %zu handles", key, numberOfHandles);
     return nil;
   }
 
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -30,7 +30,7 @@ bool CKTypedComponentActionBase::operator==(const CKTypedComponentActionBase& rh
 {
   return (_variant == rhs._variant
           && CKObjectIsEqual(_target, rhs._target)
-          && _scopedResponder == rhs._scopedResponder
+          && _scopeIdentifierAndResponderGenerator == rhs._scopeIdentifierAndResponderGenerator
           && _selector == rhs._selector
           && _block == rhs._block);
 }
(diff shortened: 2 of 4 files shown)


ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,34 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
...
+- (id)responderForKey:(CKScopedResponderKey)key
+{
+  if (key < 0) {
+    CKFailAssert(@"In valid key %d for responder.", key);
+    return nil;
+  }
+
   std::lock_guard<std::mutex> l(_mutex);
-  for (const auto &handle: _handles) {
+  for (int i=key; i<_handles.size(); i++) {
```
dshahidehpour: I don't feel good about this at all, would love suggestions.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,34 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
...
+      const auto handle = _handles[i];
```
eczarny: We should probably be a bit defensive here and bounds check the index.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
eczarny: If `key` is `INT_MAX` then this method just returns nil, right? Just making sure.

ComponentKit/Utilities/CKComponentActionInternal.h
```diff
@@ -65,7 +66,7 @@ class CKTypedComponentActionBase {
   // that runs code on destruction, making this field the first field of this
   // object saves an offset calculation instruction in the destructor.
   __weak id _target;
-  CKScopedResponder *_scopedResponder;
+  std::pair<CKScopedResponderUniqueIdentifier, CKResponderGenerationBlock> _scopeIDAndGenerator;
```
eczarny: `_scopeIDAndGenerator` isn't terribly descriptive. Maybe `_scopeIdentifierAndResponderGenerator`? I dropped the ID since I'm not really a fan of the shorthand.`

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -50,35 +50,45 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
     case CKTypedComponentActionVariant::TargetSelector:
       return _target;
     case CKTypedComponentActionVariant::Responder:
-      return _scopedResponder.responder;
+      return _scopeIDAndGenerator.second();
```
eczarny: Should this be nil-checked?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,34 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
...
+  }
+
+  std::lock_guard<std::mutex> l(_mutex);
+  auto result = std::find(_handles.begin(), _handles.end(), handle);
+  if (result == _handles.end()) {
+    CKFailAssert(@"This scope handle is not associated with this Responder.");
+    return notFoundKey;
+  }
+
+  return (int)std::distance(_handles.begin(), result);
```
eczarny: I think this warrants a comment.

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -50,35 +50,45 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
...
-  CKCAssert(_scopedResponder, @"You are creating an action that will not fire because you have an invalid scope handle.");
+  const auto handle = scope.scopeHandle();
+  CKCAssert(handle, @"You are creating an action that will not fire because you have an invalid scope handle.");
+
+  const auto scopedResponder = handle.scopedResponder;
+  const auto responderKey = [scopedResponder keyForHandle:handle];
+  _scopeIDAndGenerator = {
+    [handle globalIdentifier],
+    ^id(void) {
+      return [scopedResponder responderForKey:responderKey];
```
eczarny: A few comments here would be helpful as well. Particularly mentioning the need for the responder key and why this generator block is necessary.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
eczarny: Yeah... I'll need to chew on the problem for a while before I can offer much help.

dshahidehpour: correct!

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
dshahidehpour: (y)

ComponentKit/Utilities/CKComponentActionInternal.h
dshahidehpour: sounds great to me.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,38 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
   _handles.push_back(handle);
 }
 
-- (id)responder
+- (CKScopedResponderKey)keyForHandle:(CKComponentScopeHandle *)handle
+{
+  static CKScopedResponderKey notFoundKey = INT_MAX;
```
adamjernst: `const`

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,38 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
...
+
+  std::lock_guard<std::mutex> l(_mutex);
+  auto result = std::find(_handles.begin(), _handles.end(), handle);
+
+  if (result == _handles.end()) {
+    CKFailAssert(@"This scope handle is not associated with this Responder.");
+    return notFoundKey;
+  }
+
+  // Returning the index of an element in a vector: https://stackoverflow.com/a/15099743
```
adamjernst: I *assume* that handles is append-only, so this is safe. correct?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -227,10 +227,38 @@ - (void)addHandleToChain:(CKComponentScopeHandle *)handle
...
   std::lock_guard<std::mutex> l(_mutex);
-  for (const auto &handle: _handles) {
+
+  const size_t numberOfHandles = _handles.size();
+  if (key < 0 || key >= numberOfHandles) {
+    CKFailAssert(@"Invalid key \"%d\" for responder with %d handles", key, numberOfHandles);
+    return nil;
+  }
+
+  for (int i = key; i < numberOfHandles; i++) {
```
adamjernst: so we *start* at key and only look for *newer* generations. got it.

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -50,35 +50,54 @@ void CKConfigureInvocationWithArguments(NSInvocation *invocation, NSInteger inde
...
+
+      /** 
+       At one point in the history of ComponentKit, it was possible for a scopeResponders to
+       return a "stale" initialTarget for an action. This was often caused by retain cycles, or, 
+       "old" component hierarchies that have an extra-long lifecycle.
+       
+       To prevent this from happening in the future we now provide a key which gives the 
+       scopeResponder the wisdom to ignore older generations.
+       */
+      return [scopedResponder responderForKey:responderKey];
```
adamjernst: Love that you hid the raw indexes behind the key system.

adamjernst:
I'm increasingly unhappy with this stuff and wish we could do the "right" thing, but the right thing is way more involved for little benefit. Let's ship it.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
dshahidehpour: correct.

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
dshahidehpour: yes

dshahidehpour:
@facebook-github-bot import

PR #830, opened 2017-06-07
CKComponentLayout: cache empty vector of children

swolchok:
No reason to do a bunch of mallocs to get const empty vectors.


dshahidehpour:
@facebook-github-bot shipit

PR #831, opened 2017-06-12
Warn against capturing C++ objects passed by ref in blocks.

majak:


_docs/09-component-actions.md
```diff
@@ -108,6 +108,12 @@ Here's an example of how to handle a component action. (The API for `CKButtonCom
   </p>
 </div>
 
+<div class="note-important">
+  <p>
+    <code>CKComponentAction</code> is a C++ object. ObjC block captures C++ objects passed by reference (<code>const&</code>) by reference. That can lead to memory bugs where an action you want to trigger doesn't exist anymore. Avoid this by passing the action by value or creating a copy.
```
eczarny: I think this may be a bit clearer:

> `CKTypedComponentAction` is a C++ object which can lead to subtle problems when a `CKTypedComponentAction` reference (e.g. `CKTypedComponentAction &` or `const CKTypedComponentAction &`) is captured by a block. The reason? C++ references are not managed by ARC, which can result in crashes that are difficult to debug. The best way to avoid this problem is the pass actions by value, or by explicitly creating a copy for use within the block.

eczarny:
Thanks @majak!

PR #832, opened 2017-06-13
Gracefully handle a nil gesture recognizer class

adamjernst:
If you pass `Nil` for the gesture recognizer class, we should gracefully do nothing instead of crashing the app entirely. (`UIView addGestureRecognizer:` crashes with a nil input.)

### DIFF
--- a/ComponentKitTests/CKComponentGestureActionsTests.mm
+++ b/ComponentKitTests/CKComponentGestureActionsTests.mm
@@ -138,6 +138,16 @@
   XCTAssertNil(gesture.delegate, @"Gesture delegate should not be set");
 }
 
+- (void)testThatApplyingANilRecognizerClassResultsInNoRecognizer
+{
+  CKComponentViewAttributeValue attr = CKComponentGestureAttribute(Nil, nullptr, @selector(test));
+  UIView *view = [UIView new];
+
+  attr.first.applicator(view, attr.second);
+  XCTAssertEqual([view.gestureRecognizers count], 0u, @"Expected no gesture recognizer to be attached");
+  attr.first.unapplicator(view, attr.second);
+}
+
 - (void)testThatApplyingATapRecognizerAttributeWithDifferentTargetToViewWithExistingRecognizerUpdatesAction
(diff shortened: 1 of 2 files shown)


ComponentKitTests/CKComponentGestureActionsTests.mm
```diff
@@ -138,6 +138,16 @@ - (void)testThatWithNoDelegateActionsNoDelegateIsSet
   XCTAssertNil(gesture.delegate, @"Gesture delegate should not be set");
 }
 
+- (void)testThatApplyingANilRecognizerClassResultsInNoRecognizer
+{
+  CKComponentViewAttributeValue attr = CKComponentGestureAttribute(Nil, nullptr, @selector(test));
```
zats: Why to allow to create `CKComponentGestureAttribute` that won't do much anyway?

adamjernst: @zats I agree, but the alternative is crashing/failing an assert — and I think graceful failure is better here. A legit case might look like:

```
CKComponentGestureAttribute(enabled ? [UITapGestureRecognizer class] : Nil, ...)
```

Of course, you could also just pass a nil action in that case… but meh.

adamjernst: My original diff was just adding a CKAssert, then I thought better of it and decided to just gracefully ignore it.

zats: I guess 🤷‍♂️

adamjernst: Fair. Let's stop the bleeding — if you put out a PR to change it to a `CKAssert` I'd stamp that.

adamjernst: "stop the bleeding first" meaning stop the crashes that this causes :)

zats: agree, this is more important, let's land this one as is

PR #839, opened 2017-06-26
Introduce the 'CKComponentScopeEnumeratorProvider' protocol.

kfirapps:
- 'CKComponentScopeRoot' will conform to this protocol.
- The protocol extracts the enumration over predicates methods from 'CKComponentScopeRoot'.
- This way we can leverage these methods without exposing the internal implementation of 'CKComponentScopeRoot'.

### DIFF
--- a/ComponentKit/Core/Scope/CKComponentScopeTypes.h
+++ b/ComponentKit/Core/Scope/CKComponentScopeTypes.h
@@ -16,3 +16,36 @@ typedef int32_t CKComponentScopeHandleIdentifier;
 typedef int32_t CKComponentScopeRootIdentifier;
 
 typedef std::unordered_multimap<CKComponentScopeHandleIdentifier, id (^)(id)> CKComponentStateUpdateMap;
+
+@protocol CKScopedComponent;
+@protocol CKScopedComponentController;
+
+/**
+ Enumerator blocks allow a consumer to enumerate over all of the components or controllers that matched a predicate.
+ */
+typedef void (^CKComponentScopeEnumerator)(id<CKScopedComponent>);
+typedef void (^CKComponentControllerScopeEnumerator)(id<CKScopedComponentController>);
+
+/**
+ Scope predicates are a tool used by the framework to register components and controllers on initialization that have
+ specific characteristics. These predicates allow rapid enumeration over matching components and controllers.
+ */
+typedef BOOL (*CKComponentScopePredicate)(id<CKScopedComponent>);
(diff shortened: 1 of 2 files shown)


ComponentKit/Core/Scope/CKComponentScopeTypes.h
```diff
@@ -16,3 +16,36 @@ typedef int32_t CKComponentScopeHandleIdentifier;
...
+
+/**
+ Scope predicates are a tool used by the framework to register components and controllers on initialization that have
+ specific characteristics. These predicates allow rapid enumeration over matching components and controllers.
+ */
+typedef BOOL (*CKComponentScopePredicate)(id<CKScopedComponent>);
+typedef BOOL (*CKComponentControllerScopePredicate)(id<CKScopedComponentController>);
+
+
+@protocol CKComponentScopeEnumeratorProvider <NSObject>
```
dshahidehpour: Could we decouple this even further? The fact that we are enumerating over the Scope Root is totally in the implementation detail. Could we rename this to something like: `CKComponentPredicateMatcher`?

dshahidehpour:
Could we decouple this even further? The fact that we are enumerating over the Scope Root is totally in the implementation detail. Could we rename this to something like: `CKPredicateEnumerationProvider`?

kfirapps:
I agree with you that it would be even better to completely decouple it.

However, all this predicate concept is part of the scope root.
The predicate names, the predicate enumerators, the register methods in the scope root itself.

I'm not sure if changing here the protocol name only will make the difference.

dshahidehpour:
> The predicate names, the predicate enumerators, the register methods in the scope root itself.

Good point!

dshahidehpour:
@shipit

dshahidehpour:
@facebook-github-bot shipit

ComponentKit/Core/Scope/CKComponentScopeTypes.h
```diff
@@ -16,3 +16,36 @@ typedef int32_t CKComponentScopeHandleIdentifier;
...
 
 typedef std::unordered_multimap<CKComponentScopeHandleIdentifier, id (^)(id)> CKComponentStateUpdateMap;
+
+@protocol CKScopedComponent;
+@protocol CKScopedComponentController;
+
+/**
+ Enumerator blocks allow a consumer to enumerate over all of the components or controllers that matched a predicate.
+ */
+typedef void (^CKComponentScopeEnumerator)(id<CKScopedComponent>);
```
kfirapps: I agree with you both about the renaming and having a separate file - I'll open a task and get to this later.

PR #840, opened 2017-06-26
Appledocs copied a second time

ohliveyeah:


ryanrhee:
What is this? All I can see in the changes is that the date has been changed.

ryanrhee:
@dshahidehpour is gonna take it from here

PR #842, opened 2017-07-07
Extend the use of the scope root predicates

kfirapps:
Adds ability to leverage the scope root predicates from outside the framework.


ComponentKit/TransactionalDataSources/Common/CKTransactionalComponentDataSourceConfiguration.mm
```diff
@@ -17,6 +17,8 @@
 @implementation CKTransactionalComponentDataSourceConfiguration
 {
   CKSizeRange _sizeRange;
+  std::unordered_set<CKComponentScopePredicate> _componentPredicates;
```
dshahidehpour: Just realized, these should be const too, right?

PR #844, opened 2017-07-11
Add ability to demote CKTypedComponentAction, use it in CKButtonComponent

danzimm:
Previously demoting an object would call `CKTypedComponentAction<>(const CKTypedComponentAction<T...>&)` and error if the supplied action was a block. This provides a safe way to demote via currying.

Test added to demonstrate that demoting works as expected.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.h
+++ b/ComponentKit/Utilities/CKComponentAction.h
@@ -16,11 +16,11 @@
 @class CKComponent;
 
 namespace detail {
-  template<class...> struct any_are_reference : std::false_type {};
-  template<class T> struct any_are_reference<T> : std::is_reference<T> {};
+  template<class...> struct any_references : std::false_type {};
+  template<class T> struct any_references<T> : std::is_reference<T> {};
   template<class T, class... Ts>
-  struct any_are_reference<T, Ts...>
-    : std::conditional<std::is_reference<T>::value, std::true_type, any_are_reference<Ts...>>::type {};
+  struct any_references<T, Ts...>
+    : std::conditional<std::is_reference<T>::value, std::true_type, any_references<Ts...>>::type {};
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -134,6 +134,16 @@ class CKTypedComponentAction : public CKTypedComponentActionBase {
   }
 
   /**
+   Allows demoting an action to a simpler action while supplying defaults for the values that won't be passed in.
+   */
+  template<typename... U>
+  static CKTypedComponentAction<T...> demotedFrom(CKTypedComponentAction<T..., U...> action, U... defaults) {
+    return CKTypedComponentAction<T...>::actionFromBlock(^(CKComponent *sender, T... args) {
```
ocrickard: Alright, my one concern here is if the type is a reference, the capture will be garbage at execution time, which sucks. I think this is OK for now, but mind following up and seeing if you can add a static assert or perhaps augment this with a decayed type to force references to values?

ocrickard:
This looks great to me. I have one concern about references captured here if they're the type of the action. So for instance, demoting this action:

```
CKTypedComponentAction<const std::vector<BOOL> &> action;
```

Now when I demote, the type remains a reference, which is then captured by the block. This then points at garbage memory at the time of execution.

Thus, we need to add either a static assert OR properly handle this by decaying the type (I think?)

ComponentKit/Utilities/CKComponentAction.h
```diff
@@ -134,6 +142,16 @@ class CKTypedComponentAction : public CKTypedComponentActionBase {
   }
 
   /**
+   Allows demoting an action to a simpler action while supplying defaults for the values that won't be passed in.
+   */
+  template<typename... U, bool separator = false, class = typename std::enable_if<!detail::any_are_reference<U...>::value>::type>
```
ocrickard: I'm pretty sure this can be simplified a bit. Also, although enable_if is the right tool generally, I'm in favor of static asserts instead for our obj-c++ codebase since a lot of people aren't C++ gurus here, and with static assert you can provide a really nice compiler error telling them explicitly what to do.

ocrickard:
This looks great! You can merge from the internal dashboard whenever. I do think we could provide a nicer compiler error, but that can be dealt with later.

ComponentKit/Utilities/CKComponentAction.h
danzimm: Good call - using the static_assert simplifies things and provides a better error message. Thanks for pointing this out!

PR #845, opened 2017-07-12
Convert invocations to IMPs

ocrickard:
NSInvocation chokes on certain types of complex C++ references. I don't know exactly why yet, since it works for certain types, but not others. The memory in the receiving callsite gets garbage with some of our most complex C++ classes internally.

Converting to directly use an IMP has some major problems though: NSInvocation nils out any parameters you don't provide. We can no longer support that. Thus, we HAVE to move to a formal demotion mechanism instead of implicitly handling it for you.

The failure mode for this is awful, it crashes at runtime. Ugh. I plan to add some more checking code in the constructor of the component to validate that the selector looks right. We'll also have to do a huge search/codemod internally to make this happen. I'm not looking forward to this...

### DIFF
--- a/ComponentKit/Components/CKButtonComponent.mm
+++ b/ComponentKit/Components/CKButtonComponent.mm
@@ -159,13 +159,18 @@ typedef std::array<CKStateConfiguration, 8> CKStateConfigurationArray;
     contentEdgeInsets = [it->second UIEdgeInsetsValue];
   }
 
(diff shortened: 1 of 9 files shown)


ocrickard:
@facebook-github-bot import

ocrickard:
I'm gonna land this through an internal phabricator revision so I can stack the changes necessary for this to happen. Plus, we no longer have write access to the repo, so this has to be done internally first.

PR #846, opened 2017-07-13
Update 25-avoid-single-use-constants.md

alex904:


adamjernst:
Thanks for the feedback! Fair enough, it's a subjective call.

PR #847, opened 2017-07-26
Fix bug with block-based CKComponentActions

adamjernst:
The invariant is that if two CKComponentActions are not equal (according to `operator==`) they must have distinct identifiers. This was not true for block-based actions. Make it so.


ocrickard:
Sweet, thanks for fixing!

adamjernst:
@facebook-github-bot import

PR #848, opened 2017-07-26
Refactor CKComponentActionAttribute

adamjernst:
When this function was originally designed, it was a safe assumption that the number of unique `CKComponentAction` identifiers in the app would be at most ~200. Thus it seemed reasonable to allocate a separate "forwarder" for each identifier and keep the corresponding object in the map forever.

With the introduction of scope-based and block-based actions, this assumption is dangerous; there could potentially be tens of thousands of distinct identifiers, and each distinct action will result in the allocation of a separate forwarder that is essentially leaked forever.

Refactor by keying forwarders on `UIControlEvents` instead of `CKComponentAction::identifier`. This should result in at most ~16 forwarders being active at any given time, regardless of the number of actions.

### DIFF
--- a/ComponentKit/Utilities/CKComponentAction.mm
+++ b/ComponentKit/Utilities/CKComponentAction.mm
@@ -13,6 +13,7 @@
 #import <unordered_map>
 #import <vector>
 #import <array>
+#import <objc/runtime.h>
 
 #import "CKAssert.h"
 #import "CKComponent+UIView.h"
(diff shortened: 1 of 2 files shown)


ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -209,42 +210,34 @@ CKComponentViewAttributeValue CKComponentActionAttribute(const CKTypedComponentA
...
+          const auto it = map->find(controlEvents);
+          CKComponentActionControlForwarder *const forwarder =
+          (it == map->end())
+          ? map->insert({controlEvents, [[CKComponentActionControlForwarder alloc] initWithControlEvents:controlEvents]}).first->second
+          : it->second;
+          [control addTarget:forwarder
+                      action:@selector(handleControlEventFromSender:withEvent:)
+            forControlEvents:controlEvents];
+        }
+        list->_actions[controlEvents].push_back(action);
```
adamjernst: Now we store the list of actions that a UIControl should fire on an associated object on the control itself.

This neatly ensures that if the unapplicator never runs (it often won't!) we clean up the list of actions.

ocrickard: Ever runs*

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -209,42 +210,34 @@ CKComponentViewAttributeValue CKComponentActionAttribute(const CKTypedComponentA
...
       },
       ^(UIControl *control, id value){
-        [control removeTarget:forwarder
-                       action:@selector(handleControlEventFromSender:withEvent:)
-             forControlEvents:controlEvents];
+        CKComponentActionList *const list = objc_getAssociatedObject(control, ck_actionListKey);
+        CKCAssertNotNil(list, @"Unapplicator should always find an action list installed by applicator");
+        auto &actionList = list->_actions[controlEvents];
+        actionList.erase(std::find(actionList.begin(), actionList.end(), action));
```
ocrickard: Nice, linear search should be fine here.

ComponentKit/Utilities/CKComponentAction.mm
```diff
@@ -254,23 +247,32 @@ CKComponentViewAttributeValue CKComponentActionAttribute(const CKTypedComponentA
...
   return self;
 }
 
 - (void)handleControlEventFromSender:(UIControl *)sender withEvent:(UIEvent *)event
 {
-  // If the action can be handled by the sender itself, send it there instead of looking up the chain.
-  _action.send(sender.ck_component, CKComponentActionSendBehaviorStartAtSender, event);
+  CKComponentActionList *const list = objc_getAssociatedObject(sender, ck_actionListKey);
+  CKCAssertNotNil(list, @"Forwarder should always find an action list installed by applicator");
+  // Protect against mutation-during-enumeration by copying the list of actions to send:
```
ocrickard: Oh interesting. When might this happen?

ComponentKit/Utilities/CKComponentAction.mm
adamjernst: Actually I did mean "if it never runs." There is no guarantee that an unapplicator will run.

ocrickard:
Yay this looks fantastic!

ComponentKit/Utilities/CKComponentAction.mm
adamjernst: Frankly I can't imagine a case; you'd have to trigger a remount *during* the action callback. But better safe than sorry.

ComponentKit/Utilities/CKComponentAction.mm
ocrickard: lol fair enough.

ComponentKit/Utilities/CKComponentAction.mm
ocrickard: Yeah I think it's fine, I was just trying to imagine a case where this could happen. The only thing I could think of is someone calling cellFor... on the collection view, which I think could theoretically happen.

PR #853, opened 2017-08-15
Add assert for crashy argument count in preparation for IMP migration

toulouse:
Summary:
When we migrate to IMPs to support complex C++ types, we'll lose NSInvocation's argument nilling. The migration is necessary because with those complex C++ types, the compiler may have trouble understanding the proper way to handle arguments, and will result in crashy or undefined behavior if the wrong target is selected. Previously, we would just reject those complex argument types and simply lean on NSInvocation to handle the automatic nilling for those arguments we did not set, but this approach doesn't extend to IMPs.

The first step in addressing this is to assert when previously undefined-but-not-crashy behavior is detected. (This revision)

The second step will probably be to get logs and farm out the work to affected teams to fix their actions.

Next, we'll want to perform the migration to IMPs, and turn argument mismatches into asserting no-ops, after teams have had the opportunity to fix their low-hanging fruit, and gather logging on actions which are broken by this migration.

Lastly, we will want to evaluate the utility of performing argument typechecking with the NSMethodSignature for extra safety. Given the infrequent firing of actions the expected performance impact of this is likely negligible.

Reviewed By: ocrickard

Differential Revision: D5625064

### DIFF
--- a/ComponentKit/Utilities/CKComponentActionInternal.h
+++ b/ComponentKit/Utilities/CKComponentActionInternal.h
@@ -171,6 +171,8 @@ NSInvocation *CKComponentActionSendResponderInvocationPrepare(SEL selector, id t
 template<typename... T>
 static void CKComponentActionSendResponderChain(SEL selector, id target, CKComponent *sender, T... args) {
   NSInvocation *invocation = CKComponentActionSendResponderInvocationPrepare(selector, target, sender);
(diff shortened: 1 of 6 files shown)


toulouse:
fixing tests by borrowing approach from 845

toulouse:
@facebook-github-bot import

PR #855, opened 2017-08-19
Typo: becuase -> because

dbworku:
Why not fix it.


dshahidehpour:
@dbworku If you open up a PR internally, it would automatically merge here :)

ocrickard:
Yeah, the new workflow is waaay better. Also, I apparently no longer have write access to the repo at all, so couldn't merge this here even if I wanted to.

ocrickard:
Mind submitting on Phabricator instead?

PR #880, opened 2017-10-09
Update 01-getting-started.md

visserrgm:
Add carthage step to install readme


_docs/01-getting-started.md
```diff
@@ -12,6 +12,7 @@ Clone the Github repo, and then open the Xcode project.
 $ git clone https://github.com/facebook/componentkit
 $ cd componentkit/Examples/WildeGuess/
 $ open WildeGuess.xcodeproj
+$ carthage install
```
gkassabli: Thanks for adding this!
Let's move it to the first step and alter comment at line 9 to reflect this as well

gkassabli:
Looks great! Thanks @visserrgm for this fix!

PR #887, opened 2017-10-19
Fix for ComponentKit.podspec and command line details to push 0.20 to trunk

anton-matosov:
Remove test frameworks from the podspec
Test frameworks should never be linked to production code

There is a workaround since CocoaPods 1.2.1 that allows pushing of C++ pods again. The command line and fixes to podspec give:
```
Analyzed 1 podspec.

ComponentKit.podspec passed validation.
```

Here is the full command line
`pod spec lint ComponentKit.podspec --allow-warnings --skip-import-validation`

I am looking forward to see ComponentKit supporting CocoaPods again.


anton-matosov:
You are welcome

PR #891, opened 2017-11-09
Update readme.md to run build.sh

vividcode:
Update readme.md to run build.sh - without this, source files do not get downloaded and added to framework.


gkassabli:
Build.sh is designed to be used with CI, not initialize repo. We need to fix carthage bootstrap to be used instead

gkassabli:
Corrected the readme with correct call for carthage. Thanks for adding this anyway!

PR #898, opened 2018-02-02
RFC: Render Method in ComponentKit

kfirapps:
* *Component Tree Creation*
    * After the root component's creation in `CKBuildComponent(..)`, the *infra* creates a component tree (`CKComponentTreeNode`) recursively with the new *render* method. `CKComponentTreeNode` is the equivalent of `CKComponentScopeFrame`; but the main difference is that every component has one.

* *State*
    * Instead of accessing the state from a scope, the infra will pass it down to the render method.
    * No need for scopes anymore
* *Changes in `CKComponent` classes*
    * `CKSingleChildComponent`
        * Introduce a new a component type that has a render method: 
      `- (CKComponent *)render:(id)state`.
            The default implementation returns {}.
    * `CKGroupComponent`
        * Introduce a new a component type that has a renderGroup method: 
        `- (std::vector<CKComponent *>)renderGroup:(id)state`.
        The default implementation returns {}.
    * `CKFlexboxComponent`
        * Subclass from CKGroupComponent
        * Override the render method and map from `{CKFlexboxChild} → {CKComponent}`

### DIFF
--- a/ComponentKit/Core/CKBuildComponent.mm
+++ b/ComponentKit/Core/CKBuildComponent.mm
@@ -17,6 +17,7 @@
 #import "CKComponentScopeRoot.h"
 #import "CKComponentSubclass.h"
 #import "CKThreadLocalComponentScope.h"
+#import "CKTreeNode.h"
 
 CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
                                         const CKComponentStateUpdateMap &stateUpdates,
@@ -28,6 +29,12 @@ CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
   CKThreadLocalComponentScope threadScope(previousRoot, stateUpdates);
   // Order of operations matters, so first store into locals and then return a struct.
   CKComponent *const component = componentFactory();
+  // Build the component tree from the render function.
+  [component buildComponentTree:threadScope.newScopeRoot.rootNode
+                  previousOwner:previousRoot.rootNode
+                      scopeRoot:threadScope.newScopeRoot
+                   stateUpdates:stateUpdates];
+
   CKComponentScopeRoot *newScopeRoot = threadScope.newScopeRoot;
   [analyticsListener didBuildComponentTreeWithScopeRoot:newScopeRoot component:component];
   return {
--- a/ComponentKit/Core/CKComponent.mm
+++ b/ComponentKit/Core/CKComponent.mm
@@ -36,6 +36,7 @@
 #import "ComponentLayoutContext.h"
 #import "CKThreadLocalComponentScope.h"
 #import "CKComponentScopeRoot.h"
+#import "CKBaseTreeNode.h"
 
 CGFloat const kCKComponentParentDimensionUndefined = NAN;
 CGSize const kCKComponentParentSizeUndefined = {kCKComponentParentDimensionUndefined, kCKComponentParentDimensionUndefined};
@@ -72,6 +73,11 @@ struct CKComponentMountInfo {
   return [[self alloc] initWithView:view size:size];
 }
 
++ (instancetype)newWithViewWithoutScopeHandle:(const CKComponentViewConfiguration &)view size:(const CKComponentSize &)size
+{
+  return [[self alloc] initWithViewWithoutScopeHandle:view size:size];
+}
+
 + (instancetype)new
 {
   return [self newWithView:{} size:{}];
@@ -93,10 +99,27 @@ struct CKComponentMountInfo {
   return self;
 }
 
+- (instancetype)initWithViewWithoutScopeHandle:(const CKComponentViewConfiguration &)view
+                                          size:(const CKComponentSize &)size
+{
+  if (self = [super init]) {
+    _viewConfiguration = view;
+    _size = size;
+  }
+  return self;
+}
+
 - (void)dealloc
 {
   // Since the component and its view hold strong references to each other, this should never happen!
   CKAssert(_mountInfo == nullptr, @"%@ must be unmounted before dealloc", [self class]);
+}
+
+- (void)aquireScopeHandle:(CKComponentScopeHandle *)scopeHandle
+{
+  CKAssert(_scopeHandle == nil, @"Component(%@) already has '_scopeHandle'.", self);
+  [scopeHandle aquireFromComponentAssertIfWrong:self];
+  _scopeHandle = scopeHandle;
 }
 
 - (const CKComponentViewConfiguration &)viewConfiguration
@@ -108,6 +131,22 @@ struct CKComponentMountInfo {
 {
   CKAssertMainThread();
   return _mountInfo ? _mountInfo->viewContext : CKComponentViewContext();
+}
+
+#pragma mark - ComponentTree
+
+- (void)buildComponentTree:(CKTreeNode *)owner
+             previousOwner:(CKTreeNode *)previousOwner
+                 scopeRoot:(CKComponentScopeRoot *)scopeRoot
+              stateUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+{
+  // In this case this is a leaf component, which means we don't need to continue the recursion as it has no children.
+  __unused auto const node = [[CKBaseTreeNode alloc]
+                              initWithComponent:self
+                              owner:owner
+                              previousOwner:previousOwner
+                              scopeRoot:scopeRoot
+                              stateUpdates:stateUpdates];
 }
 
 #pragma mark - Mounting and Unmounting
--- a/ComponentKit/Core/CKCompositeComponent.mm
+++ b/ComponentKit/Core/CKCompositeComponent.mm
@@ -16,6 +16,7 @@
 #import "CKInternalHelpers.h"
 #import "CKComponentInternal.h"
 #import "CKComponentSubclass.h"
+#import "CKBaseTreeNode.h"
 
 @interface CKCompositeComponent ()
 {
@@ -62,6 +63,26 @@
 + (instancetype)newWithView:(const CKComponentViewConfiguration &)view size:(const CKComponentSize &)size
 {
   CK_NOT_DESIGNATED_INITIALIZER();
+}
+
+- (void)buildComponentTree:(CKTreeNode *)owner
+             previousOwner:(CKTreeNode *)previousOwner
+                 scopeRoot:(CKComponentScopeRoot *)scopeRoot
+              stateUpdates:(const CKComponentStateUpdateMap &)stateUpdates
+{
+  __unused auto const node = [[CKBaseTreeNode alloc]
+                              initWithComponent:self
+                              owner:owner
+                              previousOwner:previousOwner
+                              scopeRoot:scopeRoot
+                              stateUpdates:stateUpdates];
+
+  if (_component) {
+    [_component buildComponentTree:owner
+                     previousOwner:previousOwner // We pass here the previous parent, as we would CKCompositeComponent it's not an owner component.
+                         scopeRoot:scopeRoot
+                      stateUpdates:stateUpdates];
+  }
 }
 
 - (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
--- a/ComponentKit/Core/CKGroupComponent.h
+++ b/ComponentKit/Core/CKGroupComponent.h
@@ -1,0 +1,27 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <ComponentKit/CKComponent.h>
+#import <ComponentKit/CKComponentLayout.h>
+#import <ComponentKit/CKComponentOwner.h>
+
+@interface CKGroupComponent : CKComponent <CKComponentOwner>
+
+/*
+ Returns a vector of 'CKComponent' children that will be render to the screen.
+
+ If you override this method, you must override the `computeLatoutThatFits:children:` and provide a layout for these components.
+ If you don't need a custom layout, you can just use CKFlexboxComponent instead.
+
+ @param state The current state of the component.
+ */
+- (std::vector<CKComponent *>)renderGroup:(id)state;
+
+@end
--- a/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
+++ b/ComponentKit/Core/Scope/CKComponentScopeHandle.mm
@@ -47,9 +47,6 @@
     [currentScope->newScopeRoot registerComponent:component];
     return handle;
   }
-  CKCAssertNil([component.class controllerClass], @"%@ has a controller but no scope! "
-               "Make sure you construct your scope(self) before constructing the component or CKComponentTestRootScope "
-               "at the start of the test.", [component class]);
 
   return nil;
 }
@@ -57,7 +54,7 @@
 - (instancetype)initWithListener:(id<CKComponentStateListener>)listener
                   rootIdentifier:(CKComponentScopeRootIdentifier)rootIdentifier
                   componentClass:(Class)componentClass
-             initialStateCreator:(id (^)(void))initialStateCreator
+                    initialState:(id)initialState
                           parent:(CKComponentScopeHandle *)parent
 {
   static int32_t nextGlobalIdentifier = 0;
@@ -65,7 +62,7 @@
                globalIdentifier:OSAtomicIncrement32(&nextGlobalIdentifier)
                  rootIdentifier:rootIdentifier
                  componentClass:componentClass
-                          state:initialStateCreator ? initialStateCreator() : [componentClass initialState]
+                          state:initialState
                      controller:nil  // Controllers are built on resolution of the handle.
                 scopedResponder:nil  // Scoped responders are created lazily. Once they exist, we use that reference for future handles.
                          parent:parent];
@@ -183,6 +180,15 @@
     return NO;
   }
 }
+
+- (void)aquireFromComponentAssertIfWrong:(id<CKComponentProtocol>)component
+{
+  CKAssert([component isMemberOfClass:_componentClass], @"%@ has to be a member of %@ class", component, _componentClass);
+  CKAssert(!_acquired, @"scope handle cannot be aquired twice");
+  _acquired = YES;
+  _acquiredComponent = component;
+}
+
 
 - (void)resolve
 {
--- a/ComponentKit/LayoutComponents/CKBackgroundLayoutComponent.mm
+++ b/ComponentKit/LayoutComponents/CKBackgroundLayoutComponent.mm
@@ -40,6 +40,11 @@
   CK_NOT_DESIGNATED_INITIALIZER();
 }
 
+- (std::vector<CKComponent *>)renderGroup:(id)state
+{
+  return {_component, _background};
+}
+
 /**
  First layout the contents, then fit the background image.
  */
--- a/ComponentKit/LayoutComponents/CKFlexboxComponent.mm
+++ b/ComponentKit/LayoutComponents/CKFlexboxComponent.mm
@@ -17,6 +17,8 @@
 #import "CKInternalHelpers.h"
 #import "CKComponentLayout.h"
 #import "CKComponentLayoutBaseline.h"
+#import "CKThreadLocalComponentScope.h"
+#import "ComponentUtilities.h"
 
 const struct CKStackComponentLayoutExtraKeys CKStackComponentLayoutExtraKeys = {
   .hadOverflow = @"hadOverflow"
@@ -77,6 +79,13 @@ template class std::vector<CKFlexboxComponentChild>;
     component->_reuseOnlyExactSizeSpecs = CKComponentContext<CKFlexboxComponentContext>::get().reuseOnlyExactSizeSpecs;
   }
   return component;
+}
+
+- (std::vector<CKComponent *>)renderGroup:(id)state
+{
+  return  CK::map(_children, [](CKFlexboxComponentChild child) {
+    return child.component;
+  });
 }
 
 static YGConfigRef ckYogaDefaultConfig()
--- a/Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
+++ b/Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
@@ -22,49 +22,53 @@ static NSString *const oscarWilde = @"Oscar Wilde";
 
 @implementation InteractiveQuoteComponent
 {
-  CKComponent *_overlay;
+  Quote *_quote;
+  QuoteContext *_context;
 }
 
 + (instancetype)newWithQuote:(Quote *)quote
                      context:(QuoteContext *)context
 {
-  CKComponentScope scope(self);
-  const BOOL revealAnswer = [scope.state() boolValue];
+  InteractiveQuoteComponent *c = [super newWithView:{} size:{}];
+  if (c) {
+    c->_quote = quote;
+    c->_context = context;
+  }
+  return c;
+}
+
+- (CKComponent *)render:(id)state
+{
+  auto const revealAnswer = [state boolValue];
 
   CKComponent *overlay =
   revealAnswer
   ? [SuccessIndicatorComponent
-     newWithIndicatesSuccess:[quote.author isEqualToString:oscarWilde]
+     newWithIndicatesSuccess:[_quote.author isEqualToString:oscarWilde]
      successText:[NSString stringWithFormat:@"This quote is by %@", oscarWilde]
      failureText:[NSString stringWithFormat:@"This quote isn't by %@", oscarWilde]]
   : nil;
 
-  InteractiveQuoteComponent *c =
-  [super newWithComponent:
-   [CKFlexboxComponent
-    newWithView:{
-      [UIView class],
-      {CKComponentTapGestureAttribute(@selector(didTap))}
-    }
-    size:{}
-    style:{
-      .alignItems = CKFlexboxAlignItemsStretch
-    }
-    children:{
-      {[CKOverlayLayoutComponent
-        newWithComponent:[QuoteComponent newWithQuote:quote context:context]
-        overlay:overlay]},
-      {[CKComponent
-        newWithView:{
-          [UIView class],
-          {{@selector(setBackgroundColor:), [UIColor lightGrayColor]}}
-        }
-        size:{.height = 1/[UIScreen mainScreen].scale}]}
-    }]];
-  if (c) {
-    c->_overlay = overlay;
-  }
-  return c;
+  return [CKFlexboxComponent
+          newWithView:{
+            [UIView class],
+            {CKComponentTapGestureAttribute(@selector(didTap))}
+          }
+          size:{}
+          style:{
+            .alignItems = CKFlexboxAlignItemsStretch
+          }
+          children:{
+            {[CKOverlayLayoutComponent
+              newWithComponent:[QuoteComponent newWithQuote:_quote context:_context]
+              overlay:overlay]},
+            {[CKComponent
+              newWithView:{
+                [UIView class],
+                {{@selector(setBackgroundColor:), [UIColor lightGrayColor]}}
+              }
+              size:{.height = 1/[UIScreen mainScreen].scale}]}
+          }];
 }
 
 + (id)initialState
@@ -77,24 +81,6 @@ static NSString *const oscarWilde = @"Oscar Wilde";
   [self updateState:^(NSNumber *oldState){
     return [oldState boolValue] ? @NO : @YES;
   } mode:CKUpdateModeSynchronous];
-}
-
-- (std::vector<CKComponentAnimation>)animationsFromPreviousComponent:(InteractiveQuoteComponent *)previousComponent
-{
-  if (previousComponent->_overlay == nil && _overlay != nil) {
-    return {{_overlay, scaleToAppear()}}; // Scale the overlay in when it appears.
-  } else {
-    return {};
-  }
-}
-
-static CAAnimation *scaleToAppear()
-{
-  CABasicAnimation *scale = [CABasicAnimation animationWithKeyPath:@"transform"];
-  scale.fromValue = [NSValue valueWithCATransform3D:CATransform3DMakeScale(0.0, 0.0, 0.0)];
-  scale.timingFunction = [CAMediaTimingFunction functionWithName:kCAMediaTimingFunctionEaseOut];
-  scale.duration = 0.2;
-  return scale;
 }
 
 @end
--- a/Examples/WildeGuess/WildeGuess/Components/MonochromeQuoteComponent.mm
+++ b/Examples/WildeGuess/WildeGuess/Components/MonochromeQuoteComponent.mm
@@ -15,8 +15,22 @@
 #import "QuoteContext.h"
 
 @implementation MonochromeQuoteComponent
+{
+  NSString *_text;
+  QuoteContext *_context;
+}
 
 + (instancetype)newWithText:(NSString *)text context:(QuoteContext *)context
+{
+  auto const c = [super new];
+  if (c) {
+    c->_text = text;
+    c->_context = context;
+  }
+  return c;
+}
+
+- (CKComponent *)render:(id)state
 {
   CKComponent *quoteTextComponent =
   [CKInsetComponent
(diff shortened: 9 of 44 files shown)


ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -17,6 +17,44 @@
...
+                               scopeRoot:scopeRoot
+                               component:c
+                               stateUpdates:stateUpdates];
+
+  auto const children = [c render:node.state];
+
+  // In case of that the render method returns one child, we will set it on the parent component in order to take it into account during layout time.
+  // A better solution would be to unify the render callbacks with the layout process and have a direct access to the component children.
+  if (children.size() == 1) {
+    c.childComponent = children[0];
```
adamjernst: What happens if `CKBuildComponentTree` is called simultaneously on two different threads? For example, this happens quite frequently when rendering is performed on the main thread and a background thread in parallel.

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -28,6 +66,13 @@ CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
   CKThreadLocalComponentScope threadScope(previousRoot, stateUpdates);
   // Order of operations matters, so first store into locals and then return a struct.
   CKComponent *const component = componentFactory();
+  // Build the component tree from the render function.
+  CKBuildComponentTree(component,
+                       threadScope.newScopeRoot.rootNode,
+                       previousRoot.rootNode,
+                       threadScope.newScopeRoot,
+                       stateUpdates);
```
adamjernst: I see — it seems that you only call `CKBuildComponentTree` exactly once, immediately after render, before anyone else has a reference to the given component.

One of the principles I think it's worth striving for: data structures should be completely immutable after their creation. I am not sure if that will be possible in this particular case, but I think it would be useful to try.

ComponentKit/Core/CKComponent.h
```diff
@@ -55,4 +57,9 @@ struct CKComponentViewContext {
  */
 - (id)nextResponder;
 
+/**
+ This method composes an infra-driven component tree.
```
adamjernst: As a product developer, I don't know what this means.

ComponentKit/Core/CKComponent.h
```diff
@@ -55,4 +57,9 @@ struct CKComponentViewContext {
...
+ */
+- (std::vector<CKComponent *>)render:(id)state;
```
adamjernst: Does this need to be on `CKComponent`? Could it instead be part of a separate subclass, à la `CKCompositeComponent`?

ComponentKit/Core/CKComponent.mm
```diff
@@ -252,6 +265,11 @@ - (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
                           restrictedToSize:(const CKComponentSize &)size
                       relativeToParentSize:(CGSize)parentSize
 {
+  if (_childComponent) {
+    CKComponentLayout firstChild = [_childComponent layoutThatFits:constrainedSize parentSize:parentSize];
+    return {self, firstChild.size, {{{0,0}, firstChild}}};
+  }
```
adamjernst: This is why I'd prefer to keep it in a separate subclass — the cruft here seems unfortunate. Now layoutThatFits has two meanings, depending on whether there is a `childComponent` or not.

ComponentKit/Core/CKCompositeComponent.mm
```diff
@@ -64,6 +67,11 @@ + (instancetype)newWithView:(const CKComponentViewConfiguration &)view size:(con
...
 }
 
+- (std::vector<CKComponent *>)render:(id)state
+{
+  return {_component};
+}
+
 - (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
                           restrictedToSize:(const CKComponentSize &)size
                       relativeToParentSize:(CGSize)parentSize
```
adamjernst: so this method is now dead code, right? why is it still here?

ComponentKit/Core/Scope/CKComponentTreeNode.h
```diff
@@ -0,0 +1,28 @@
+//
+//  CKComponentTreeNode.h
+//  ComponentKit
+//
+//  Created by Kfir Schindelhaim on 1/23/18.
```
adamjernst: nit: needs fixing

ComponentKit/LayoutComponents/CKBackgroundLayoutComponent.mm
```diff
@@ -40,6 +40,11 @@ + (instancetype)newWithView:(const CKComponentViewConfiguration &)view size:(con
   CK_NOT_DESIGNATED_INITIALIZER();
 }
 
+- (std::vector<CKComponent *>)render:(id)state
+{
+  return {_component, _background};
```
adamjernst: how does the infrastructure know to apply the same logic that is currently in `computeLayoutThatFits`?

ComponentKit/LayoutComponents/CKFlexboxComponent.mm
```diff
@@ -56,13 +58,22 @@ + (instancetype)newWithView:(const CKComponentViewConfiguration &)view
...
     component->_children = children.take();
   }
+  
   return component;
 }
 
+- (std::vector<CKComponent *>)render:(id)state
+{
+  return  CK::map(_children, [](CKFlexboxComponentChild child) {
+    return child.component;
```
adamjernst: this loses all the flexbox props. who applies them?

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
```diff
@@ -23,24 +23,33 @@
...
-  const BOOL revealAnswer = [scope.state() boolValue];
-
   CKComponent *overlay =
-  revealAnswer
-  ? [SuccessIndicatorComponent
-     newWithIndicatesSuccess:[quote.author isEqualToString:oscarWilde]
-     successText:[NSString stringWithFormat:@"This quote is by %@", oscarWilde]
-     failureText:[NSString stringWithFormat:@"This quote isn't by %@", oscarWilde]]
-  : nil;
+  [SuccessIndicatorComponent
```
adamjernst: why is this component created in +new instead of render?

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
```diff
@@ -52,19 +61,16 @@ + (instancetype)newWithQuote:(Quote *)quote
     }
     children:{
       {[CKOverlayLayoutComponent
-        newWithComponent:[QuoteComponent newWithQuote:quote context:context]
-        overlay:overlay]},
+        newWithComponent:[QuoteComponent newWithQuote:_quote context:_context]
+        overlay:(revealAnswer ? _overlay : nil)]},
```
adamjernst: I think it's fairly problematic that we now **always** create `SuccessIndicatorComponent`, instead of only creating it when `revealAnswer` is `YES`.

Don't you think that's wasteful from a performance angle? Is there a solution?

Examples/WildeGuess/WildeGuess/Components/MonochromeQuoteComponent.mm
```diff
@@ -15,16 +15,30 @@
...
+}
 
 + (instancetype)newWithText:(NSString *)text context:(QuoteContext *)context
+{
+  auto const c = [super new];
+  if (c) {
+    c->_text = text;
+    c->_context = context;
+  }
+  return c;
```
adamjernst: This is a good example of the additional boilerplate required.

Examples/WildeGuess/WildeGuess/Components/QuoteComponent.mm
```diff
@@ -19,10 +19,26 @@
...
+    c->_context = context;
+  }
+  return c;
+}
+
+- (std::vector<CKComponent *>)render:(id)state
+{
+  return {
+    quoteComponent(_quote, _context)
+  };
```
adamjernst: I'll be honest, 16 additional lines of boilerplate makes me sweat a little :)

ComponentKit/LayoutComponents/CKFlexboxComponent.mm
adamjernst: Or does this just exist to ensure that the children are built by the infrastructure?

adamjernst: If so, what happens if I forget to return a child? That's a failure case we should consider.

ComponentKit/Core/CKBuildComponent.mm
kfirapps: `CKBuildComponentTree` is being called only once during the component creation, it cannot be call twice. However, you are right that if we are going to implement reconciliation and reuse a component tree node - it won't work. This was a first quick solution for the RFC.

An ideal solution would be to access the tree node from the layout computation. That means that the component itself can be attached to a two different nodes at any given time and it doesn't matter as it's immutable. The information about the children is being encapsulated in the tree node.

ComponentKit/Core/CKBuildComponent.mm
kfirapps: I agree.

ComponentKit/Core/CKComponent.h
kfirapps: The initializer of `CKCompositeComponent` takes a child component as a parameter. If we add the the render method only for `CKCompositeComponent`, how do we make sure that we create the child only in the render method and not in the constructor?

ComponentKit/Core/CKComponent.mm
kfirapps: I agree that this part is not ideal. It's on our to-do list to figure out a better solution.

ComponentKit/Core/CKCompositeComponent.mm
kfirapps: What about old components that didn't migrate to the render method? the idea is to allow mix-and-match between "old" and new components.

ComponentKit/Core/CKComponent.h
aCorrado: I think @adamjernst  was suggesting to have a CKRenderEnabledComponent subclass

ComponentKit/LayoutComponents/CKBackgroundLayoutComponent.mm
kfirapps: The layout process is the same as it was before. The render method only being called in order to construct the component tree (`CKComponentTreeNode`). In the future, we could unify the two.

ComponentKit/LayoutComponents/CKFlexboxComponent.mm
kfirapps: Same as the comment above. The render only creates the component tree, it's not responsible to compute the layout. Means that the `CKFlexboxComponent` works exactly as before. We compute the layout with `computeLayoutThatFits:` and this method hasn't been changed.

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
kfirapps: Just because it was saved as a prop before. But it actually makes the animation incorrect in this case as this component was created only on a specific state.

We should move it to the render method and alternate the animation method to receive the state in order to solve this issue.

Examples/WildeGuess/WildeGuess/Components/InteractiveQuoteComponent.mm
kfirapps: Yes, agree. We need to create it only when needed. See comment above.

ComponentKit/Core/CKComponent.h
kfirapps: What would be the benefit to have it on a separate subclass and not on `CKComponent`?

ComponentKit/LayoutComponents/CKFlexboxComponent.mm
kfirapps: > what happens if I forget to return a child?

I think it's similar to what we have today - if you create a component in the constructor but doesn't reflect it in your layout it won't be in the component layout, right?
People usually subclass from `CKCompositeComponent` which takes care of the layout for you. However, if you have more than one child you have to implement the `computeLayoutThatFits:` in order to reflect it in the layout.

ComponentKit/Core/CKComponent.h
kfirapps: I think I understand your point better now, we could have something like:
`CKComponent` would be a leaf component.
`CKChildComponent` would implement render that returns a child.
`CKFlexboxComponent` would be the "group" component, its render returns a vector of children.

In this case, we have to check in the infra the return type of the component from the render method and operate differently for each use case.

ComponentKit/Core/CKCompositeComponent.mm
ocrickard: Should we provide a new superclass for components that want to use the render method?

ComponentKit/Core/Scope/CKComponentTreeNode.h
```diff
@@ -0,0 +1,28 @@
...
+//
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKComponentScopeTypes.h>
+#import <ComponentKit/CKComponentScopeRoot.h>
+#import <ComponentKit/CKComponentScopeHandle.h>
+
+@interface CKComponentTreeNode : NSObject
```
ocrickard: docblock?

Examples/WildeGuess/WildeGuess/Components/MonochromeQuoteComponent.mm
ocrickard: Yep, I tried to fix all this with templates/code-gen before, but couldn't get it to be minimal.

ocrickard:
Very cool! I think it might be worth taking a pass to reduce the boilerplate if that's possible.

ComponentKit/Core/CKCompositeComponent.mm
kfirapps: We are going to try this idea. I'm still wondering - what is the main advantage of this separation?
If we implement the render in `CKCompositeComponent` we receive part of the conversion for "free".
What is the down side of this implementation?

ocrickard:
@eczarny may have some opinions as well. He was deeply involved in my proposals around typed props.

ComponentKit/Core/CKCompositeComponent.mm
```diff
@@ -64,6 +65,33 @@ + (instancetype)newWithView:(const CKComponentViewConfiguration &)view size:(con
   CK_NOT_DESIGNATED_INITIALIZER();
 }
 
+- (CKComponent *)render:(id)state
+{
+  return _component;
```
kfirapps: I've decided to implement the render method here in order to allow mix-and-match of the "old" components with the "new" ones. Otherwise, we might have a `CKCompositeComponent` in the component tree; in this case it will not continue to call `buildComponentTree` recursively.

ComponentKit/Core/CKGroupComponent.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ You have to override this method to computes your component's layout.
+
+ @param constrainedSize A min and max size. This is computed as described in the description. The CKComponentLayout you
+ return MUST have a size between these two sizes. This is enforced by assertion.
+ @param children A vector of children that the component provided in the 'renderGroup:' method.
+ */
+- (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
+                                  children:(std::vector<CKComponent *>)children;
```
kfirapps: @adamjernst, this is a first attempt to address the inconsistency between the components in the `renderGroup:` method and ones we have in the layout. We can assert and make sure that the `CKComponentLayout` contains the same components in its children vector. However, if we assert it it will break the current behaviour of ComponentKit.

kfirapps: We probably also need to implement:

```
- (CKComponentLayout)computeLayoutThatFits:(CKSizeRange)constrainedSize
                          restrictedToSize:(const CKComponentSize &)size
                      relativeToParentSize:(CGSize)parentSize
                                  children:(std::vector<CKComponent *>)children
```

ComponentKit/Core/CKCompositeComponent.mm
ocrickard: Fair point @kfirapps. I think you're seeing us be really scared of interface creep. We've traditionally been really wary of components with more than a single responsibility. Perhaps we should relax some of our constraints here in favor of a new path forward. Perhaps to ease the change, what's the plan for full conversion such that we remove this? Or is it permanent?

ComponentKit/Core/CKGroupComponent.h
```diff
@@ -0,0 +1,37 @@
...
+
+/*
+ Returns a vector of 'CKComponent' children that will be render to the screen.
+
+ If you overrides this method, you must override the `computeLatoutThatFits:children:` and provide a layout for these components.
+ If you don't need a custom layout, you can just use CKFlexboxComponent instead.
+
+ @param state The current state of the component.
+ */
+- (std::vector<CKComponent *>)renderGroup:(id)state;
```
ocrickard: Ohhhh, I see, you've created a vector here.

ComponentKit/Core/CKGroupComponent.h
```diff
@@ -0,0 +1,37 @@
...
+#import <ComponentKit/CKComponent.h>
+#import <ComponentKit/CKComponentLayout.h>
+#import <ComponentKit/CKComponentTreeOwner.h>
+
+@interface CKGroupComponent : CKComponent <CKComponentTreeOwner>
+
+/*
+ Returns a vector of 'CKComponent' children that will be render to the screen.
+
+ If you overrides this method, you must override the `computeLatoutThatFits:children:` and provide a layout for these components.
```
ocrickard: "If you override"

ComponentKit/Core/CKGroupComponent.h
```diff
@@ -0,0 +1,37 @@
...
+ If you don't need a custom layout, you can just use CKFlexboxComponent instead.
+
+ @param state The current state of the component.
+ */
+- (std::vector<CKComponent *>)renderGroup:(id)state;
+
+/**
+ You have to override this method to computes your component's layout.
```
ocrickard: "to  compute"

ComponentKit/Core/CKGroupComponent.h
```diff
@@ -0,0 +1,37 @@
...
+
+ @param constrainedSize A min and max size. This is computed as described in the description. The CKComponentLayout you
```
ocrickard: > This is computed as described in the description

Adds nothing to the comment. Remove?

ComponentKit/Core/ComponentTree/CKComponentGroupTreeNode.mm
```diff
@@ -0,0 +1,66 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKComponentGroupTreeNode.h"
+
+#import <ComponentKit/CKComponent.h>
+#import <ComponentKit/CKComponentInternal.h>
+#import "CKThreadLocalComponentScope.h"
```
ocrickard: order of imports

ComponentKit/Core/ComponentTree/CKComponentGroupTreeNode.mm
```diff
@@ -0,0 +1,66 @@
...
+#include <tuple>
+
+#import "CKMutex.h"
+
+struct CKComponentTreeNodeComparator {
+  bool operator() (const std::tuple<Class, NSUInteger> &lhs, const std::tuple<Class, NSUInteger> &rhs) const
+  {
+    return std::get<0>(lhs) == std::get<0>(rhs) && std::get<1>(lhs) == std::get<1>(rhs);
+  }
+};
```
ocrickard: No docs for this or the hasher?

ComponentKit/Core/ComponentTree/CKComponentGroupTreeNode.h
```diff
@@ -0,0 +1,19 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKComponentTreeNode.h"
+
+#include <vector>
+
+@interface CKComponentGroupTreeNode : CKComponentTreeNode
```
ocrickard: You need a lengthy docblock for this. Until I get a full explanation, I won't understand fully how all this fits together.

ComponentKit/Core/ComponentTree/CKComponentGroupTreeNode.mm
```diff
@@ -0,0 +1,66 @@
...
+  {
+    return [std::get<0>(n) hash] ^ std::get<1>(n);
+  }
+};
+
+typedef std::unordered_map<std::tuple<Class, NSUInteger>, CKComponentTreeNode *, CKComponentTreeNodeHasher, CKComponentTreeNodeComparator> CKScopeNodeMap;
+
+@implementation CKComponentGroupTreeNode
+{
+  CKScopeNodeMap _children;
```
ocrickard: A note on why the mutation of this map is safe is needed somewhere in this file.

ComponentKit/Core/ComponentTree/CKComponentTreeNode.h
```diff
@@ -0,0 +1,35 @@
...
+
+- (instancetype)initWithComponent:(CKComponent *)component
+                           parent:(CKComponentTreeNode *)parent
+                   previousParent:(CKComponentTreeNode *)previousParent
+                        scopeRoot:(CKComponentScopeRoot *)scopeRoot
+                     stateUpdates:(const CKComponentStateUpdateMap &)stateUpdates;
+
+@property (nonatomic, strong, readonly) CKComponent *component;
+@property (nonatomic, strong, readonly) CKComponentScopeHandle *handle;
+@property (nonatomic, readonly) id state;
```
ocrickard: nit: why strong above, and this one randomly is omitted?

ComponentKit/Core/ComponentTree/CKComponentTreeNode.h
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <ComponentKit/CKComponentScopeTypes.h>
+#import <ComponentKit/CKComponentScopeRoot.h>
+#import <ComponentKit/CKComponentScopeHandle.h>
+
+@interface CKComponentTreeNode : NSObject
```
ocrickard: Docblocks!

ComponentKit/Core/ComponentTree/CKComponentTreeNode.h
```diff
@@ -0,0 +1,35 @@
...
+                     stateUpdates:(const CKComponentStateUpdateMap &)stateUpdates;
+
+@property (nonatomic, strong, readonly) CKComponent *component;
+@property (nonatomic, strong, readonly) CKComponentScopeHandle *handle;
+@property (nonatomic, readonly) id state;
+
+- (const std::tuple<Class, NSUInteger> &)componentKey;
+- (CKComponentTreeNode *)childForComponentKey:(const std::tuple<Class, NSUInteger> &)key;
+
+- (std::tuple<Class, NSUInteger>)createComponentKeyForChildWithClass:(id<CKComponentProtocol>)componentClass;
```
ocrickard: This is really confusing. Excited to see if the docblocks make it less so.

ComponentKit/Core/ComponentTree/CKComponentTreeNode.mm
```diff
@@ -0,0 +1,105 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "CKComponentTreeNode.h"
+
+#import <ComponentKit/CKComponent.h>
+#import <ComponentKit/CKComponentInternal.h>
+#import "CKThreadLocalComponentScope.h"
```
ocrickard: again: import order

ComponentKit/Core/ComponentTree/CKComponentTreeNode.mm
```diff
@@ -0,0 +1,105 @@
...
+
+- (std::tuple<Class, NSUInteger>)createComponentKeyForChildWithClass:(id<CKComponentProtocol>)componentClass
+{
+  CKFatal(@"%@ shouldn't be called on a leaf component", NSStringFromSelector(_cmd));
+  return std::make_tuple(componentClass, 0);
+}
+
+- (void)setChild:(CKComponentTreeNode *)child forComponentKey:(const std::tuple<Class, NSUInteger> &)componentKey
+{
+  CKFatal(@"%@ shouldn't be called on a leaf component", NSStringFromSelector(_cmd));
```
ocrickard: I'm imagining when I could see this fatal... I don't know if I'd be able to understand what I should do instead....

ComponentKit/Core/ComponentTree/CKComponentTreeOwner.h
```diff
@@ -0,0 +1,34 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+/**
+ This protocol defines how a component behaves when the component tree is being construct by the method
+ 'buildComponentTree:previousParent:scopeRootstateUpdates'.
+
+ If its owner base, it will attache all the children from the render method to its own CKComponentTreeNode.
```
ocrickard: spelling/grammar.

ComponentKit/Core/ComponentTree/CKComponentTreeOwner.h
```diff
@@ -0,0 +1,34 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+/**
+ This protocol defines how a component behaves when the component tree is being construct by the method
```
ocrickard: This explanation doesn't really inform me of anything.

ComponentKit/Core/ComponentTree/CKComponentTreeOwner.h
```diff
@@ -0,0 +1,34 @@
...
+
+ For example:
+ CKChildComponent is an owner component.
+ CKFlexboxComponent isn't; it receives its children as props in its constructor.
+
+ Default values:
+ CKChildComponent returns YES
+ CKGroupComponent returns NO
+ */
+- (BOOL)ownerComponent;
```
ocrickard: Should anyone ever implement this? Or is this just for the base class?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -47,9 +47,10 @@ + (CKComponentScopeHandle *)handleForComponent:(id<CKComponentProtocol>)componen
     [currentScope->newScopeRoot registerComponent:component];
     return handle;
   }
-  CKCAssertNil([component.class controllerClass], @"%@ has a controller but no scope! "
-               "Make sure you construct your scope(self) before constructing the component or CKComponentTestRootScope "
-               "at the start of the test.", [component class]);
+
+//  CKCAssertNil([component.class controllerClass], @"%@ has a controller but no scope! "
+//               "Make sure you construct your scope(self) before constructing the component or CKComponentTestRootScope "
+//               "at the start of the test.", [component class]);
```
ocrickard: Probably just for development, but this needs to be cleaned up

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
```diff
@@ -184,6 +185,15 @@ - (BOOL)acquireFromComponent:(id<CKComponentProtocol>)component
   }
 }
 
+- (void)aquireFromComponentAssertIfWrong:(id<CKComponentProtocol>)component
```
ocrickard: Why is this different from above?

Examples/WildeGuess/WildeGuess/Components/QuoteComponent.mm
```diff
@@ -19,10 +19,24 @@
...
 + (instancetype)newWithQuote:(Quote *)quote context:(QuoteContext *)context
 {
-  return [super newWithComponent:quoteComponent(quote, context)];
+  auto const c = [super new];
+  if (c) {
+    c->_quote = quote;
+    c->_context = context;
+  }
+  return c;
+}
```
ocrickard: I really dislike this. Can we not figure out a way to type the props and provide the generic initializer?

ComponentKit/Core/CKBuildComponent.mm
```diff
@@ -28,6 +28,12 @@ CKBuildComponentResult CKBuildComponent(CKComponentScopeRoot *previousRoot,
   CKThreadLocalComponentScope threadScope(previousRoot, stateUpdates);
   // Order of operations matters, so first store into locals and then return a struct.
   CKComponent *const component = componentFactory();
+  // Build the component tree from the render function.
+  [component buildComponentTree:threadScope.newScopeRoot.rootNode
```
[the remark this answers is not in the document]
kfirapps: It's being done once only during the component tree creation.

ComponentKit/Core/CKChildComponent.h
```diff
@@ -0,0 +1,24 @@
...
+#import <ComponentKit/CKComponentTreeOwner.h>
+
+@interface CKChildComponent : CKComponent <CKComponentTreeOwner>
+
+/**
+ Returns a child component that needs to be rendered from this component.
+
+ @param state The current state of the component.
+ */
+- (CKComponent *)render:(id)state;
```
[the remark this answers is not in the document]
kfirapps: Yes that's the way Litho works. The first version did return a vector of children. However, changing to this approach allowed us to optimize the component tree creation time and memory. 
In most of the cases we actually return a single child anyway. In case we need to return more than one there are two options:
1. Use `CKFlexboxComponent` (Row or Column in Litho)
2. Inherit from `CKGrouopComponent` and override the `renderGroup` method.

In the 2nd case, we have to provide a custom layout.

ComponentKit/Core/CKComponent.mm
```diff
@@ -93,12 +101,29 @@ - (instancetype)initWithView:(const CKComponentViewConfiguration &)view
   return self;
 }
 
+- (instancetype)initWithoutScopeHandleWithView:(const CKComponentViewConfiguration &)view
```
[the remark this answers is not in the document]
kfirapps: Do you have a suggestion for a better name? this meant be used by the infra only.

ComponentKit/Core/CKComponent.mm
```diff
@@ -110,6 +135,32 @@ - (CKComponentViewContext)viewContext
   return _mountInfo ? _mountInfo->viewContext : CKComponentViewContext();
 }
 
+#pragma mark - Render
+
+- (CKComponent *)render:(id)state
```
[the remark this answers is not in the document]
kfirapps: Yes, we can take a look in to that.

ComponentKit/Core/CKCompositeComponent.mm
kfirapps: The equivalent component in the new API is `CKChildComponent`. That means that once we finish the migration we could get rid of this component.

ComponentKit/Core/CKGroupComponent.h
kfirapps: Thanks. Typo :)

ComponentKit/Core/ComponentTree/CKComponentTreeNode.mm
kfirapps: This in theory could never happen, unless we have a bug in the infrastructure. I guess it's more as a safety assert for us. How would you suggest to handle such a case instead?

ComponentKit/Core/Scope/CKComponentScopeHandle.mm
kfirapps: As we create the scope explicitly for a specific component, we don't need to check `isMemberOfClass` and `_acquired` - which could make it faster.

Examples/WildeGuess/WildeGuess/Components/QuoteComponent.mm
kfirapps: I had the impression that you tried a year ago and you couldn't find a good solution to proceed with. We don't like it either, but I wouldn't make it as a blocker for the render change.

ComponentKit/Core/CKComponent.mm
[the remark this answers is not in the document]
kfirapps: Obj-C generics don't really helps here. We use the state of the component only inside the component - which anyway is going to have type `id`.

kfirapps:
We will review this change internally and sync it to the open source later.

PR #902, opened 2018-02-23
[ComponentKit] Fix the failing test on CKComponentLayoutTests

kfirapps:
Using a block to verify that the component have been deallocated instead of `NSPointerArray`.


kfirapps:
Will land it internally.

PR #910, opened 2018-05-19
FIX: wrong code in `component controllers`

ObesityChow:
example code in this chapter is wrong. 

which `+ (Class<CKComponentControllerProtocol>)controllerClass` should be implemented in subclass of `CKCompositeComponent`


ocrickard:
Yep! Thanks for contributing. We definitely need to fix that.

Issue #913, opened 2018-06-20
ComponenKit should depend on Yoga 1.6 instead of 1.6.0


pkrmf:
We should try to avoid setting a dependency on a patch version. Imagine the following situation:

Application pulls 2 Frameworks as dependencies, **Framework A** and **ComponentKit**.
- **Framework A** has a dependency on **Yoga ~>1.8**
- **ComponentKit(master)** has a dependency on **Yoga ~>1.6.0**

Cocoapods won't allow the consuming application to pod install, since it will treat **1.6.0 to 1.8 as a major version change**, even though its not and there should be no breaking API changes.

**The correct approach here should be that ComponentKit has a dependency on Yoga ~>1.6 instead of 1.6.0**, this way, the application will be able to install Framework A, ComponentKit and Yoga 1.8.x

Let me know if you guys are ok with me updating the podspec!

madcato:
This issue is causing me this problem:

ComponentKit uses the identifier **YGJustifySpaceEvenly** defined in Yoga version **1.7** . But Cocoapods don't let me update Yoga to that version, because **yoga ~>1.6.0** configuration.

alickbass:
Seems fair, @pkrmf. Do you want to submit a PR?

pkrmf:
@alickbass Sure!

pkrmf:
@alickbass [PR](https://github.com/facebook/componentkit/pull/914) 

We should be looking to publishing the podspec to cocoapods master repo

pkrmf:
@alickbass just a reminder about publishing the new podspec to cocoapods. Is there any way I could help on that?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

alickbass:
@pkrmf I have just pushed to trunk! Thanks for your contribution and please let me know if everything is working for you!

pkrmf:
@alickbass you are the man! Thank you so much!

alickbass:
I am gonna close this issue. Feel free to reopen it in case something is not working

PR #779, opened 2017-03-09
Version 0.20

eczarny:

PR #914, opened 2018-06-22
Fixing podspec by removing dependency on patch version

pkrmf:
- ComponentKit will no longer have a dependency on the Yoga's patch version. We are going from 1.6.0 to 1.6.
- Bumped version of ComponentKit to 0.22.1. Let me know if that is OK. We should be looking to pushing this new podspec to cocoapods master repo

This PR solves [#913](https://github.com/facebook/componentkit/issues/913)


ComponentKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name = 'ComponentKit'
-  s.version = '0.22'
+  s.version = '0.22.1'
```
alickbass: The new version will contain some breaking changes in the framework, so I would rather make it 0.23 instead

alickbass:
Could you also add other changes like in [this PR](https://github.com/facebook/componentkit/pull/779/files)

ComponentKit.podspec
pkrmf: I would think Yoga 1.6.0 to 1.x shouldn’t have any breaking API changes. At the end, the third digit never involves api changes, it just for bug fixes. If that is not true in yoga’s case, then I can change the version to 0.23

Ideally, breaking API changes should only happen on major versions. In this scenario, if our dependency was moving from yoga 1.6 to 2.0

alickbass: Yoga doesn't, but the ComponentKit does. So it's better to be safe

pkrmf:
@alickbass I made the changes you requested

pkrmf:
@alickbass is the CI responsible for tagging the repository and then publishing the podspec to the cocoapods master repo?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

alickbass:
I don't think so. I will have to do it manually, after this PR gets merged

Issue #920, opened 2018-10-22
Support new build system


leavez:
when using the new build system in Xcode, which is default for xcode 10, it has an build error

```
error: Multiple commands produce '/Users/AAA/Library/Developer/Xcode/DerivedData/CKTableViewTransactionalDataSource-hfnosmzcdvuxuaanjgxnwmxagvqk/Build/Products/Debug-iphonesimulator/ComponentKit/ComponentKit.framework/Info.plist':
1) Target 'ComponentKit' (project 'Pods') has copy command from '/Users/gao/Parttime/CKTableViewTransactionalDataSource/Example/Pods/ComponentKit/ComponentKit/Info.plist' to '/Users/gao/Library/Developer/Xcode/DerivedData/CKTableViewTransactionalDataSource-hfnosmzcdvuxuaanjgxnwmxagvqk/Build/Products/Debug-iphonesimulator/ComponentKit/ComponentKit.framework/Info.plist'
2) Target 'ComponentKit' (project 'Pods') has process command with output '/Users/AAA/Library/Developer/Xcode/DerivedData/CKTableViewTransactionalDataSource-hfnosmzcdvuxuaanjgxnwmxagvqk/Build/Products/Debug-iphonesimulator/ComponentKit/ComponentKit.framework/Info.plist'
```

It's because the new build system cannot allow the info.plist being into 'Compile Sources'. In the podfile, the following could lead this problem. We should exclude the info.plist form source_files

```
 s.source_files = 'ComponentKit/**/*', 'ComponentTextKit/**/*'
```

alickbass:
Great! Thanks for submitting this issue! I will take a look at it!

PR #922, opened 2018-11-25
Fix new build system compatiblity

sebikaplun:
This issue
https://github.com/facebook/componentkit/issues/920

was being caused because the podspec includes all files, so the Info.plist was also being imported and the XCode warned that 2 info.plist were added the the project. 

This PR excludes the plist file


alickbass:
Thanks a lot for submitting this PR!

PR #926, opened 2019-03-14
Update Responder Chain documentation link

ajfigueroa:
The original [link](https://developer.apple.com/library/ios/documentation/EventHandling/Conceptual/EventHandlingiPhoneOS/event_delivery_responder_chain/event_delivery_responder_chain.htm) seems to no longer exist and takes users to an overview of UIKit.

After some digging, I found a more appropriate [link](https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/using_responders_and_the_responder_chain_to_handle_events) for the Responder Chain

### DIFF
--- a/_docs/08-responder-chain.md
+++ b/_docs/08-responder-chain.md
@@ -4,7 +4,7 @@ layout: docs
 permalink: /docs/responder-chain.html
 ---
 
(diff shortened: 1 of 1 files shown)


fabiomassimo:
🙌

PR #932, opened 2019-06-11
Replace `CKStackLayoutComponent` with `CKFlexboxComponent`

hanton:
Replace `CKStackLayoutComponent` with `CKFlexboxComponent` in the documents.


fabiomassimo:
Thank you for working on this!
Unfortunately this is not relevant anymore because we have recently published a new version of the website where this is issue has been already solved.

PR #933, opened 2019-06-15
Reorder doc numbers according to the nav bar order

hanton:
Fix a bug that clicking `Avoid Width 100%` section's `Next->` button will jump to `Component Hosting View` section, not the `Use Designated Initializer Style` section.


fabiomassimo:
Thank you for working on this!
Unfortunately this is not relevant anymore because we have recently published a new version of the website where this is issue has been already solved.

PR #941, opened 2019-10-03
used grammer

Rohangoyal2001:
carthage :- Carthage


fabiomassimo:
Hi @Rohangoyal2001, thank you for your contribution and interest in ComponentKit.
I think your suggested changes won't be necessary here as the reference to Carthage you proposed a change for, are meant to be used in a shell command where there is no need for capitalization.

I'll close this and let me know if something is not clear!

PR #944, opened 2019-10-17
Fix naming issue when adopting `CKDataSourceStateModifying` protocol

hanton:
Fix parameter naming issue when adopting `CKDataSourceStateModifying` protocol


cuva:
Hey @hanton. Thanks for submitting this, unfortunately we didn't get to it in time. Closing it as it looks like the changes are already reflected internally.

PR #945, opened 2020-01-02
[CK] Update CI build

kfirapps:


fabiomassimo:
I think you might want to replace the `--no-ri --no-rdoc` option with `--no-document` instead to keep the same behavior that was guaranteed with `--no-ri` ans `--no-rdoc`

kfirapps:
I think `-N` and `--no-document` are equivalent?

https://guides.rubygems.org/command-reference/#gem-install

kfirapps:
Closing. I'll land it internally :)

PR #946, opened 2020-03-20
Get CK compiled with Clang 10

maxovtsin:
Summary: - Fixed a few errors emitted by Clang 10

Differential Revision: D20549974


maxovtsin:
Good catch, thanks! 👍 
Missed the assert.

PR #956, opened 2020-09-17
Allow CKFlexboxBuilder to render children without initial child

kparentfb:
I found that `CKFlexboxComponentBuilder()` will never render its children if it's not initially constructed with an explicit child. For example:
 
```
auto flexbox = CKFlexboxComponentBuilder();
flexbox.child(componentA);
flexbox.build(); // componentA will not render
```

```
auto flexbox = CKFlexboxComponentBuilder().child(nil);
flexbox.child(componentA);
flexbox.build(); // componentA renders
```

I suspect it’s due to the logic here: https://fburl.com/diffusion/is68e3zp

Since hasActiveChild is false in the absence of an initial child prop, the new child isn’t pushed back, which prevents hasActiveChild from correctly recomputing, and no new children will ever be pushed onto the _children vector.

### DIFF
--- a/ComponentKit/LayoutComponents/FlexboxComponentBuilder.h
+++ b/ComponentKit/LayoutComponents/FlexboxComponentBuilder.h
@@ -369,11 +369,11 @@ class __attribute__((__may_alias__)) FlexboxComponentBuilder
    */
   auto &child(NS_RELEASES_ARGUMENT CKComponent *c)
   {
-    if (PropBitmap::isSet(PropsBitmap, FlexboxComponentPropId::hasActiveChild)) {
-      _children.push_back(_currentChild);
+    if (!PropBitmap::isSet(PropsBitmap, FlexboxComponentPropId::hasActiveChild)) {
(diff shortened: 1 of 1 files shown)


Andrey-Mishanin:
This is happening because you've split the builder setup in multiple statements. As you might have noticed, the `child()` method returns a slightly different type (with the corresponding bit set in the bitmap), and that information is later used in the `build()` method to decide whether we need to add the current child or not. In your case setting this bit is not reflected in the bitmap that `build()` sees, so the child is not rendered.

In general, we advise against storing builders in variables exactly because it can cause bugs like this. If you have to do it, make sure you always reassign the builder variable, so this becomes a compile-time error instead of a runtime bug. (We could forbid storing builders in variables by removing the copy constructor but that's not currently possible for other uninteresting reasons.)

kparentfb:
Thanks for the detailed reply @Andrey-Mishanin! This makes sense to me, I think I've even run into the builder's typing issue elsewhere.

PR #972, opened 2022-06-12
Fix a bunch of issues from the Xcode 14 upgrade

lanza:
The clang-14 in Xcode 14 introduces a bunch of new errors including many
missing `self->` and a couple missing copy-assignment/copy-constructors.
Fix them all here so that WildeGuess builds with Xcode14.


cuva:
Thanks @lanza!

